PR #119, opened 2014-03-27
Incorrect signature

mattbennett:


richardasaurus:
This is old and merged, but couldn't https://github.com/JukkaL/mypy be a good solution for this issue?

PR #155, opened 2014-09-24
WIP; require config in each entrypoint/dependency

mattbennett:
Potential pattern to support entrypoints/dependencies with non-global config. Raised for discussion.

Summary:

* You should be able to pass individually relevant config values into entrypoints/dependencies
* These may be bare values _or_ a callable that returns the value when called with the hosting ``container``
* Changes applied here for ``@timer``, ``@rpc`` and ``rpc_proxy``
* The timer implementation works fine

This pattern fundamentally clashes with the re-use of ``RpcConsumer``s and ``QueueConsumer``s by multiple ``RpcProvider`` instances (which may now use different AMQP URIs). Now may be the right time to rebuild our RPC implementation around shared channels, rather than shared dependencies.

### DIFF
--- a/nameko/dependencies.py
+++ b/nameko/dependencies.py
@@ -460,3 +460,23 @@ def prepare_dependencies(container):
     return chain(
         prepare_injection_providers(container, include_dependencies=True),
         prepare_entrypoint_providers(container, include_dependencies=True))
+
+
+class ConfigValue(object):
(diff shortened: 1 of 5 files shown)


mattbennett:
It turns out there are many better ways to do this. Cleaning up this PR, finally.

PR #187, opened 2015-01-09
Dependency redux

mattbennett:


nameko/containers.py
```diff
@@ -127,9 +139,14 @@ def __init__(self, service_cls, worker_ctx_cls, config):
...
             config.get(MAX_WORKERS_CONFIG_KEY) or DEFAULT_MAX_WORKERS)
 
-        self.dependencies = DependencySet()
-        for dep in prepare_dependencies(self):
-            self.dependencies.add(dep)
+        # the relationship between extensions and the container is symbiotic;
+        # containers are responsible for starting and stopping their
+        # extensions, and extensions are able to interact with the container
+        # via its :attr:`ServiceContainer.interface`.
+        self.dependencies = ExtensionSet()
```
davidszotten: the `ExtensionSet` class is a bit strange to me. in `discover_extensions` we combine providers and entrypoints, only to put them in a set with methods that loop over them to split them up again...?

nameko/containers.py
```diff
@@ -127,9 +139,14 @@ def __init__(self, service_cls, worker_ctx_cls, config):
...
+        for name, ext in discover_extensions(service_cls):
+            bound = ext.bind(name, self.interface)
```
davidszotten: don't feel like this interaction reads very well. will come back to this

nameko/containers.py
```diff
@@ -140,8 +157,13 @@ def __init__(self, service_cls, worker_ctx_cls, config):
         self._died = Event()
 
     @property
+    def interface(self):
```
davidszotten: what is this for?

nameko/containers.py
```diff
@@ -140,8 +157,13 @@ def __init__(self, service_cls, worker_ctx_cls, config):
...
+        # TODO: a safe subset of the container to make available to extensions
+        return self
+
+    @property
     def entrypoints(self):
```
davidszotten: see above. we may just want to keep two separate sets

nameko/contextdata.py
```diff
@@ -2,33 +2,27 @@
...
+    context_key = None
 
     def acquire_injection(self, worker_ctx):
         return worker_ctx.data.get(self.context_key)
 
 
-@injection
-def language():
-    return DependencyFactory(ContextDataProvider, LANGUAGE_CONTEXT_KEY)
+class language(ContextDataProvider):
```
davidszotten: CamelCase?

nameko/dependencies.py
```diff
@@ -3,64 +3,65 @@
 """
 from __future__ import absolute_import
```
davidszotten: `-> extensions.py`?

nameko/dependencies.py
```diff
@@ -3,64 +3,65 @@
...
+    container = None
 
-class DependencyProvider(object):
-
-    bound = False
-    name = "<unbound-provider>"
+    def __init__(self, *args, **kwargs):
+        # sharing status is not persisted into state;
+        # cloned extensions will never be shared
+        self.__shared = kwargs.pop('shared', False)
```
davidszotten: i though this was going to happen via different `bind` implementations rather than params

nameko/dependencies.py
```diff
@@ -3,64 +3,65 @@
...
+        self.__state = (args, kwargs)
```
davidszotten: not quite `state` i guess

nameko/dependencies.py
```diff
@@ -3,64 +3,65 @@
...
+        super(Extension, self).__init__()
 
-    def prepare(self):
-        """ Called when the service container starts.
+    def before_start(self):
+        """ Called before the service container starts.
 
-        DependencyProviders should do any required initialisation here.
+        Extensions should do any required initialisation here.
```
davidszotten: possibly a few words about what the before/start split

nameko/dependencies.py
```diff
@@ -68,75 +69,71 @@ def kill(self):
...
+        # if this extension is shared and there's already a cloned instance,
+        # return that
+        if self.__shared:
+            shared_extensions.setdefault(container, {})
+            shared = shared_extensions[container].get(self.sharing_key)
+            if shared:
+                return shared
+
+        # clone this extension and associate it with the container
+        clone = self.clone()
```
davidszotten: if this is the only use case, we could make name and container params to clone

nameko/dependencies.py
```diff
@@ -68,75 +69,71 @@ def kill(self):
...
+            shared_extensions[container][self.sharing_key] = clone
+
+        return clone
+
+    def clone(self):
+        if self.__state is None:
+            raise RuntimeError('Undefined extension state. Did you forget to '
+                               'call super().__init__() in your subclass?')
+
+        if self.__clone:  # TODO: need this even with the check in bind above?
```
davidszotten: what is this trying to protect against?

nameko/dependencies.py
```diff
@@ -68,75 +69,71 @@ def kill(self):
...
-
-        For example, with an instance of `:class:~nameko.rpc.RpcProvider`
-        called ``rpc``::
-
-            >>> deps = list(rpc.nested_dependencies)
-            >>> deps
-            [<nameko.rpc.RpcConsumer object at 0x10d125690>,
-             <nameko.messaging.QueueConsumer object at 0x10d5b8f50>]
-            >>>
+    def extensions(self):
```
davidszotten: could we have a more descriptive name here? (not sure what yet) 

(esp. given that we include self)

nameko/dependencies.py
```diff
@@ -68,75 +69,71 @@ def kill(self):
...
+        yield self
+
+    @property
+    def sharing_key(self):
+        return type(self)
 
     def __repr__(self):
-        if not self.bound:
-            return '<{} [unbound] at 0x{:x}>'.format(
+        if self.name is UNBOUND_NAME:
```
davidszotten: a little odd to use a string as a sentinel. (i.e. should this just be `==` do we want stronger guarantees)

nameko/dependencies.py
```diff
@@ -146,11 +143,12 @@ def __repr__(self):
...
 
 
-class EntrypointProvider(DependencyProvider):
-    pass
+class InjectionProvider(Extension):
 
-
-class InjectionProvider(DependencyProvider):
+    def __init__(self, *args, **kwargs):
+        # InjectionProviders cannot be shared
```
davidszotten: goes away with subclass

nameko/dependencies.py
```diff
@@ -239,233 +267,76 @@ def injections(self):
...
-            def foobar():
-                pass
-
-    """
-    def registering_decorator(wrapper, fn, args, kwargs):
-        """ Verify that ``decorator_func`` returns a DependencyFactory and
-        register ``fn`` as an entrypoint.
+    def __init__(self, *args, **kwargs):
+        # Entrypoints cannot be shared
+        kwargs.pop('shared', False)
```
davidszotten: and here

nameko/dependencies.py
```diff
@@ -239,233 +267,76 @@ def injections(self):
...
 
-        return fn
+        def registering_decorator(fn, args, kwargs):
+            instance = cls(*args, **kwargs)
+            register_entrypoint(fn, instance)
+            return fn
 
-    @wraps(decorator_func)
-    def wrapper(*args, **kwargs):
         if len(args) == 1 and isinstance(args[0], types.FunctionType):
```
davidszotten: can `wrapt` no longer help?

nameko/dependencies.py
```diff
@@ -239,233 +267,76 @@ def injections(self):
...
             # usage without arguments to the decorator:
             # @foobar
             # def spam():
             #     pass
-            return registering_decorator(wrapper, args[0], tuple(), {})
```
davidszotten: what happens for

```
@foobar()
def spam():
    pass
```

nameko/containers.py
mattbennett: It is a bit weird. If we make the superclass responsible for calling the container lifecycle methods instead then we can definitely refer to the entrypoints and injections/dependencies explicitly.

nameko/containers.py
mattbennett: Nope. This was as straightforward as I could make it with the container knowing about all extension instances. I can't believe I didn't think about using the superclass!

nameko/containers.py
mattbennett: This is an experiment / talking point. The idea is to pass a limited interface of some sort into the extensions, rather than giving them unfettered access to the whole container.

nameko/dependencies.py
mattbennett: Yep, was just waiting for an initial pass on the review first.

nameko/dependencies.py
mattbennett: I'd forgotten what we agreed. Using a ``SharedExtension`` superclass is nicer.

nameko/dependencies.py
mattbennett: Yeah that might be nice

nameko/dependencies.py
mattbennett: It's not the entire state of the extension no, only the initialisation parameters. Maybe ``__params``?

nameko/dependencies.py
mattbennett: Good idea

nameko/dependencies.py
mattbennett: It's possible we'll lose these. It is mostly me asserting that my implementation worked correctly, but you're not supposed to make clones of clones. It might work but it's not correct.

An alternative would be to delete the clone method during the clone process. That would ensure safety without the extra code, but might be more difficult to understand for any developer coming across it.

nameko/dependencies.py
mattbennett: Sorry, yes. The docstring here is only a token effort!

nameko/dependencies.py
mattbennett: Wrapt only supports decorators with optional *kwargs*, not both args and kwargs. I guess the comparison we're doing below is considered icky - it means you can never write a decorator that genuinely accepts a function as its first arg e.g.

```python
def foo():
    pass

@bar(foo)
def baz():
    pass

File: nameko/dependencies.py (hunk 17 above)
Comment (mattbennett) [inline nameko/dependencies.py]: This works fine. I will add some test coverage

File: nameko/containers.py (hunk 2 above)
Comment (davidszotten) [inline nameko/containers.py]: let's have a chat about this. need to think about how we want "sub" extensions and containers to relate

File: nameko/messaging.py [hunk 18]
```diff
@@ -136,14 +137,13 @@ def publish(msg, **kwargs):
         return publish
 
 
-@dependency
-def queue_consumer():
-    return DependencyFactory(QueueConsumer)
+# backwards compat
```
davidszotten: we're making some pretty big changes. might just call it 2.0

nameko/timer.py
```diff
@@ -5,40 +5,43 @@
...
-    If the `config_key` is given the value for that key in the config will be
-    used as the interval otherwise the `interval` provided will be used.
+        Either `interval` or `config_key` must be provided. If given,
+        `config_key` specifies a key in the config will be used as the
+        interval.
 
-    Example::
+        Example::
 
-        class Foobar(object):
```
davidszotten: `class Service(object):`?

nameko/containers.py
```diff
@@ -127,9 +127,25 @@ def __init__(self, service_cls, worker_ctx_cls, config):
         self.max_workers = (
             config.get(MAX_WORKERS_CONFIG_KEY) or DEFAULT_MAX_WORKERS)
 
-        self.dependencies = DependencySet()
-        for dep in prepare_dependencies(self):
-            self.dependencies.add(dep)
+        self.entrypoints = SpawningSet()
+        self.dependencies = SpawningSet()
+        self.subextensions = SpawningSet()
```
mattbennett: I've done away with the ``ExtensionSet`` and we now deal with the three categories explicitly. I'm not super keen on ``subextensions`` as a name but that is what they are.

davidszotten: do we need to distinguish them at all from `dependencies`?

nameko/containers.py
```diff
@@ -140,22 +156,23 @@ def __init__(self, service_cls, worker_ctx_cls, config):
...
-    def entrypoints(self):
-        return filter(is_entrypoint_provider, self.dependencies)
+    def extensions(self):
+        return self.entrypoints | self.dependencies | self.subextensions
 
     @property
-    def injections(self):
-        return filter(is_injection_provider, self.dependencies)
+    def interface(self):
+        # TODO: design interface to expose to extensions
```
davidszotten: todo

nameko/containers.py
```diff
@@ -194,25 +211,24 @@ def stop(self):
...
-            # it should be safe now to stop any injection as there is no
+            # it should be safe now to stop any dependency as there is no
             # active worker which could be using it
-            dependencies.injections.all.stop()
+            self.dependencies.all.stop()
 
-            # finally, stop nested dependencies
-            dependencies.nested.all.stop()
+            # finally, stop remaining extensions
+            self.subextensions.all.stop()
```
davidszotten: do we have anything that depends on this separation?

nameko/containers.py
```diff
@@ -316,8 +333,8 @@ def spawn_managed_thread(self, run_method, protected=False):
         """ Spawn a managed thread to run ``run_method``.
 
         Threads can be marked as ``protected``, which means the container will
-        not forcibly kill them until after all dependencies have been killed.
-        Dependencies that require a managed thread to complete their kill
+        not forcibly kill them until after all extensions have been killed.
+        Extensions that require a managed thread to complete their kill
         procedure should ensure to mark them as ``protected``.
```
davidszotten: maybe an example of why one might want to use this

nameko/containers.py
```diff
@@ -348,11 +364,15 @@ def _run_worker(self, worker_ctx, handle_result):
 
         with _log_time('ran worker %s', worker_ctx):
 
-            self.dependencies.injections.all.inject(worker_ctx)
+            for dependency in self.dependencies:
```
davidszotten: seems a little wasteful to serialise this. let's see if we can think of a nice way to get it back in parallel

nameko/containers.py
```diff
@@ -383,7 +403,7 @@ def _run_worker(self, worker_ctx, handle_result):
                 del exc_info
 
                 self.dependencies.all.worker_teardown(worker_ctx)
-                self.dependencies.injections.all.release(worker_ctx)
+                # release?
```
davidszotten: did we have an intended use-case? i.e. some extension that wanted to use a dependency in another extensions teardown? think we ended up just using teardown tidy up from `acquire_injection`

nameko/extensions.py
```diff
@@ -0,0 +1,308 @@
...
+import inspect
+import types
+import weakref
+
+from eventlet.event import Event
+
+from logging import getLogger
+_log = getLogger(__name__)
+
+
```
davidszotten: not sure what convention we have wrt blank lines here

nameko/extensions.py
```diff
@@ -0,0 +1,308 @@
...
+ENTRYPOINT_EXTENSIONS_ATTR = 'nameko_entrypoints'
+
+shared_extensions = weakref.WeakKeyDictionary()
+
+
+class Extension(object):
+
+    __clone = False
```
davidszotten: maybe `__is_clone` (if we need this at all)

nameko/extensions.py
```diff
@@ -0,0 +1,308 @@
...
+    __params = None
+
+    def __new__(cls, *args, **kwargs):
+        inst = super(Extension, cls).__new__(cls, *args, **kwargs)
+        inst.__params = (args, kwargs)
+        return inst
+
+    def __init__(self, *args, **kwargs):
```
davidszotten: let's put the note somewhere else

nameko/extensions.py
```diff
@@ -0,0 +1,308 @@
...
+    attr_name = None
+
+    def bind(self, service_name, attr_name):
+        """
+        """
+        self.service_name = service_name
+        self.attr_name = attr_name
+
+    def acquire_injection(self, worker_ctx):
+        """ Called before worker execution. An Dependency should return
```
davidszotten: typo "An Dep"

nameko/extensions.py
```diff
@@ -0,0 +1,308 @@
...
+
+def is_dependency(obj):
+    return isinstance(obj, Dependency)
+
+
+def is_entrypoint(obj):
+    return isinstance(obj, Entrypoint)
+
+
+def iter_extensions(extension):
```
davidszotten: maybe a comment mentioning that this is depth-first

nameko/containers.py
```diff
@@ -140,22 +156,23 @@ def __init__(self, service_cls, worker_ctx_cls, config):
...
-        return filter(is_entrypoint_provider, self.dependencies)
+    def extensions(self):
+        return self.entrypoints | self.dependencies | self.subextensions
 
     @property
-    def injections(self):
-        return filter(is_injection_provider, self.dependencies)
+    def interface(self):
+        # TODO: design interface to expose to extensions
+        return self
```
davidszotten: this might be a good place to make sure extensions only have a weakref to the container. still investigating whether this is strictly needed, but it might be

test/standalone/test_rpc_proxy.py
```diff
@@ -13,10 +13,10 @@
 from nameko.exceptions import RpcConnectionError
 
 
-class ContextReader(InjectionProvider):
+class ContextReader(Dependency):
     """ Access values from the worker context data.
 
-    This is a test facilty! Write specific InjectionProviders to make use of
+    This is a test facilty! Write specific Dependencys to make use of
```
davidszotten: typo from sed

test/test_container.py
```diff
@@ -47,43 +49,38 @@ def kill(self, exc=None):
...
+        self._log_call(('worker_teardown', worker_ctx))
+        super(CallCollectorMixin, self).worker_teardown(worker_ctx)
 
-@injection
-def call_collector():
-    return DependencyFactory(CallCollectingInjectionProvider)
 
+foobar = CallCollectingEntrypoint.decorator
+
+# compat
```
davidszotten: compat in a test?

test/test_db.py
```diff
@@ -39,8 +39,8 @@ def test_concurrency():
     service_instance = Mock()
 
     def inject(worker_ctx):
```
davidszotten: possibly out of scope, but this looks like it's messing with internals in a way it probably shouldn't

test/test_extension_sharing.py
```diff
@@ -0,0 +1,116 @@
...
+
+from mock import call
+
+from nameko.extensions import Dependency, SharedExtension, Extension
+from nameko.testing.utils import get_extension
+
+
+class CallCollectorMixin(object):
+
+    calls = defaultdict(lambda: defaultdict(list))
```
davidszotten: clear between tests?

test/test_extensions.py
```diff
@@ -0,0 +1,231 @@
...
+    # extension instances are different between dependencies
+    assert inj1 != inj2
+    assert inj1.ext != inj2.ext
+
+
+def test_clones_marked_as_clones():
+    container = Mock()
+
+    ext = SimpleExtension()
+    assert ext._Extension__clone is False
```
davidszotten: any reason not to just use `ext.is_clone()` here?

nameko/extensions.py
```diff
@@ -0,0 +1,308 @@
...
+        discards these and disconnects from rabbit as soon as possible.
+
+        Extensions should not raise during kill, since the container
+        is already dying. Instead they should log what is appropriate and
+        swallow the exception to allow the container kill to continue.
+        """
+
+    def clone(self, container):
+        if self.is_clone:
+            raise RuntimeError('Cloned extensions cannot be cloned.')
```
davidszotten: do we have some common errors that might cause this worth mentioning? (a la "did you ..."?)

test/test_extensions.py
```diff
@@ -0,0 +1,231 @@
...
+
+    container = container_factory(Service, {})
+    container.start()
+
+    assert len(container.extensions) == 2
+    dyn_inj = get_extension(container, DynamicInjection)
+    assert dyn_inj.ext.arg == "argument_for_extension"
+
+
+def test_is_extension():
```
davidszotten: i guess these are just chasing coverage, but a negative test might otherwise also be in order

test/test_provider_collector.py
```diff
@@ -0,0 +1,34 @@
...
+
+    collector.unregister_provider(provider1)
+    assert provider1 not in collector._providers
+
+    # unregister missing provider is a no-op
+    collector.unregister_provider(provider1)
+
+    # stop() should block while a provider remains
+    with pytest.raises(eventlet.Timeout):
+        with eventlet.Timeout(0):
```
davidszotten: guess we might want some small, but non-zero amount of time here (0.1?)

nameko/containers.py
mattbennett: I don't think it's possible to abstract away the differences. "Dependencies" can be injected into service code; they are the top level. Whereas "extensions" is the catch-all for everything managed by the container, including entrypoints.

We danced around having entrypoints and dependencies manage their own subextensions so the container didn't need to know about them. With more intelligent ``stop`` behaviour that may be be possible, but it currently isn't. If it does become possible then it'll be pretty easy to transition to it from here, and the ugly "subextensions" variable will go away.

nameko/containers.py
mattbennett: Yep. I will at least make it a ``weakref`` for now

nameko/containers.py
mattbennett: Good call

nameko/containers.py
mattbennett: Yeah, there's potentially quite a lot of I/O going on in ``acquire_injection``, depending on how lazy the injections are.

nameko/containers.py
mattbennett: It's not safe to use another dependency during teardown (there's no guarantee of order; teardown all happens in parallel) so I don't think there is a use case. The worker is about to be discarded so I don't think it's really required.

Not sure I follow your last sentence though.

nameko/extensions.py
mattbennett: No. You're most likely to hit this in tests where you're manually creating extensions.

nameko/extensions.py
mattbennett: Good call

test/test_extension_sharing.py
mattbennett: Whoops, probably a good idea!

test/test_extensions.py
mattbennett: I think these tests precede ``is_clone``. Will update them

test/test_provider_collector.py
mattbennett: ``collect.stop`` doesn't yield, so this is fine. The ``ProviderCollector`` is slated for deprecation anyway...

nameko/extensions.py
mattbennett: We don't have one but I wouldn't be opposed to it.

nameko/extensions.py
mattbennett: We won't when we start stashing the container.

nameko/containers.py
mattbennett: I've reverted to using an ``inject`` method on ``Dependency`` for now, which works but seems a bit backwards to me.

I find the ``SpawningSet`` a bit clunky really. I'd rather have a ``spawning_imap`` helper or similar that just accepts a callable. Would be more explicit.

davidszotten: we keeping `inject`?

nameko/containers.py
davidszotten: think we're still missing this

davidszotten: ah, lost when landing `bind-clone-alternative`

test/test_extensions.py
```diff
@@ -0,0 +1,232 @@
...
+    # extension instances are different between dependencies
+    assert inj1 != inj2
+    assert inj1.ext != inj2.ext
+
+
+def test_is_bound():
+    container = Mock()
+
+    ext = SimpleExtension()
+    assert ext.is_bound is False
```
davidszotten: normally `assert not ext.is_bound` / `assert ext.is_bound`

nameko/extensions.py
```diff
@@ -0,0 +1,322 @@
...
+
+    @property
+    def container(self):
+        """ The :class:`~nameko.containers.ServiceContainer` instance to
+        which this Extension is bound, otherwise `None`.
+        """
+        # cannot raise here; fails during introspection
+        return self.__container
+
+    @property
```
davidszotten: is there a reason this is a `property` (as opposed to a regular method)?

nameko/extensions.py
```diff
@@ -0,0 +1,322 @@
...
+        swallow the exception to allow the container kill to continue.
+        """
+
+    def bind(self, container):
+        """ Get an instance of this Extension to bind to `container`.
+        """
+
+        def clone(prototype):
+            if prototype.is_bound:
+                raise RuntimeError('Bound extensions cannot be bound.')
```
davidszotten: possibly "cannot call `bind` on bound extensions"?

nameko/extensions.py
davidszotten: bump?

nameko/extensions.py
```diff
@@ -0,0 +1,322 @@
...
+
+        instance = clone(self)
+
+        # recurse over sub-extensions
+        for name, ext in inspect.getmembers(self, is_extension):
+            setattr(instance, name, ext.bind(container))
+        return instance
+
+    @property
+    def container(self):
```
davidszotten: how come this indirection?

nameko/extensions.py
```diff
@@ -0,0 +1,322 @@
...
+            cls = type(prototype)
+            args, kwargs = prototype.__params
+            instance = cls(*args, **kwargs)
+            instance.__container = container
+            return instance
+
+        instance = clone(self)
+
+        # recurse over sub-extensions
+        for name, ext in inspect.getmembers(self, is_extension):
```
davidszotten: should this be inspecting the class?

davidszotten:
note to add a test for gc

nameko/extensions.py
mattbennett: No, sub-extensions can be added at init time

test/test_db.py
mattbennett: What makes you say that?

nameko/containers.py
mattbennett: I think for this PR, yes

test/test_db.py
davidszotten: this is accessing container internals directly. would be good if that wasn't required for extension authors. could this not e.g. just call `spawn_worker` or similar?

nameko/extensions.py
```diff
@@ -0,0 +1,319 @@
...
+        swallow the exception to allow the container kill to continue.
+        """
+
+    def bind(self, container):
+        """ Get an instance of this Extension to bind to `container`.
+        """
+
+        def clone(prototype):
+            if prototype.is_bound():
+                raise RuntimeError('Cannot `bind` a bound extensions.')
```
davidszotten: "bound extensions" or "a bound extension"

davidszotten:
:+1: 

awesome work!

davidszotten:
included in 2.0

PR #188, opened 2015-01-19
Bind clone alternative

mattbennett:


nameko/extensions.py
```diff
@@ -21,6 +21,7 @@
 
 class Extension(object):
 
+    container = None
     __clone = False
```
davidszotten: i guess we can drop `__clone` and have `is_clone` just look for `container`?

nameko/extensions.py
```diff
@@ -72,19 +73,26 @@ def kill(self):
...
 
-    def clone(self, container):
-        if self.is_clone:
-            raise RuntimeError('Cloned extensions cannot be cloned.')
+    def bind(self, container):
+        """ Get an instance of this Extension to bind to `container`.
+        """
+
+        def clone(prototype):
+            if prototype.is_clone:
```
davidszotten: actually, we could make `clone` an internal concept and refer to them as `bound` ?

nameko/extensions.py
mattbennett: I think so

nameko/extensions.py
mattbennett: Yep, ``is_clone`` will become ``is_bound``. Cloning is an implementation detail.

PR #190, opened 2015-01-23
try harder to serialize exceptions

### LINKED ISSUE #189: exception serialisation bug
```
from nameko.exceptions import serialize

def test_serialize_args():
    cause = Exception('oops')
    exc = CustomError('something went wrong', cause)

    assert json.dumps(serialize(exc))
```

fails. `serialize` probably needs to stringify all args, not just `exc`

davidszotten:
fixes #189

### DIFF
--- a/nameko/rpc.py
+++ b/nameko/rpc.py
@@ -210,18 +210,17 @@ class Responder(object):
         if exc_info is not None:
             error = serialize(exc_info[1])
 
-        # disaster avoidance serialization check
-        # `result` and `error` must both be json serializable, otherwise
-        # the container will commit suicide assuming unrecoverable errors
-        # (and the message will be requeued for another victim)
-        for item in (result, error):
-            try:
-                json.dumps(item)
-            except Exception:
-                result = None
-                exc_info = sys.exc_info()
-                # `error` below is guaranteed to serialize to json
-                error = serialize(UnserializableValueError(item))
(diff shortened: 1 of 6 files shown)


nameko/utils.py
```diff
@@ -19,6 +20,29 @@ def repr_safe_str(value):
...
+    if isinstance(value, dict):
+        return {
+            safe_for_json(key): safe_for_json(val)
+            for key, val in value.iteritems()
+        }
+    if isinstance(value, collections.Iterable):
+        return map(safe_for_json, value)
+
+    try:
+        return unicode(value)
```
mattbennett: Did we land on ``unicode`` instead of ``repr``?

nameko/rpc.py
```diff
@@ -210,18 +210,17 @@ def send_response(self, container, result, exc_info, **kwargs):
...
-                error = serialize(UnserializableValueError(item))
+        # disaster avoidance serialization check: `result` must be json
+        # serializable, otherwise the container will commit suicide assuming
+        # unrecoverable errors (and the message will be requeued for another
+        # victim)
+        try:
+            json.dumps(result)
+        except Exception:
+            exc_info = sys.exc_info()
+            # `error` below is guaranteed to serialize to json
```
mattbennett: actually true now ;)

mattbennett:
:+1:

PR #192, opened 2015-01-27
Permissions usecases

mattbennett:
Raised for discussion only.

Tests demonstrating various permissions use cases and their caveats:
* Permission to request RPC from any service in the cluster
* Permission to request RPC from a specific service (not currently possible)
* Permission to listen to events from any service in the cluster
* Permission to listen to events from a specific service
* Permission to dispatch events into the cluster


mattbennett:
This was useful for discussion but is not helpful as a PR, so closing.

PR #194, opened 2015-01-30
More helpful error from entrypoint_waiter

mattbennett:


test/testing/test_services.py
```diff
@@ -373,6 +374,20 @@ class ExampleEvent(Event):
...
 
+def test_entrypoint_waiter_timeout(container_factory, rabbit_config):
+    container = container_factory(Service, rabbit_config)
+    container.start()
+
+    class ExampleEvent(Event):
+        type = "eventtype"
+
+    with pytest.raises(EntrypointWaiter.Timeout) as exc_info:
+        with entrypoint_waiter(container, 'handle', timeout=0.1):
```
davidszotten: possibly add another zero (`0.01`)?

noisyboiler:
Test looks good to me - but it just looks like you're testing a test helper, which is not expected in `test_service`. Does this relate to framework code? Do we need to test test helpers? Apologies if i've missed the point here!

mattbennett:
The code in ``nameko/testing`` is providing test helpers for other users of the library. So if you build services in nameko you can use these helpers when writing tests for them. We should make sure they work as expected :)

PR #196, opened 2015-02-09
fix for newer pyflakes

davidszotten:


test/legacy/test_proxy.py
```diff
@@ -258,7 +252,9 @@ def test_fallback_to_call():
 def test_timeout(nova, constructor):
     connection = ANY
     context = Mock()
-    context_factory = lambda: context
+
+    def context_factory():
+        return context
```
mattbennett: Why is this one different?

davidszotten: used twice

mattbennett:
:+1:

PR #197, opened 2015-02-09
Docs 2.0

mattbennett:


docs/examples/advanced_http.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class Service(object):
+
+    @http('GET', '/404')
+    def does_not_exist(self, value):
+        return 404, "Does Not Exist"
+
+    @http('GET', '/google')
+    def redirect(self,):
```
junkafarian: stray comma

docs/installation.rst
```diff
@@ -0,0 +1,17 @@
+.. _installation:
+
+Installation
+============
+
+RabbitMQ
+--------
+
+Several of nameko's built-in features rely on RabbitMQ.
```
junkafarian: maybe just link off to https://www.rabbitmq.com/download.html

docs/examples/testing/unit_test.py
```diff
@@ -6,6 +6,8 @@
 
 
 class ConversionService(object):
+    """ Service under test
+    """
     math = RpcProxy("math_service")
```
junkafarian: can we please call this `maths`. It's one of my most hated americanisms

junkafarian: Also, if we encourage the convention, we might want to do a pass on these examples adding the `_rpc` suffix to proxy declarations

docs/examples/travis.py
```diff
@@ -0,0 +1,43 @@
...
+    @rpc
+    def status_message(self, owner, repo):
+        status = self.webservice.repo_status(owner, repo)
+        outcome = "passing" if status['last_build_result'] else "failing"
+
+        return "Project {slug} {outcome} since {timestamp}.".format({
+            'repo': status['slug'],
+            'outcome': outcome,
+            'timestamp': status['last_build_finished_at']
+        })
```
junkafarian: nice

docs/built_in_extensions.rst
```diff
@@ -0,0 +1,47 @@
...
+
+Normal RPC calls block until the remote method completes, but proxies also have an "async" calling mode to background or parallelize RPC calls:
+
+.. literalinclude:: examples/async_rpc.py
+
+
+Events (Pub-Sub)
+----------------
+
+Nameko Events is an asynchronous messaging system, similar to the Publish-Subscriber pattern. Services dispatch events that may be received by zero or more others:
```
tonybajan: this is textbook publish-subscribe pattern, don't think you need the "similar"

docs/cli.rst
```diff
@@ -0,0 +1,45 @@
...
+======================
+
+Nameko ships with a command line interface to make hosting and interacting with services as easy as possible.
+
+Running a Service
+-----------------
+
+::
+
+    $ nameko run [<module>[:<ServiceClass>]]
```
tonybajan: `namko run` by itself doesn't seem to be valid – too many square brackets here

docs/examples/advanced_http.py
```diff
@@ -0,0 +1,12 @@
+from nameko.web import http
+
+
+class Service(object):
+
+    @http('GET', '/404')
+    def does_not_exist(self, value):
+        return 404, "Does Not Exist"
```
tonybajan: I'm not sure I'd use this as an example, as `does_not_exist` clearly does exist and deliberately returns a 404 – it's not a fallback hander. Perhaps a 204 or 403 instead?

docs/examples/http.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class Service(object):
+
+    @http('GET', '/get/<int:value>')
+    def get_method(self, value):
+        return {'value': value}
+
+    @http('POST', '/post')
+    def do_post(self, value):
```
tonybajan: maybe `body` over `value` (assuming this is the post body?)

docs/examples/testing/large_integration_test.py
```diff
@@ -1,15 +1,12 @@
 """
-In a large integration test of highly connected services, it's preferable to
-limit interactions to only those under test.
-
 This file defines several toy services that interact to form a shop of the
 famous ACME Corporation. The AcmeShopService relies on the StockService,
 InvoiceService and PaymentService to fulfil its orders. They are not best
 practice examples! They're minimal services provided for the test at the
 bottom of the file.
```
tonybajan: I'd still follow PEP8 by making adding an `Error` suffix to the exceptions. There's also a confusion between `NotImplementedError` (the built-in exception) and `NotImplemented` (the built-in sentinel for comparators) which isn't a thing that can be raised.

docs/key_concepts.rst
```diff
@@ -0,0 +1,87 @@
...
+A dependency is an opportunity to hide code that isn't part of the core service logic. The dependency's interface to the service should be as simple as possible.
+
+Declaring dependencies in your service is a good idea for :ref:`lots of reasons <benefits_of_dependency_injection>`, and you should think of them as the gateway between service code and everything else. That includes other services, external APIs, and even databases.
+
+.. _workers:
+
+Workers
+^^^^^^^
+
+Workers are created when an entrypoint fires. A worker is just an instance of the service class, but with interfaces to its dependencies injected into it.
```
tonybajan: I'm not sure "interfaces" is the clearest word here – it's often used to describe something implementation free / pure virtual, but this is where you get the concrete implementation of the dependency in the nameko lifecycle.

docs/key_concepts.rst
```diff
@@ -0,0 +1,87 @@
...
+
+Note that a worker only lives for the execution of one method - services are stateless from one call to the next, which encourages the use of dependencies.
```
tonybajan: typography: hyphen should be an en dash

docs/license.rst
```diff
@@ -0,0 +1,14 @@
+License
+=======
+
```
tonybajan: Should include the line `Copyright 2015 Lifealike Ltd ` here. (see the appendix to the license)

docs/similar_projects.rst
```diff
@@ -0,0 +1,16 @@
+Similar and Related Projects
+============================
+
+Celery
+------
+
+`Celery <http://celery.readthedocs.org/>`_ is a distributed task queue. It lets you define "tasks" as Python functions and execute them on a set of remote workers, not unlike nameko RPC.
+
+Celery is usually used as an add-on to existing applications, to defer processing or otherwise outsource some work to a remote machine. You could also achieve this with nameko, but Celery includes many more primitives for the distribution of tasks and the collection of results.
+
```
tonybajan: Ideal place for a recipe for soup

docs/testing.rst
```diff
@@ -0,0 +1,78 @@
...
+.. literalinclude:: examples/testing/unit_test.py
+
+In some circumstances it's helpful to provide an alternative dependency, rather than use a mock. This may be a fully functioning replacement (e.g. a test database session) or a lightweight shim that provides partial functionality.
+
+.. literalinclude:: examples/testing/unit_test_alternative_dependency.py
+
+Integration Testing
+-------------------
+
+Integration testing in nameko means testing the interface between a number of services. The recommended way is to run all the services being tested in the normal way, and trigger behaviour by manually "firing" an entrypoint:
```
tonybajan: "manually" doesn't mean much here – especially I was expecting a standalone rpc proxy to make the call and it's actually using `entrypoint_hook`. Maybe `and trigger behaviour by "firing" an entrypoint using a helper:` ?

davidszotten:
https://pypi.python.org/pypi/sphinxcontrib-spelling/ might be worth a look, used by e.g. django

docs/what_is_nameko.rst
```diff
@@ -0,0 +1,38 @@
...
+It comes with built-in support for:
+
+    * RPC over AMQP
+    * Asynchronous events (pub-sub) over AMQP
+    * Simple HTTP GET and POST
+    * Websocket RPC and subscriptions
+
+Out of the box you can build a service that can respond to RPC messages, dispatch events on certain actions, and listen to events from other services. It could also have HTTP interfaces for clients that can't speak AMQP, and a websocket interface for, say, Javascript clients.
+
+Nameko is also extensible. The main principals are agnotic to the way services communicate. You can define your own transport mechanisms and service dependencies to mix and match as desired.
```
davidszotten: principles

docs/what_is_nameko.rst
```diff
@@ -0,0 +1,38 @@
...
+    * You want to write your backend as microservices.
+    * You want to do it in Python.
+
+Nameko scales from a single instance of a single service, to a cluster with many instances of many different services.
+
+
+When shouldn't I use nameko?
+----------------------------
+
+Nameko is not a web framework. It has built-in HTTP entrypoints but they're limited to what is useful in the realm of microservices. If you want to build a webapp for consumption by humans you should use `flask <http://flask.pocoo.org>`_.
```
davidszotten: maybe _something like_ flask

docs/what_is_nameko.rst
```diff
@@ -0,0 +1,38 @@
...
+
+You shouldn't use nameko to write clients. The library comes with tools for clients to communicate with nameko clusters using the built-in AMQP entrypoints, but the client itself should be some other process.
```
davidszotten: clients? not clear what this means

docs/key_concepts.rst
davidszotten: possibly another sentence, or an example would make this clearer

docs/key_concepts.rst
```diff
@@ -0,0 +1,87 @@
...
+.. _dependency_injection:
+
+Dependency Injection
+--------------------
+
+Adding a dependency to a service class is declarative. That is, the attribute on the class is a declaration, rather than the interface that workers can actually use.
+
+The class attribute is a :class:`~nameko.extensions.DependencyProvider`. It is responsible for providing an object that is injected into service workers.
+
+DependencyProviders implement a :meth:`~nameko.extensions.DependencyProvider.get_dependency` method, the result of which is injected into a newly created worker.
```
davidszotten: "DependencyProviders". i guess either in backticks or as two words?

docs/key_concepts.rst
```diff
@@ -0,0 +1,87 @@
...
+
+The lifecycle of a worker is:
+
+    #. Entrypoint fires
+    #. Worker instantiated from service class
+    #. Dependencies injected into worker
+    #. Method executes
+    #. Worker is destroyed
+
+In code this looks something like::
```
davidszotten: maybe _pseudocode_

docs/key_concepts.rst
```diff
@@ -0,0 +1,87 @@
...
+
+    worker = Service()
+    worker.other_rpc = worker.other_rpc.get_dependency()
+    worker.method()
+    del worker
+
+DependencyProviders live for the duration of the service, whereas the injected dependency can be unique to each worker.
```
davidszotten: again, backtics or two words i think

docs/key_concepts.rst
```diff
@@ -0,0 +1,87 @@
...
+Concurrency
+-----------
+
+Nameko is built on top of the `eventlet <http://eventlet.net/>`_ library, which provides concurrency via "greenthreads". The concurrency model is co-routines with implicit yielding.
+
+Implicit yielding relies on `monkey patching <http://eventlet.net/doc/patching.html#monkeypatching-the-standard-library>`_ the standard library, to trigger a yield when a thread waits on I/O. If you host services with ``nameko run`` on the command line, nameko will apply the monkey patch for you.
+
+Each worker executes in its own greenthread. The maximum number of concurrent workers can be tweaked based on the amount of time each worker will spend waiting on I/O.
+
+Workers are stateless so are inherently thread safe, but dependencies should ensure they support thread-safety.
```
davidszotten: not sure about the phrase "support thread-safety".

docs/installation.rst
davidszotten: +1 for link to rabbit docs, but can also include basics on ox x/ubuntu (brew install / apt-get install)

docs/key_concepts.rst
junkafarian: surely "ensure they are thread-safe"

docs/cli.rst
```diff
@@ -0,0 +1,45 @@
...
+::
+
+    $ nameko shell
+
+Launch an interactive python shell for working with remote nameko services. This is a regular interactive interpreter, with a special module ``n`` added
+to the builtin namespace, providing the ability to make RPC calls and dispatch events.
+
+Making an RPC call to "target_service":
+
+::
```
davidszotten: don't know if pygments need a hand with guessing python-console here. if we're keeping the banner, maybe `emphasize-lines` (http://sphinx-doc.org/markup/code.html) could help

docs/built_in_extensions.rst
```diff
@@ -0,0 +1,47 @@
+Built-in Extensions
+===================
+
+RPC
+---
+
+Nameko has a built-in implementation for RPC over AMQP. It includes the ``@rpc`` entrypoint, a proxy for services to talk to other services, and a standalone proxy that clients can use to make RPC calls to a cluster:
```
davidszotten: that _non-nameko_ clients (possible "e.g. a django or flask app" or similar)

docs/examples/rpc.py
```diff
@@ -0,0 +1,20 @@
+from nameko.rpc import rpc, RpcProxy
+
+
+class ServiceY(object):
+    name = "service_y"
+
+    @rpc
+    def append_identifier(self, value):
```
davidszotten: don't love this as an example

docs/examples/standalone_rpc.py
```diff
@@ -0,0 +1,8 @@
+from nameko.standalone.rpc import ClusterRpcProxy
+
+config = {
+    'AMQP_URI': 'amqp://guest:guest@localhost/nameko'
+}
+
+with ClusterRpcProxy(config) as cluster_rpc:
+    cluster_rpc.service_x.remote_method("hello")  # "x-y-hello"
```
davidszotten: don't love this as an example

also, for this, where the caller example is tightly coupled to the service example, it might be better to keep them in the same file (e.g. here i think the response output has drifted out of sync with the method)

docs/examples/async_rpc.py
```diff
@@ -0,0 +1,6 @@
+with ClusterRpcProxy(config) as cluster_rpc:
+    hello_res = cluster_rpc.service_x.remote_method.async("hello")
+    world_res = cluster_rpc.service_x.remote_method.async("world")
+    # do work while waiting
+    hello_res.wait()  # "x-y-hello"
```
davidszotten: i wonder if we could re-write these as doctests and run them (with say a pytest hook that spins up the example services)

this should be  `res.result()`, not `wait()`

docs/built_in_extensions.rst
```diff
@@ -0,0 +1,47 @@
...
+
+.. literalinclude:: examples/rpc.py
+
+.. literalinclude:: examples/standalone_rpc.py
+
+Normal RPC calls block until the remote method completes, but proxies also have an "async" calling mode to background or parallelize RPC calls:
```
davidszotten: "async" -> asynchronous ?

docs/examples/events.py
```diff
@@ -0,0 +1,23 @@
+from nameko.events import EventDispatcher, event_handler
+from nameko.rpc import rpc
+
+class ServiceA(object):
+    """ Event dispatching service.
+    """
+    name = "service_a"
```
davidszotten: we use both x/y and a/b for service names. does this matter?

docs/examples/events.py
```diff
@@ -0,0 +1,23 @@
+from nameko.events import EventDispatcher, event_handler
+from nameko.rpc import rpc
+
+class ServiceA(object):
+    """ Event dispatching service.
```
davidszotten: nicer with trailing quotes on the same line?

docs/built_in_extensions.rst
```diff
@@ -0,0 +1,47 @@
...
+
+
+Events (Pub-Sub)
+----------------
+
+Nameko Events is an asynchronous messaging system, similar to the Publish-Subscriber pattern. Services dispatch events that may be received by zero or more others:
+
+.. literalinclude:: examples/events.py
+
+There is also a standalone dispatcher that can be used to imitate an event being dispatched from a service:
```
davidszotten: do we definitely want to document/promote this? if so, do we want to consider it "imitation"?

docs/built_in_extensions.rst
```diff
@@ -0,0 +1,47 @@
...
+
+.. literalinclude:: examples/standalone_events.py
+
+
+HTTP GET & POST
+---------------
+
+Nameko's HTTP entrypoint supports simple GET and POST:
+
```
davidszotten: probably need to explain that the default handler will serialize the return value as json (if we want to keep that as the default), and similarly, that post-data should be json (with content-type?)

docs/examples/advanced_http.py
```diff
@@ -0,0 +1,12 @@
...
+
+    @http('GET', '/google')
+    def redirect(self,):
+        return 302, "", {'Location': 'https://www.google.com/'}
```
davidszotten: with 3 headers, it's `status, headers, payload` (probably worth documenting; possibly along with the fact that this is built on top of werkzeug and also accepts a werkzeug response object for complete customisation)

davidszotten: though a 302 (and 404) is a slightly odd example.

docs/built_in_extensions.rst
```diff
@@ -0,0 +1,47 @@
...
+Service methods can return specific status codes and set headers:
+
+.. literalinclude:: examples/advanced_http.py
+
+
+Websockets
+----------
+
+Nameko's websocket implementation allows for RPC over a websocket and a facility to broadcast messages to subscribed listeners:
```
davidszotten: probably want to label this as "experimental" for now

docs/testing.rst
```diff
@@ -0,0 +1,78 @@
...
+----------
+
+Nameko's conventions are designed to make testing as easy as possible. Services are likely to be small and single-purpose, and dependency injection makes it simple to replace and isolate pieces of functionality.
+
+Unit Testing
+------------
+
+Unit testing in nameko usually means testing a single service in isolation - i.e. without any or most of its dependencies.
+
+Tne :func:`~nameko.testing.services.worker_factory` utility will create a worker from a given service class, with its dependencies replaced by :class:`mock.MagicMock` objects. Dependency functionality can then be imitated by adding side effects and return values:
```
davidszotten: maybe `MagicMock` objects from the mock(link) library

davidszotten: mention that examples use pytest(link) which is what nameko's own test suite is using, but all testing utilities provided are test framework agnostic

docs/examples/testing/unit_test.py
```diff
@@ -16,15 +18,9 @@ def inches_to_cm(self, inches):
...
-# =============================================================================
-# Begin test
-# =============================================================================
-
 
 def test_conversion_service():
-    # create instance
-    # dependencies are replaced with mocks unless otherwise given - see
-    # :class:`nameko.testing.services.worker_factory`
+    # create worker with mock dependencies
```
davidszotten: probably worth a comment/link to http://www.voidspace.org.uk/python/mock/mock.html#mock.Mock.side_effect

docs/testing.rst
```diff
@@ -0,0 +1,78 @@
...
+
+.. literalinclude:: examples/testing/unit_test.py
+
+In some circumstances it's helpful to provide an alternative dependency, rather than use a mock. This may be a fully functioning replacement (e.g. a test database session) or a lightweight shim that provides partial functionality.
```
davidszotten: if we're spinning out the sqla stuff we might need a new example here

docs/examples/testing/integration_test.py
```diff
@@ -19,25 +21,26 @@ def remote_method(self, value):
...
 
-def test_service_integration():
+def test_service_x_y_integration():
+
+    # run services in the normal manner
     config = {'AMQP_URI': 'amqp://guest:guest@localhost:5672/'}
     runner = ServiceRunner(config)
     runner.add_service(ServiceX)
     runner.add_service(ServiceY)
     runner.start()
```
davidszotten: probably worth making clear that we are using hooks to fire x, but the interaction between x and y (which we are testing) is just as if under normal operation

nameko/testing/services.py
```diff
@@ -286,7 +295,9 @@ def restrict_entrypoints(container, *entrypoints):
 
     **Usage**
 
-    The following service definition has two entrypoints for "method"::
+    The following service definition has two entrypoints for "bar":
```
davidszotten: i think disabling a `timer` entrypoint would make a good example

docs/testing.rst
```diff
@@ -0,0 +1,78 @@
...
+
+Note that the context manager waits not only for the entrypoint method to complete, but also for any dependency teardown. Dependency-based loggers such as (TODO: link to bundled logger) for example would have also completed.
+
+
+Barebones conftest.py
+---------------------
+
+Below is a barebones conftest for pytest that includes configuration and fixtures for communicating with RabbitMQ as well as the eventlet monkey patch. It is a good base for service tests if you're using pytest.
+
+.. literalinclude:: examples/testing/conftest.py
```
davidszotten: not completely sure, but don't think we need to monkeypatch greenpipe anymore

docs/api.rst
```diff
@@ -0,0 +1,7 @@
+API Reference
+=============
+
+.. toctree::
+   :maxdepth: 2
+
+   api/nameko
```
davidszotten: not sure we want to blanket include everything here. is there much benefit over e.g. just browsing the code?

docs/index.rst
```diff
@@ -1,13 +1,83 @@
...
+More Information
+----------------
+
+.. toctree::
+   :maxdepth: 2
+
+   about_microservices
+   dependency_injection_benefits
+   similar_projects
+   license
```
davidszotten: - getting in touch
- contributing

docs/api.rst
mattbennett: yeah let's drop the whole thing. we can add it back eventually with more prose around the various sections

docs/built_in_extensions.rst
mattbennett: Hmm. Probably best to drop it from the docs for now, at least this page.

docs/built_in_extensions.rst
mattbennett: Yep, I'm going to add a whole page on serialization and will link to it from here.

docs/cli.rst
mattbennett: Ah, wishful thinking. I'd like to change the discovery mechanism so this would be valid. Will make it right for now though

docs/examples/advanced_http.py
mattbennett: Better example than 302 for a status that includes a header in the server response?

docs/examples/async_rpc.py
mattbennett: Nice idea. I'll just fix the code for now though

docs/examples/events.py
mattbennett: Yep. Actually recommended by PEP8

docs/examples/events.py
mattbennett: I figured it was better to help distinguish between the arbitrary choices. a/b work together, as do x/y

docs/examples/http.py
mattbennett: I thought the values were unpacked, but it looks like I was wrong.

@davidszotten was this the original implementation?

docs/examples/standalone_rpc.py
mattbennett: Happy to use an alternative example if you can suggest one.

docs/what_is_nameko.rst
```diff
@@ -0,0 +1,38 @@
...
+
+Nameko scales from a single instance of a single service, to a cluster with many instances of many different services.
+
+
+When shouldn't I use nameko?
+----------------------------
+
+Nameko is not a web framework. It has built-in HTTP entrypoints but they're limited to what is useful in the realm of microservices. If you want to build a webapp for consumption by humans you should use something like `flask <http://flask.pocoo.org>`_.
+
+You shouldn't use nameko to write clients. The library comes with tools for clients to communicate with nameko clusters using the built-in AMQP entrypoints, but the client itself should be some other process.
```
shauns: Not really sure what you mean by 'client' here.

docs/index.rst
```diff
@@ -1,13 +1,83 @@
...
+----------
+
+This section covers most things you need to know to create and run your own nameko services.
+
+.. toctree::
+   :maxdepth: 2
+
+   what_is_nameko
+   key_concepts
+   installation
```
shauns: I'd put this way nearer the front, maybe even on the front page -- or at least the pip install

docs/index.rst
```diff
@@ -1,13 +1,83 @@
...
+   cli
+   built_in_extensions
```
shauns: "built-in" and "extension" seem mutually exclusive. I know extension is the internal name but maybe we can come up with something clearer given this is for fresh readers

docs/examples/testing/unit_test.py
```diff
@@ -16,15 +18,9 @@ def inches_to_cm(self, inches):
...
     service = worker_factory(ConversionService)
 
     # replace "math" service
```
shauns: confusing as to whether you're talking about the service you just made. worth clarifying this is mocking the injected proxy to a different service

docs/testing.rst
shauns: -1 on "firing". For beginners to the framework I think we should talk more specifically about making RPC calls/dispatching events/etc.

docs/testing.rst
```diff
@@ -0,0 +1,78 @@
...
+replace_injections
+^^^^^^^^^^^^^^^^^^
+
+.. autofunction:: nameko.testing.services.replace_injections
+   :noindex:
+
+Complete Example
+^^^^^^^^^^^^^^^^
+
+The following integration testing example makes use of both scope-restricting helpers:
```
shauns: This example is huge and blocks the other sections below.

nameko/extensions.py
```diff
@@ -138,19 +138,20 @@ def bind(self, container, attr_name):
...
 
-    def acquire_injection(self, worker_ctx):
-        """ Called before worker execution. A Dependency should return
+    def get_dependency(self, worker_ctx):
+        """ Called before worker execution. A DependencyProvider should return
         an object to be injected into the worker instance by the container.
         """
 
     def inject(self, worker_ctx):
+        """ TODO when we have better parallization than spawningset,
```
shauns: Spelling

davidszotten:
note that merging the spelling branch invalidated a bunch of comments

docs/cli.rst
shauns: Too late maybe more `module.Class` is more common. https://docs.djangoproject.com/en/1.7/ref/settings/

docs/cli.rst
```diff
@@ -0,0 +1,45 @@
...
+Nameko ships with a command line interface to make hosting and interacting with services as easy as possible.
+
+Running a Service
+-----------------
+
+::
+
+    $ nameko run [<module>[:<ServiceClass>]]
+
+Discover and run a service class.
```
shauns: This isn't enough for me to know how to use Nameko in production. Which might be fine! But its definitely missing.

docs/cli.rst
```diff
@@ -0,0 +1,45 @@
...
+
+
+Interacting with a running service
```
shauns: plural

docs/examples/advanced_http.py
shauns: 201

docs/built_in_extensions.rst
shauns: Future is pretty standard terminology for what you get out of `async`, think it should be referenced.

docs/cli.rst
```diff
@@ -0,0 +1,45 @@
...
+----------------------------------
+
+::
+
+    $ nameko shell
+
+Launch an interactive python shell for working with remote nameko services. This is a regular interactive interpreter, with a special module ``n`` added
```
shauns: Think this is a prime spot for n as a docstring

docs/index.rst
shauns: release notes/changelog

docs/key_concepts.rst
```diff
@@ -0,0 +1,87 @@
...
+Methods are exposed to the outside world with :ref:`entrypoint <entrypoints>` decorators.
+
+.. literalinclude:: examples/example.py
+
+.. _entrypoints:
+
+Entrypoints
+^^^^^^^^^^^
+
+Entrypoints are gateways into the service methods they decorate. While a service is running, an entrypoint may "fire" and the decorated method would be executed by a service :ref:`worker <workers>`.
```
shauns: Bit opaque at this point in the docs. Worth mentioning what an entrypoint normally is i.e. something responding to an RPC call or handling an event.

docs/key_concepts.rst
mattbennett: :+1: for typography critique

docs/built_in_extensions.rst
marbru: And/or maybe add an extra code block with a curl command exemplifying the request-response.

docs/key_concepts.rst
```diff
@@ -0,0 +1,83 @@
...
+This section introduces nameko's central concepts.
+
+Anatomy of a Service
+--------------------
+
+A nameko service is just a Python class. The class encapsulates the application logic in its methods and declares any :ref:`dependencies <dependencies>` as attributes.
+
+Methods are exposed to the outside world with :ref:`entrypoint <entrypoints>` decorators.
+
+.. literalinclude:: examples/example.py
```
marbru: I think it would help if you could extend this example with another block of code showing a standalone rpc proxy calling the rpc method, and then you can add comments pinpointing "this is a dependency", "this is an entrypoint", "here we are instatiating a worker". I think it helps to see each of the concepts "in action".

docs/key_concepts.rst
```diff
@@ -0,0 +1,83 @@
...
+.. _dependencies:
+
+Dependencies
+^^^^^^^^^^^^
+
+Most services depend on something other than themselves. Nameko encourages these things to be implemented as dependencies.
+
+A dependency is an opportunity to hide code that isn't part of the core service logic. The dependency's interface to the service should be as simple as possible.
+
+Declaring dependencies in your service is a good idea for :ref:`lots of reasons <benefits_of_dependency_injection>`, and you should think of them as the gateway between service code and everything else. That includes other services, external APIs, and even databases.
```
marbru: It maybe appropriate to mention at some point in this document something like "Nameko includes a bunch of built-in entrypoints and dependencies" (and then a link to the built-in extensions) "or you can make your own" (and the the link to writing dependencies). 

This is because I found a bit confusing that the section describing the built-in stuff is calls it "extensions", when that term isn't mentioned anywhere in this terminology doc. Including the links from here would help make the connection IMO.

nameko/contrib/sqlalchemy.py
```diff
@@ -3,12 +3,12 @@
...
 from sqlalchemy.orm import sessionmaker
 
-from nameko.extensions import Dependency
+from nameko.extensions import DependencyProvider
 
 ORM_DB_URIS_KEY = 'ORM_DB_URIS'
 
 
-class OrmSession(Dependency):
+class OrmSession(DependencyProvider):
```
marbru: I'm surprised this OrmSession is not listed on the built-in extensions section.

nameko/testing/services.py
```diff
@@ -286,30 +299,32 @@ def restrict_entrypoints(container, *entrypoints):
...
-    To disable the entrypoints other than on "foo"::
-
-        restrict_entrypoints(container, "foo")
+    To disable the timer entrypoint on ``foo``, leaving just the RPC
+    entrypoint on ``bar``:
 
-    To maintain both the rpc and the event_handler entrypoints on "bar"::
+    .. code-block:: python
 
         restrict_entrypoints(container, "bar")
```
marbru: You can also provide a list of entrypoint names, should you want to keep several, is this right? Might be worth mentioning/showing an example in this docstring.

docs/examples/testing/integration_test.py
```diff
@@ -19,25 +21,26 @@ def remote_method(self, value):
...
 
+    # manually fire the "remote_method" entrypoint and verify response
     container = get_container(runner, ServiceX)
```
marbru: I found it a bit confusing that in every example in the integration test section, we are obtaining the container in a different manner (`get_container` here, `container_factory` on restrict_entrypoints, `ServiceContainer` on replace_injections). 

Also, as a matter of fact, nowhere else in the docs is mentioned what a container is, which adds to the confusion.

docs/testing.rst
mattbennett: I take your point about "firing", but that's why it's in quotes. It's tricky because we're actually *not* making RPC calls or dispatching events, instead triggering the action that would happen as if we had.

docs/built_in_extensions.rst
mattbennett: I want to avoid future because we don't match the ``concurrent.futures`` API

docs/built_in_extensions.rst
mattbennett: Thanks @marbru. I'll add this when we finalise the http behaviour.

docs/what_is_nameko.rst
mattbennett: Yeah this is a bungled explanation of the mistake we made with nameko-chat. I think I'll drop the whole paragraph.

docs/testing.rst
mattbennett: Yeah, I will probably split this into different sections so it isn't so imposing.

docs/cli.rst
mattbennett: Yeah there's a whole missing section on deployment or possibly a "service development guide" describing some of the patterns we use in platform (that aren't enforced but we think are sensible)

docs/testing.rst
```diff
@@ -0,0 +1,80 @@
...
+
+.. literalinclude:: examples/testing/large_integration_test.py
+
+Other Helpers
+-------------
```
marbru: Don't really get why these helpers are in a separate section, when they also only apply to integration testing.

nameko/contrib/sqlalchemy.py
mattbennett: Yep this is confusing. It's about to get moved out into a separate codebase as a "community extension". Hopefully that will also help clarify the whole extension/entrypoint/dependency confusion.

nameko/testing/services.py
mattbennett: Good point, will add

docs/examples/testing/entrypoint_waiter.py
```diff
@@ -0,0 +1,29 @@
...
+
+def test_event_interface(rabbit_config):
+
+    container = ServiceContainer(ServiceB, rabbit_config)
+    container.start()
+
+    dispatch = event_dispatcher(rabbit_config)
+
+    # prints "service b received payload" before "exited"
+    with entrypoint_waiter(container, 'hello'):
```
marbru: Should this not be 'handle_event' instead of 'hello'? Or else I don't get it.

docs/cli.rst
davidszotten: this follows e.g. gunicorn http://docs.gunicorn.org/en/latest/run.html

docs/examples/testing/integration_test.py
mattbennett: I've added a section to the "key concepts" section introducing the container. I'll also tidy up these examples so we don't use so many different utilities.

The difference BTW is that ``get_container`` extracts an existing container out of a ``ServiceRunner`` (which is just a wrapper around multiple container) and ``ServiceContainer`` actually constructs a container. ``container_factory`` is a test fixture that also constructs a new container.

docs/testing.rst
mattbennett: They're not strictly for integration testing, just useful in that context. Putting them under the integration testing heading suggests that you shouldn't use them elsewhere, which isn't the case.

mattbennett:
Re-raised against 2.0 branch in https://github.com/onefinestay/nameko/pull/207

PR #201, opened 2015-02-12
better run discovery

davidszotten:


nameko/cli/run.py
```diff
@@ -52,17 +63,34 @@ def import_service(module_name):
...
+            if inspect.getmembers(potential_service, is_entrypoint):
+                found_services.append(potential_service)
+
+        if not found_services:
+            raise CommandError(
+                "Failed to find anything that looks like a service in module "
+                "{!r}".format(module_name)
+            )
+
+    else:
```
mattbennett: You can't specify multiple classes? e.g. ``nameko run foo:Foo bar:Bar``

nameko/cli/run.py
```diff
@@ -1,14 +1,16 @@
 """Run a nameko service.
```
mattbennett: No longer true

nameko/cli/run.py
```diff
@@ -1,14 +1,16 @@
...
 
-Given a python path to a module containing a nameko service, will host and run
-it. By default this assumes a service class named ``Service``, but this can be
-provided via ``nameko run module:ServiceClass``.
+Given a python path to a module containing one or more nameko services, will
+host and run them. By default this will try to find "service-looking" classes
```
mattbennett: By default it will find classes that look like services (anything with nameko entrypoints), ...

nameko/cli/run.py
mattbennett: Could do with some usage examples in here.

nameko/cli/run.py
```diff
@@ -131,10 +159,10 @@ def main(args):
     if '.' not in sys.path:
         sys.path.insert(0, '.')
 
-    cls = import_service(args.service)
+    services = import_service(args.service)
```
mattbennett: Why is this now ``services``?

davidszotten: why not?

mattbennett:
:+1:

PR #215, opened 2015-02-27
Python3

davidszotten:
experimental python 3 support


noisyboiler:
is this using the new async library?

davidszotten:
lol, no. just eventlet py3 support

noisyboiler:
@mattbennett @davidszotten 
I'd like to kick off an asyncio branch. This is mostly for selfish reasons such as exploring this new library and figuring out how Nameko ticks. It may get no further than a proof of concept but it sounds like tremendous fun to me.

mattbennett:
This is tremendous.

Where are you up to with https://github.com/davidszotten/rabbitman? nameko/testing/rabbit.py is nice enough but it would be better to delegate to a separate library I think.

@noisyboiler asyncio support would be a big and fundamental change but you're welcome to give it a crack! A more reasonable goal might be to support regular threads as well as greenthreads, which probably adds value without costing too much (shouldn't cost a service developer anything). We could perhaps become fully concurrency model agnostic (i.e. including asyncio) but I think it will be hard, and perhaps not that useful.

davidszotten:
i've been going back and forth on rabbitman. this (`rabbit.py`) is where i landed for now, but happy to discuss again

mattbennett:
:+1:

mattbennett:
:+1: :+1:

PR #216, opened 2015-03-03
Support for stashing sensitive variables on Rpc and EventHandler

mattbennett:
entrypoints


nameko/events.py
```diff
@@ -101,64 +101,77 @@ def dispatch(event_type, event_data):
...
+                                    [queue]- (service Y handler-meth)
+
+            requeue_on_error : bool  # TODO: defined by Consumer actually..
+                If true, handlers will return the event to the queue if an
+                error occurs while handling it. Defaults to False.
+            reliable_delivery : bool
+                Is true, events will be held in the queue until there is a
+                handler to consume them. Defaults to True.
+            sensitive_variables : string or tuple of strings
+                Mark an argument or part of an argument as sensitive. Stashed
```
davidszotten: "stashed" sounds a bit colloquial.

nameko/events.py
```diff
@@ -101,64 +101,77 @@ def dispatch(event_type, event_data):
...
+                                  \
+                                    [queue]- (service Y handler-meth)
+
+            requeue_on_error : bool  # TODO: defined by Consumer actually..
+                If true, handlers will return the event to the queue if an
+                error occurs while handling it. Defaults to False.
+            reliable_delivery : bool
+                Is true, events will be held in the queue until there is a
+                handler to consume them. Defaults to True.
+            sensitive_variables : string or tuple of strings
```
davidszotten: do we definitely want to allow both a single or a tuple of strings?

nameko/utils.py
```diff
@@ -1,9 +1,108 @@
...
+    method = getattr(entrypoint.container.service_cls, entrypoint.method_name)
+    argspec = inspect.getcallargs(method, None, *args, **kwargs)
+    del argspec['self']
+
+    def redact(data, keys):
+        key = keys[0]
+        if len(keys) == 1:
+            try:
+                data[key] = REDACTED
+            except (KeyError, IndexError):
```
davidszotten: you can also get `TypeError` here if the argument doesn't match (e.g. trying to redact an integer thinking it's a dict)

test/test_utils.py
```diff
@@ -50,3 +54,107 @@ def test_repr_safe_str(value, repr_safe_value):
     res = repr_safe_str(value)
     assert res == repr_safe_value
     assert isinstance(res, bytes)
+
+
+class TestGetRedactedArgs(object):
+
+    @pytest.mark.parametrize("sensitive_variables, expected", [
+        (tuple(), {'a': 'A', 'b': 'B'}),  # no sensitive variables
```
davidszotten: `()` is an empty tuple

nameko/events.py
mattbennett: Perhaps not. I was matching the interface of ``expected_exceptions``.  A common usecase would be something like:

``` python
class Service(object):
    @rpc(sensitive_variables=("foo",), expected_exceptions=FooError)
    def method(self, foo, bar):
        pass
```

The parenthesis seem slightly superfluous given we don't require them for ``expected_exceptions``.

nameko/utils.py
mattbennett: Good catch

test/test_utils.py
mattbennett: I find this more obvious -- easier to distinguish it amongst all the other parenthesis. Happy to change if you'd prefer though.

davidszotten:
:+1: though we might want to hold of on landing this as part of 2.0

mattbennett:
Re-raised against master -- https://github.com/onefinestay/nameko/pull/253

PR #217, opened 2015-03-10
require string responses for http

davidszotten:


davidszotten:
raised against wrong base

PR #219, opened 2015-03-14
http improvement suggestions

kollhof:
This PR should serves as a discussion, and should probably be split into two. 

configurable base URL for http server
=============================

Lets say I have a service `foobar` with a couple of URLs it serves `/spam` and  `/ni`.
I would like to serve this service at http://example.org/foobar or maybe at the root http://foobar.example.org
without having to change any of my application code. I may even choose to deploy multiple versions of the service e.g. http://example.org/foobar/v1 and http://example.org/foobar/v2
The application must be able to correctly generate absolute URLs for it's own. This may be required if the generated URLs are passed back to callers (html pages, service discovery URLs, ...)

My implementation added a config `WEB_SERVER_BASE_URL` which defaults to `''`.
I have not touched the websocket implementation, I need to do some reading first. I believe it should be possible to configure the URLs similarly via the config. I also think the service itself should be able to specify under which URL e.g. RPC should be served (instead of locking it to `/ws`). 

e.g.
```python

class Foobar(object):
    @http('GET', '/spam')
    def spam(self): 
            return 'check out ' + self.url_for(self.ni, 5)

    @http('GET', '/ni/<int:ham>')
    def ni(self, ham):
            ...
``` 
Running `Foobar` with defaults would serve `spam` and `ni` at the root of the domain. Running it with `WEB_SERVER_BASE_URL='/foobar/v1'` would serve `spam` and `ni` at `/foobar/v1/spam` and `/foobar/v1/ni` respectively.  

Raw Request object for http handlers
============================

It is more likely than not that I may want to access data from the incoming HTTP request, i.e. more than just the URL parameters/path segments and POST data. I also think it should be possible to parameterize the URL for a POST as I would for the GET and they should be handled consistently.

```python
@http('GET' , '/foo/<int:bar>')
def handle(self, req, bar):
    ...

@http('POST' , '/spam/<int:ham>')
def handle(self, req, ham):
    ...
```
it would also be consistent when creating hybird handlers like:
```python
@http(('GET', 'POST') , '/spam/<int:ham>')
def handle(self, req, ham):
    ...
```

I suggest to always pass the raw werkzeug `Request`  to the handler methods as first param after `self`, followed by all URL params. This is little overhead for any handler that does not need it, would be consistent across http-methods (GET, POST, ...) and would allow greater flexibility with handling incoming requests.

Another way of handling would be to make the req optional e.g.:
```python
@http('GET' , '/foo/<int:bar>', req=True)
def handle(self, bar, req):
    ...
```
or to make the request object available as an injection(don't very much like that).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,3 +23,9 @@ spelling:
 test_docs:
 	spelling
 	docs
+
+install-dev:
+	pip install -e .[develop]
+
+install:
+	pip install -e .
--- a/nameko/web/handlers.py
+++ b/nameko/web/handlers.py
@@ -41,13 +41,13 @@ def response_from_result(result):
 class HttpRequestHandler(Entrypoint):
     server = WebServer()
 
-    def __init__(self, method, url, expected_exceptions=()):
+    def __init__(self, method, url, pass_req=False, expected_exceptions=()):
         self.method = method
         self.url = url
         self.expected_exceptions = expected_exceptions
 
-    def get_url_rule(self):
-        return Rule(self.url, methods=[self.method])
+    def get_url_rule(self, base_url):
+        return Rule(base_url + self.url, methods=[self.method])
 
     def setup(self):
         self.server.register_provider(self)
@@ -57,12 +57,7 @@ class HttpRequestHandler(Entrypoint):
         super(HttpRequestHandler, self).stop()
 
     def process_request_data(self, request):
-        if request.method == 'POST':
-            data = request.get_data()
-            args = (data,)
-        else:
-            args = ()
-
+        args = (request,)
         kwargs = request.path_values
         return args, kwargs
 
--- a/nameko/web/server.py
+++ b/nameko/web/server.py
@@ -14,6 +14,7 @@ from nameko.extensions import ProviderCollector, SharedExtension
 
 WEB_SERVER_HOST_CONFIG_KEY = 'WEB_SERVER_HOST'
 WEB_SERVER_PORT_CONFIG_KEY = 'WEB_SERVER_PORT'
+WEB_SERVER_BASEURL_CONFIG_KEY = 'WEB_SERVER_BASE_URL'
 
 
 class HttpOnlyProtocol(HttpProtocol):
@@ -81,8 +82,9 @@ class WebServer(ProviderCollector, SharedExtension):
 
     def make_url_map(self):
         url_map = Map()
+        base_url = self.container.config.get(WEB_SERVER_BASEURL_CONFIG_KEY, '')
         for provider in self._providers:
-            rule = provider.get_url_rule()
+            rule = provider.get_url_rule(base_url)
             rule.endpoint = provider
             url_map.add(rule)
         return url_map
--- a/nameko/web/websocket.py
+++ b/nameko/web/websocket.py
@@ -55,7 +55,7 @@ class WebSocketServer(SharedExtension, ProviderCollector):
             'data': data,
         })
 
-    def get_url_rule(self):
+    def get_url_rule(self, base_url):
         return Rule('/ws', methods=['GET'])
 
     def handle_request(self, request):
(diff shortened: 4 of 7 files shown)


davidszotten:
`base_url`:
this is moving into the territory of "web framework" which isn't really the sweet spot for nameko

davidszotten:
`request` as first arg: I think the current default makes sense, again for the very basic use-case we have in mind. we could do a better job of promoting and giving examples of subclassing our providers to change their behaviour, and this might be a good example of that.

kollhof:
You kinda already moved into the territory of "web framework" by providing a http extension.
I would argue even for a simple web-api it is important to separate the application logic and thus the component's URL space from hosting of the service, i.e the domain and base URL it is served from. (You would not specify the domain and port in your `@http` as you also should not specify the root). 
It is very little overhead and the default would still work out of the box.

With respect to GET and POST handling, I think `@http` is and interesting proof of concept at the moment but it is starting off too high level. Nameko's events evolved from messaging. Messaging gave us low level tools and allowed us to figure out where we wanted to go with the next level of abstraction. 
I think a similar approach would be suitable for the http implementations, start with basic support for URL routing and req, resp, then simplify and abstract further. This would allow us to experiment with everything the werkzeug request and response objects give, but within the nameko framework and runtime. 

Now about using nameko as a web-framework. Lets say I have an easy way to deploy and manage nameko services locally and in a production environment. I have CI and CD figured out. I have also invested into logging, monitoring and debugging nameko applications.  And, I have a team who knows their way around nameko. Let's also assume I have heavily invested into building my own extensions.
Why would I want to introduce yet another framework with different deployment mechanisms, different ways to manage config, monitoring, logging, and different dev approaches?  
I may have to build new libraries to use the tools I have built for nameko, or at least refactor them to be able to use them outside of nameko etc. I then need to integrate the two systems e.g. via nameko.standalone or some other ways.
I would rather spend my resources on making nameko.web work for me and have one consistent application framework to worry about. 
Just a thought.

mattbennett:
I like the base URL suggestion, although @davidszotten has suggested merging it with the existing ``WEB_SERVER_HOST`` and ``WEB_SERVER_PORT`` config keys to make a ``WEB_SERVER_URI`` setting.

On nameko not being a "web framework", the intention is to avoid competing (in the minds of the community) with much more established solutions like flask. We don't want to reinvent that wheel and we don't want people to think we're trying to either. But @kollhof raises a good point about having a consistent framework and making use of existing investment in nameko services; ultimately that's why we added HTTP support in the first place. And lots, if not the majority of people, use HTTP as the transport between microservices.

In the name of pragmatism, and the fact that @kollhof's is the first concrete use-case we have for HTTP, we should be open-minded about the "web framework" label. It's probably sufficient to say loudly at the outset that we don't include all the bells-and-whistles that you'd expect of a fully-featured web framework. in the box. It's quite possible that someone will write them anyway; the question is how much we should provide from the outset.

I think for the "built-in extensions" we should strive to:
* provide for the 80% case
* with a lightweight implementation (no '000 line modules, rely on external libraries to do the heavy lifting)
* not get in the way of extending to the full 100%
* provide best-practice templates for others who want to create extensions

For HTTP the two difficult points are: identifying the 80% use-case, and the notion of best-practise for extensions. With AMQP messaging, we consciously don't give the service developer access to the message object, only the payload. Headers are available via a dependency. I think what we're achieving there is separation between the *service code* and the *transport*. And obviously Events and RPC are even higher abstractions.

It's a bit murkier with HTTP. The werkzeug Request object gives you access to form data, headers, cookies and file uploads. Requiring a dependency to access each of those seems a bit dogmatic, but the request object gives services lower-level access than we've made available elsewhere. I don't like the idea of service code looping over chunks of uploaded files, for example.

None of this precludes anyone writing their own extensions, of course. But back to the 80% case, it would be a bit silly if everyone basically built something like the above because it wasn't included out of the box.

Perhaps I'm agonising about it too much and the built-in version should just be nice and useable. Either that or provide a low-level base and a sensible abstraction on top, much as we did with AMQP messaging and then Events.

I like the parameterized POST URLs and hybrid handlers too.

mattbennett:
Implemented in https://github.com/onefinestay/nameko/pull/224 and https://github.com/onefinestay/nameko/pull/225. Thanks @kollhof for the suggestions :)

PR #222, opened 2015-03-23
Mandatory service name

mattbennett:
Mostly uninteresting except the changes to ``get_service_name`` and the third commit

### DIFF
--- a/nameko/cli/run.py
+++ b/nameko/cli/run.py
@@ -16,7 +16,7 @@ import sys
 from eventlet import backdoor
 
 from nameko.constants import AMQP_URI_CONFIG_KEY
-from nameko.exceptions import CommandError
+from nameko.exceptions import CommandError, ConfigurationError
 from nameko.extensions import ENTRYPOINT_EXTENSIONS_ATTR
 from nameko.runners import ServiceRunner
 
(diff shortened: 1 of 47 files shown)


nameko/containers.py
```diff
@@ -29,7 +29,12 @@
 
 
 def get_service_name(service_cls):
-    return getattr(service_cls, "name", service_cls.__name__.lower())
+    service_name = getattr(service_cls, 'name', None)
+    if service_name is None:
+        raise RuntimeError('Service class must define a `name` attribute')
```
davidszotten: do we want some custom `ConfigurationError`? (e.g. to catch and display something nicer than a traceback in `nameko run`) or are we happy with a traceback in this case?

test/test_container.py
```diff
@@ -527,3 +527,30 @@ def test_stop_during_kill(container, logger):
...
         ]
+
+
+def test_get_service_name():
+
+    class Service():
+        name = "str"
+
+    class UnicodeService():
+        name = u"unicode"
```
davidszotten: could we have some non-ascii chars in here for good measure?

davidszotten:
at least `index.rst` also needs updating. may want e.g. `hello_service` or some other `foo_service` as the name (as opposed to `hello_world`; c.f. hello world example)

nameko/cli/run.py
```diff
@@ -110,7 +110,13 @@ def _bad_call():
 def run(services, config, backdoor_port=None):
     service_runner = ServiceRunner(config)
     for service_cls in services:
-        service_runner.add_service(service_cls)
+        try:
```
davidszotten: don't think this is the right place to catch these (`run` may be used from other libraries)

`nameko/cli/main.py` is probably the place (already catches and prints `CommandError`s)

mattbennett: Quite right, thanks

davidszotten:
:+1:

PR #224, opened 2015-03-25
single config key for web extension

davidszotten:


nameko/web/server.py
```diff
@@ -47,10 +64,8 @@ def __init__(self):
 
     @property
     def bind_addr(self):
-        return (
-            self.container.config.get(WEB_SERVER_HOST_CONFIG_KEY, ''),
-            self.container.config.get(WEB_SERVER_PORT_CONFIG_KEY, 8000),
-        )
+        address_str = self.container.config.get(WEB_SERVER_CONFIG_KEY, '8000')
```
kollhof: `8000` looks odd as a fallback value, wouldn't `localhost:8000` or `0.0.0.0:8000` be a better default in terms of readability/understandability of the code?

mattbennett: Eventlet will bind to ``0.0.0.0`` if you pass an empty string for the address, so ``0.0.0.0:8000`` would be explicit.

``` python
>>> sock = eventlet.listen(('', 8000))
>>> sock.getsockname()
... ('0.0.0.0', 8000)
```

mattbennett:
:+1: pending @kollhof's suggestion

PR #225, opened 2015-03-25
http entrypoint to take a requset arg

davidszotten:


mattbennett:
This is fine. I'd like to add something to the docs explaining that it's intentionally low-level and should probably be subclassed, but that can wait for another PR.

mattbennett:
:+1:

Issue #233, opened 2015-04-10
By default, rabbitmq doesn't allow nameko to connect


vtbassmatt:
At least on Windows (haven't tested elsewhere), rabbitmq's default installation denies guest access to the nameko vhost. This means that the Nameko docs are slightly misleading - there is one configuration step required before anything works :)

As administrator, run `rabbitmqctl add_vhost nameko` followed by `rabbitmqctl set_permissions -p nameko guest ".*" ".*" ".*"` in order to address the issue. I'll create a PR against the docs.

mattbennett:
@vtbassmatt thanks for the bug report! This is addressed in https://github.com/onefinestay/nameko/pull/237, and we tried to make the error message friendlier too - https://github.com/onefinestay/nameko/pull/240

vtbassmatt:
Great, thanks!

PR #234, opened 2015-04-10
Update installation.rst for Windows

vtbassmatt:
Address #233 with Windows-specific instructions.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -38,4 +38,9 @@ On debian-based operating systems::
 
 For other platforms, consult the `RabbitMQ installation guidelines <https://www.rabbitmq.com/download.html>`_.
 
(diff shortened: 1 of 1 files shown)


davidszotten:
hey, nice catch. though i wonder if we should consider changing our defaults to use the default rabbit vhost (`/`) instead.

vtbassmatt:
Seems like that would work too. I'm new to rabbitmq so don't know what the community most expects.

Sent from Outlook for iPhone




On Fri, Apr 10, 2015 at 3:27 PM -0700, "David Szotten" <notifications@github.com> wrote:












hey, nice catch. though i wonder if we should consider changing our defaults to use the default rabbit vhost (/) instead.

mattbennett:
Addressed by https://github.com/onefinestay/nameko/pull/237. Thanks!

PR #237, opened 2015-04-11
Use default vhost

mattbennett:


nameko/messaging.py
```diff
@@ -350,8 +350,11 @@ def on_iteration(self):
             self.should_stop = True
 
     def on_connection_error(self, exc, interval):
-        _log.warn('broker connection error: {}. '
-                  'Retrying in {} seconds.'.format(exc, interval))
+        _log.warn(
+            "Error connecting to broker at {} ({}).\n"
+            "Is the broker running? "
+            "Do you have permission to access to the vhost?\n"
+            "Retrying in {} seconds.".format(self.amqp_uri, exc, interval))
```
mattbennett: Re https://github.com/onefinestay/nameko/issues/236, I think this is as good as we can do. Kombu doesn't give you anything more specific than "socket closed"

davidszotten: rather that doing this here (i.e. on all connection errors), maybe we could do a connection-check, e.g. in `setup`

it doesn't look like rabbit is of much help (e.g. bad creds result in a pause when connecting, before a disconnect)

pika tries to track how far it gets during connection and gives heuristic error messages, but amqp doesn't help very much. best i could do was something like this

```
from amqp import Connection

class ConnectionTest(Connection):
    def __init__(self, *a, **k):
        try:
            super(ConnectionTest, self).__init__(*a, **k)
        except IOError as exc:
            if not hasattr(self, '_wait_tune_ok'):
                raise
            elif self._wait_tune_ok:
                print "probably bad creds"
            else:
                print "probably bad vhost"
```

nameko/cli/run.py
```diff
@@ -178,7 +178,7 @@ def init_parser(parser):
         metavar='module[:service class]',
         help='python path to one or more service classes to run')
     parser.add_argument(
-        '--broker', default='amqp://guest:guest@localhost:5672/nameko',
```
davidszotten: i guess these are technically backwards incompatible

nameko/cli/shell.py
```diff
@@ -15,7 +15,7 @@
 
 def init_parser(parser):
     parser.add_argument(
-        '--broker', default='amqp://guest:guest@localhost:5672/nameko',
+        '--broker', default='amqp://guest:guest@localhost:5672/',
```
davidszotten: single slash on purpose? (which i think makes vhost `None` meaning use amqp lib defaults)

we could use more defaults by having our default be just `amqp://localhost`

davidszotten:
could possibly decouple "new default" from "better error reporting"

nameko/messaging.py
mattbennett: I like this idea. I'll back it out for a separate PR though

davidszotten:
:+1: 
changes/version bump?

PR #240, opened 2015-04-13
check amqp uri for better error messages

davidszotten:


nameko/messaging.py
```diff
@@ -65,6 +66,58 @@ def unpack_message_headers(self, worker_ctx_cls, message):
...
+        'password': conninfo.password,
+        'login_method': conninfo.login_method,
+        'virtual_host': conninfo.virtual_host,
+        'insist': conninfo.insist,
+        'ssl': conninfo.ssl,
+        'connect_timeout': conninfo.connect_timeout,
+        'heartbeat': conninfo.heartbeat,
+    }, **conninfo.transport_options or {})
+    with ConnectionTester(**opts):
+        pass
```
mattbennett: Is it possible to do this with a subclass, to avoid lifting this code? e.g.

``` python

class TransportTester():
    Connection = ConnectionTester

def verify_amqp_uri(uri):
    conn = Connection(uri)
    transport = TransportTester(conn.transport.client)
    transport.establish_connection()
```

nameko/amqp.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class TestTransport(Transport):
+    Connection = ConnectionTester
+
+
+def verify_amqp_uri(amqp_uri):
+    connection = Connection(amqp_uri)
+    if connection.transport_cls != 'amqp':
+        return
```
mattbennett: I think this is leftover from the kombu lift. Probably shouldn't get a free pass if not AMQP

mattbennett:
:+1: nice

mattbennett:
:+1:

PR #244, opened 2015-04-20
Kombu serialization support

ayoshi:
Hi.

I had a spare time on hand, and decided to take a stab at it. It proved a bit harder than I expected.
Let me know if there are any glaring omissions.

All tests pass correctly with default being JSON. I didn't add any tests for alternative serializers.
If the code looks right , I'l try to add tests for other serialization format, pickle for example.


mattbennett:
Thanks @ayoshi for contributing! I think you also need to specify the ``accept`` kwarg to the ``Consumer`` constructor (on https://github.com/onefinestay/nameko/blob/master/nameko/messaging.py#L336)

A test that demonstrated registering and using custom serializer would be great.

ayoshi:
@mattbennett 
Hi.

I've done a second iteration, and the problem is much hairier than it looks.

Basically, kombu's behavior when you pass serializer is different from when you simply pass ```json``` as a serializer to a publisher and ```accept``` to a consumer.

In case no values are passed, it doesn't care about message type, and accepts them silently, defaulting to 'text' if string is passed and 'data' if it sees binary data. In such cases it will NOT serialize them at all (This is what happens in the current Nameko version). If you pass ```accept``` however, It refuses to accept any other types and encodings. 

This breaks some of the tests which rely on passing unserialized data (test_serialization.py, test_rpc_responder.py) but those are easy to fix, I did it in one of the commits, creating parametrized fixtures to test both json and pickle.

There is a bigger problem, however:

In unit tests in many places the Container is mocked with ```empty_config``` fixture. Since the serializer/accept values will be set to None in this case, kombu will refuse to decode the message, the moment ```accept``` list is empty.
During normal lifecycle this doesn't happen in, since I use Container. __init__ to set up the serializer correctly for the container.

This behavior breaks many tests for py27-oldest tox environment (kombu-3.0.1)

It seems the behavior changes somewhere after kombu-3.0.1 (I didn't check when exactly) - if None is passed, kombu will revert to default behavior - so all the other versions pass all the unit tests.

It might be that I missed something else, I'm not familiar with all the details of behavior yet.

One option would be to refactor empty_config fixture to use DEFAULT_KOMBU_SERIALIZER.
That would fix the tests, and probably is the correct way to do things - this will make all messages during the test serialized, and more importantly will set ```accept``` properly for test Consumers

The other would be to add None/None Content types to the default accept list, but IMHO this will just hide improper behavior - I think every message in the system probably should be serialized.

I'm submitting what I already have into this pull request for you to review - let me know how you would like it worked around.

Otherwise - all the test work in other versions ( I tested by setting DEFAULT_KOMBU_SERIALIZER ) to ```pickle```

ayoshi:
After I can fix this py27 - I'll refactor the commit to be more manageable - and I can also add parametrized fixture over ```container_factory``` fixture to run all the tests both with json and pickle.

mattbennett:
Hi @ayoshi,

I confess I spent an hour or two on this over the weekend (before you raised this PR) and discovered that   by default kombu was not serializing to json if the data type was text or bytes. I agree with you that every message should be explicitly serialized, rather than letting kombu decide.

I'll have a look at the error in kombu 3.0.1, but I'd probably prefer bumping the minimum version to working around the inconsistency in nameko.

mattbennett:
The kombu commit that changes the behaviour is https://github.com/celery/kombu/commit/83cc99eca481ac35c23b52c8d293fe9f79dd8dc5, which landed in v3.0.11.

I definitely think the correct thing to do is always specify a serializer, and avoid kombu's logic that sometimes doesn't bother encoding. 

The failing tests all use the ``rabbit_manager`` fixture to publish a message, which does not set ``content_type`` or ``content_encoding``. Hence the consumers in kombu < 3.0.11 refuse to accept them.

Most of the tests using ``rabbit_manager.publish`` are dispatching events, so could and probably should use ``nameko.standalone.events.event_dispatcher`` instead, which was added after these tests were written.

The other places ``rabbit_manager.publish`` gets used should specify the content type and encoding using the ``properties`` kwarg:

``` python
msg_properties = {
    'content_type': 'application/json',
    'content_encoding': 'utf-8',
}
rabbit_manager.publish(vhost, 'spam', '', 'shrub', properties=msg_properties)
```

I'll put my other comments inline

nameko/constants.py
```diff
@@ -18,8 +18,10 @@
 
 MAX_WORKERS_CONFIG_KEY = 'max_workers'
 PARENT_CALLS_CONFIG_KEY = 'parent_calls_tracked'
+KOMBU_SERIALIZER_CONFIG_KEY = 'kombu_serializer'
 
 DEFAULT_MAX_WORKERS = 10
 DEFAULT_PARENT_CALLS_TRACKED = 10
+DEFAULT_KOMBU_SERIALIZER = 'json'
```
mattbennett: Think I'd prefer dropping the ``KOMBU`` from these variables.

nameko/standalone/events.py
```diff
@@ -31,6 +34,11 @@ def event_dispatcher(nameko_config, **kwargs):
     def dispatch(service_name, event_type, event_data):
         conn = Connection(nameko_config[AMQP_URI_CONFIG_KEY])
 
+        if KOMBU_SERIALIZER_CONFIG_KEY in nameko_config.keys():
+            serializer = nameko_config[KOMBU_SERIALIZER_CONFIG_KEY]
+        else:
+            serializer = DEFAULT_KOMBU_SERIALIZER
```
mattbennett: Nicer to use the ``get`` method with a default:

``` python
serializer = nameko_config.get(SERIALIZER_CONFIG_KEY, DEFAULT_SERIALIZER)
```

Not sure why we do ``config.get(MAX_WORKERS_CONFIG_KEY) or DEFAULT_MAX_WORKERS)`` further up. Maybe you can fix that too...

davidszotten: `foo.get(key) or default` is different to `foo.get(key, default)` if e.g. `foo[key] == None  # or 0`

ayoshi: Yes, they are all different with regard to None.
I'm not sure yet whether to allow None at all,  or where to allow it. I'll
have a new variant, which seems to be a bit cleaner, I will post it after
fixing the tests

davidszotten: i think we should just drop the `foo.get(key) or default` pattern, as the behaviour is somewhat unexpected (certain config values just being ignored)

i did some digging for though the git history of this for the `max_workers`, and i can't see any direct reasons to override e.g. an explicit `None` or `0`. It won't work of course, but neither will setting the value to `foo`. We may add some validation/sanity checking at some point later, but for now i think it's fine to "allow" those config values and just let any errors bubble

ayoshi:
Phew!

I think I'm done. Summary on the change - since that's a lot of commits:

1. Messaging / Events / RPC are now aware of accept/serialize.
The settings are on the container level, but can be overridden by directly accessing the values after setup
None is always replaced by default serializer, so we don't send or accept unserialized messages. 
Again, it can be overriden post-setup.

2. Config settings : SERIALIZER_CONFIG_KEY/ DEFAULT_SERIALIZER ( removed kombu reference )

3. ALL versions including py27-oldest work properly, actually that specific version failed, because it turned out that test suite relied on all sorts of implicit behaviors, which became apparent after serialization support had been introduced, so it should stay :)

3. Most of the changes required test overhauls, that is what has been done:

    * All tests using rabbit_manager to publish now explicitly send content_type in places where we use kombu.consumers to verify
    * Mocking of the container has been factored out and parametrized as a pytest fixture. That means that all tests using containers will test both json and pickle as serializers.
    * Most of the messages that we sent raw are now sent as json strings '"msg"' instead of 'msg'. That is a minor change, so kombu consumers can deal with them, since we only accept serialized messages now.
    *  test that specifically relied on JSON have been generalized and parametrized with both json/pickle
    

Because of parametrized mock_container all relevant features are tested both with json and pickle, so serialization is properly covered by test 100% 

Over to you.

davidszotten:
amazing work!

need to read the code, but a small initial comment from reading the summary: i'm not sure we should be encouraging the use of pickle (even as an example serializer/alternative to json)

ayoshi:
@davidszotten 

I used pickle to support serializer testing, because it's already in the standard library and is supported by kombu. I could use yaml, but I think pickle+json a better bet for testing, since some unittests relied on json losing types during serialization, and yaml behaves the same way. I wanted to see how something like pickle would perform.

As for documentation - yes, I agree - it would be better to use yaml in examples. 

Though, for the record - I'm not really afraid of people using pickle - for many systems security concerns apply on a different level - and when you're working on a distributed system and your types are not yet stable, pickle is much more convenient, you can work quickly without fighting with serialization formats.

mattbennett:
Thanks @ayoshi, this is great. Really good to see the implicit behaviour made explicit, and I agree that json+pickle is a good combination of serializers for testing. 

I think parametrizing the ``mock_container`` is too broad though. Most of the tests where that fixture is used don't actually publish or receive messages, so we're running them twice without adding much value (or coverage). Some of these tests are fairly long in the tooth and not very accessible I'm afraid, so it's not obvious.

I think I'd prefer to have parametrized tests around the individual components, e.g. make an RPC call and handle it (end to end) using both serialization formats; verify the standalone RPC proxy and event dispatcher honour the serialization format; publish a message and consume it using the ``queue_iterator`` again using both formats. You could optionally check the format on the wire by inspecting the rabbit queues, although that might be a bit of a faff. 

I'd also like to see a test that defines and registers a new trivial serializer and demonstrates that end to end. It's not strictly nameko's responsibility to test this but I'd like to verify that it can be done.

I'll put my other small comments inline.

nameko/rpc.py
```diff
@@ -351,6 +359,11 @@ def _call(self, *args, **kwargs):
             transport_options={'confirm_publish': True},
         )
 
+        serializer = (
+            container.config.get(SERIALIZER_CONFIG_KEY) or
+            DEFAULT_SERIALIZER
+        )
```
mattbennett: Should probably use ``config.get(SERIALIZER_CONFIG_KEY, DEFAULT_SERIALIZER)`` as elsewhere

ayoshi:
@mattbennett 

With mock_container factored out - we can simply use another one (something like mock_container_with_serializer)  to single out tests that do publish/receive message and parametrize that specific fixture - how does that sound? If that's an option - let me know which tests you want with mock container.

Otherwise - I wanted to make sure all the tests that really send/recieve messages work with both serializers - it's only 20 tests that are affected, and all seem to be in relevant places ( test_events, test_messaging, test_rpc ) - nobody else uses mock containers, though I'm not sure about specific tests -
It seemed like a most bang for the buck.  I'm not familiar with test codebase enough to create a new array of tests specifically to test serializers - and it looks like time investment would be too prohibitive for me.

The difference in run time is 69s vs 73s on my laptop.

I'll see if I can add simple custom serializer test.

ayoshi:
@mattbennett 

1. Ok, many tests with mock container actually publish raw data, so don't need to be with parametrized mock container. Could you let me know if there are tests which don't use mock containers need to be tested with both serializers?

2. Though I still think that having mock container as a fixture is a good idea, do you want to remove it from all tests, and apply only in tests that test publishing?

mattbennett:
Sorry, these tests are in more of a mess than I realised. The reason that many of them don't honour the serializer is that the extensions are constructed manually by the tests, rather than via the container. They're in relevant places but written in a way that requires you to wire up the internals.

To answer your questions above:

1. I don't think there are any existing tests that can be easily parametrized to use both serializers. test_messaging::test_publish_to_rabbit and test_messaging::test_consume_from_rabbit should test both serializers but should be rewritten rather than tweaked.

2. I like the ``mock_container`` fixture. I think we should keep it and just drop the parametrization.

Thank you very much for your efforts on this. Don't want to step on your toes but I'd be happy to add the explicit serialization tests on top of this work. Or otherwise I could clean up the tests so the serializer parametrization worked in the way you'd expect.

ayoshi:
@mattbennett 

I'll drop parametrization from mock_container ( the rest of the test fixes are OK, I think ). I also want to reformat the commit to be smaller ( per file ) - there were too many changes during experimentation,
factor out an unserializable fixture ( it's used twice ), will fix a small mistake in parameters and will pass it over to you - If you could add a couple of explicit tests that would be great - It'll take me too much time to go through the tests at this stage of familiarity with nameko's codebase.

After that I'll see if I can help cleaning up the tests a bit in general - but unfortunateIy I can't give a hard promise - I'm currently overloaded a bit, maybe piece by piece reorganisation one commit at a time.

ayoshi:
@mattbennett 

Ok, I redid the commits so they would be more manageable, removing all the intermediate changes. I also removed parametrization from container mock, and cleaned up small things. 

Passing it over to you, to add separate serialization tests. Let me know if anything is lacking.

mattbennett:
@ayoshi thank you very much, looks great. I'll add the separate serialization tests on top.

mattbennett:
@ayoshi apologies for the long delay on this. I took a long vacation and have been swamped since I got back.

I've raised https://github.com/onefinestay/nameko/pull/264 which brings this up to date with master and adds high-level test coverage. Will ask one of the other collaborators to review my additions and then we'll get this merged. Thank you again for your contribution.

Issue #210, opened 2015-02-18
use ipython for nameko shell


davidszotten:
if available

see e.g. https://github.com/django/django/blob/9996158db467f9bef8243fcc36dc3602570e3613/django/core/management/commands/shell.py

dAnjou:
I'd strongly vote for [bpython](http://www.bpython-interpreter.org/).

davidszotten:
i guess we could look for various shells. the django code linked to above looks for ipython and bpython.

mattbennett:
Merged with https://github.com/onefinestay/nameko/pull/252

iho:
Hi! 
ipython is pretty good, but how about  this https://github.com/jonathanslenders/ptpython ?

davidszotten:
rather than trying to keep up with a growing list of alternative repl's, maybe we can refactor things a little to make it easy for people to roll their own

PR #245, opened 2015-04-22
Optional interactive interpreters

Mytho:
Implements issue #210.

This allows the optional use of bpython or IPython as the interactive interpreter. I used the referenced [Django code](https://github.com/django/django/blob/9996158db467f9bef8243fcc36dc3602570e3613/django/core/management/commands/shell.py) as an example for this implementation.

### DIFF
--- a/nameko/cli/shell.py
+++ b/nameko/cli/shell.py
@@ -31,17 +31,18 @@ class ShellRunner(object):
         embed(banner1=self.banner, user_ns=self.local)
 
     def plain(self):
+        code.interact(banner=self.banner, local=self.local)
+
+    def start_shell(self, name):
+        available_shells = [name] if name else SHELLS
+
         # Support the regular Python interpreter startup script if someone
         # is using it.
         startup = os.environ.get('PYTHONSTARTUP')
         if startup and os.path.isfile(startup):
             with open(startup, 'r') as f:
                 eval(compile(f.read(), startup, 'exec'), self.local)
-
-        code.interact(banner=self.banner, local=self.local)
-
-    def start_shell(self, name):
-        available_shells = [name] if name else SHELLS
+            del os.environ['PYTHONSTARTUP']
 
         for name in available_shells:
             try:
--- a/nameko/cli/shell.py
+++ b/nameko/cli/shell.py
@@ -13,10 +13,19 @@ from nameko.standalone.rpc import ClusterRpcProxy
 from nameko.standalone.events import event_dispatcher
 
 
+SHELLS = ['bpython', 'ipython']
+
+
 def init_parser(parser):
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


davidszotten:
hey, thanks for the contribution! will try to find time to review as soon as possible

nameko/cli/shell.py
```diff
@@ -13,10 +13,19 @@
...
+
+
 def init_parser(parser):
     parser.add_argument(
         '--broker', default='amqp://guest:guest@localhost',
         help='RabbitMQ broker url')
+    parser.add_argument(
+        '--interface', choices=SHELLS,
+        help='Specify an interactive interpreter interface.')
+    parser.add_argument(
```
davidszotten: don't think we need a separate option for this (think django has this for historical reasons. tried digging through the source and got as far as the svn->git conversion and lost track. from what i can tell, django originally had `--plain` to stop defaulting to ipython, and only later added `--interface` to support bpython

here i guess we could just add 'plain' as another choices in `SHELLS`?

nameko/cli/shell.py
```diff
@@ -42,6 +51,27 @@ def make_nameko_helper(config):
...
+    from IPython import embed
+    embed(banner1=banner, user_ns=local)
+
+
+def run_shell(shell=None, banner=None, local=None):
+    available_shells = [shell] if shell else SHELLS
+
+    for shell in available_shells:
+        try:
+            return globals()[shell](banner, local)
```
davidszotten: not so keen on using `globals` like this

nameko/cli/shell.py
```diff
@@ -53,11 +83,17 @@ def main(args):
...
-    if startup and os.path.isfile(startup):
-        with open(startup, 'r') as f:
-            eval(compile(f.read(), startup, 'exec'), ctx)
+    try:
+        if args.plain:
+            raise ImportError
+
+        run_shell(shell=args.interface, banner=banner, local=ctx)
+    except ImportError:
+        # Support the regular Python interpreter startup script if someone
```
davidszotten: from what i can tell (using latest versions on pypi, might differ between versions), calling `embed` in bpython causes pythonstartup to be executed, but in ipython. one option is to run it "manually" for all shells (like this), and then unset the environment variable to prevent it from running again. thoughts?

Mytho:
Thanks for your constructive feedback.

I have:
- removed the `--plain` argument and made `plain` an available shell.
- added `ShellRunner` responsible for starting the interactive shells, so the globals are gone.
- added the manual startup script before each of the interactive shells.

davidszotten:
cool. had a quick look at drying up the tests and noticed they fail if bpython or ipyhton are installed. added a fix on top (see separate pr)

PR #247, opened 2015-04-23
get_container method in nameko.testing.utils instead of nameko.utils

Mytho:


davidszotten:
nice catch. actually, a whole heap of the examples have bugs

(c.f. `$ for x in docs/examples/*.py; do python $x; done`)

need to experiment with `timeout`, since at least the standalone rpc example will just wait forever for a reply, but would be nice to add a test that checks that they at least don't raise exceptions

mattbennett:
Thanks @Mytho. This has been folded into #250

PR #250, opened 2015-04-29
Fix examples

mattbennett:


docs/examples/test_examples.py
```diff
@@ -0,0 +1,268 @@
...
+
+from nameko.containers import ServiceContainer
+from nameko.standalone.events import event_dispatcher
+from nameko.standalone.rpc import ServiceRpcProxy
+from nameko.testing.services import entrypoint_waiter
+from nameko.testing.websocket import make_virtual_socket
+
+
+@pytest.yield_fixture
+def container_factory(rabbit_config):
```
mattbennett: Now might be the time to promote these fixtures out of ``conftest.py`` so we can import rather than reimplement them here.

docs/examples/README
```diff
@@ -0,0 +1,7 @@
+This directory (nameko/docs/examples) contains example services and code
+snippets for reference or use in the documentation. `test_examples.py`
+contains pytest-based tests to check the examples work as expected.
+These tests are verified to pass as part of nameko's release process.
+
+The `testing` subdirectory contains examples demonstrating best practice use
+of the test utilities that ship with nameko. These tests must also verified to pass as part of nameko's release process.
```
davidszotten: a note about how to do this "verification" (add to suite in ...)?

docs/examples/event_broadcast.py
```diff
@@ -3,7 +3,9 @@
 class ListenerService(object):
     name = "listener"
 
-    @event_handler("controller", "ping", handler_type=BROADCAST)
+    @event_handler(
+        "monitor", "ping", handler_type=BROADCAST, reliable_delivery=False
```
davidszotten: should `broadcast` make the default go the other way?

docs/examples/test_examples.py
```diff
@@ -0,0 +1,268 @@
+""" Tests for the files in nameko/docs/examples
```
davidszotten: i would expect this file to live in the `tests` directory with our other tests

docs/examples/test_examples.py
```diff
@@ -0,0 +1,268 @@
...
+    def make_container(service_cls, config, worker_ctx_cls=None):
+        container = ServiceContainer(service_cls, config, worker_ctx_cls)
+        all_containers.append(container)
+        return container
+
+    yield make_container
+
+    for c in all_containers:
+        try:
+            c.stop()
```
davidszotten: kill?

docs/examples/test_examples.py
```diff
@@ -0,0 +1,268 @@
...
+        except:
+            pass
+
+
+@pytest.yield_fixture()
```
davidszotten: is this copied from the web tests?

docs/examples/test_examples.py
```diff
@@ -0,0 +1,268 @@
...
+def websocket():
+    active_sockets = []
+
+    def socket_creator():
+        ws_app, wait_for_sock = make_virtual_socket(
+            '127.0.0.1', 8000)
```
davidszotten: in the other web tests we jump through hoops to make sure we pick a random (free) port

docs/examples/test_examples.py
```diff
@@ -0,0 +1,268 @@
...
+        container = container_factory(Service, rabbit_config)
+        container.start()
+
+        res = requests.get(self.base_url + "/privileged")
+        assert res.status_code == 403
+        assert res.text == 'Forbidden'
+
+        res = requests.get(self.base_url + "/headers")
+        assert res.status_code == 201
+        assert res.headers['location'] == 'https://www.example.com/widget/1'
```
davidszotten: location? is this leftover from when this was a redirect?

docs/examples/test_examples.py
```diff
@@ -0,0 +1,268 @@
...
+        container_a.start()
+        container_b.start()
+
+        with ServiceRpcProxy('service_a', rabbit_config) as service_a_rpc:
+
+            with patch.object(ServiceB, 'handle_event') as handle_event:
+
+                with entrypoint_waiter(container_b, 'handle_event'):
+                    service_a_rpc.dispatching_method()
+                assert handle_event.call_args_list == [call("payload")]
```
davidszotten: is it confusing that the variable `payload` ends up containing the string `"payload"`?

docs/examples/test_examples.py
```diff
@@ -0,0 +1,268 @@
...
+        from events import ServiceA, ServiceB
+
+        container_a = container_factory(ServiceA, rabbit_config)
+        container_b = container_factory(ServiceB, rabbit_config)
+        container_a.start()
+        container_b.start()
+
+        with ServiceRpcProxy('service_a', rabbit_config) as service_a_rpc:
+
+            with patch.object(ServiceB, 'handle_event') as handle_event:
```
davidszotten: are we not aiming for 100% coverage here?

docs/examples/test_examples.py
```diff
@@ -0,0 +1,268 @@
...
+            waiter_2 = entrypoint_waiter(container_2, 'ping')
+
+            with waiter_1, waiter_2:
+                dispatch("monitor", "ping", "payload")
+            assert ping.call_count == 2
+
+
+class TestExample(object):
+
+    def test_example(self, container_factory, rabbit_config):
```
davidszotten: is this now badly named?

docs/examples/test_examples.py
```diff
@@ -0,0 +1,268 @@
...
+
+        from example import Service
+
+        container = container_factory(Service, rabbit_config)
+        container.start()
+
+        with ServiceRpcProxy('service', rabbit_config) as service_rpc:
+            assert service_rpc.method() == None
```
davidszotten: typically `is None`

docs/examples/test_examples.py
```diff
@@ -0,0 +1,268 @@
...
+
+
+class TestFibonacci(object):
+
+    def test_fibonaccci(self, container_factory, rabbit_config):
```
davidszotten: this is a strange example. possibly unhelpfully so

docs/examples/test_examples.py
```diff
@@ -0,0 +1,268 @@
...
+        from rpc import ServiceX, ServiceY
+
+        container_x = container_factory(ServiceX, rabbit_config)
+        container_y = container_factory(ServiceY, rabbit_config)
+        container_x.start()
+        container_y.start()
+
+        # async example doesn't import due to undefined variables
+        # use execfile with a local namespace
+        from nameko.standalone.rpc import ClusterRpcProxy
```
davidszotten: move to top?

docs/examples/event_broadcast.py
mattbennett: Yeah that would be nicer

docs/examples/test_examples.py
```diff
@@ -0,0 +1,268 @@
...
+        }
+
+        dirpath = os.path.dirname(os.path.realpath(__file__))
+        filepath = os.path.join(dirpath, 'async_rpc.py')
+
+        with entrypoint_waiter(container_x, 'remote_method'):
+            execfile(filepath, globals(), ns)
+
+
+class TestServiceContainer(object):
```
davidszotten: i wonder if we still need these examples now that we have a cli

nameko/web/server.py
```diff
@@ -93,6 +93,7 @@ def start(self):
     def stop(self):
         self._is_accepting = False
         self._gt.kill()
+        self._sock.close()
```
davidszotten: this is the listening socket, right? so calls in-flight can still finish and send responses?

tox.ini
```diff
@@ -41,6 +41,14 @@ commands =
     sphinx-build -n -b html -d build/doctrees . build/html
 
 
+[testenv:examples]
```
davidszotten: might want to run this with at least py2 and py3 (c.f. top of this file)

docs/examples/test_examples.py
mattbennett: If we put it in the ``tests`` directory it'll be executed on every test run, as opposed to only the ``examples`` testenv. You also have to manually import the modules because the ``docs`` directory is outside the normal package structure. Not opposed to it though.

docs/examples/test_examples.py
mattbennett: Yeah, this was lifted from ``conftest.py`` which still uses ``stop`` (which we should also fix). c.f. my note about hoisting the implementations so we can import them here.

docs/examples/test_examples.py
mattbennett: I didn't want to duplicate all that machinery, but we probably shouldn't hardcode the port like this.

docs/examples/test_examples.py
mattbennett: 201 returns a location header. It's the URI of the created resource.

docs/examples/test_examples.py
mattbennett: Asking the question is probably enough to answer it ;) I'll change it.

docs/examples/test_examples.py
davidszotten: probably wouldn't be a bad thing to always run (e.g. to make sure examples work for all versions we support)

nameko/web/server.py
mattbennett: It is the listening socket, and the greenthread doing the accepting is already dead.

Unrelated to this change though, I don't think the webserver is a good citizen with regard to letting in-flight calls finish. It doesn't track the greenthreads spawned for each request, so ``stop`` may exit while they're still running, in which case the container will kill them ungracefully. I'll raise as a separate issue.

docs/examples/test_examples.py
mattbennett: Fair enough, will drop.

docs/examples/test_examples.py
mattbennett: That was my original aim, but I don't think we'll get there without contorting the examples. Also coverage doesn't count the code executed via ``execfile``.

docs/examples/test_examples.py
mattbennett: They're present in the docs (http://nameko.readthedocs.org/en/latest/key_concepts.html#running-services) so I think they should be tested here.

nameko/web/server.py
mattbennett: Actually this is fine. The ``WebServer`` is a subextension so it's stopped after all the workers have gracefully completed. By the time this method is called there won't be any in-flight calls.

mattbennett:
For visibility, this is waiting on #266 to land so I can address the final comments

docs/examples/test_examples.py
mattbennett: It's tricky to have this file as ``test/test_examples.py`` and also check coverage of the ``docs/examples`` directory. I've settled on giving it a separate entry in the Makefile and adding it as one of the steps under ``make test``.

docs/examples/test_examples.py
mattbennett: This fix landed in coverage ``4.0b3``, so we can aim for 100% coverage now.

.travis.yml
```diff
@@ -13,21 +13,8 @@ addons:
 install:
   - pip install tox
 
-env:
```
davidszotten: the reason we had the former was to get separate (and parallel) builds in travis

tox.ini
```diff
@@ -23,27 +25,16 @@ deps =
...
+whitelist_externals = make
 
 commands =
-    pip install --editable .[dev]
-    flake8 nameko test
-    pylint --rcfile=pylintrc nameko -E
-    coverage run --concurrency=eventlet --source nameko -m \
-        pytest test {posargs:--timeout=30}
-    coverage report --show-missing --fail-under=100
+    codebase: pip install --editable .[dev]
```
davidszotten: not sure i find this style more readable. how come the switch?

tox.ini
```diff
@@ -1,20 +1,22 @@
 [tox]
-envlist = {py27,py33,py34}-{oldest,pinned,latest}, docs, spelling
+envlist = {py27,py33,py34}-{oldest,pinned,latest}-codebase, {py27,py34}-examples, docs
```
davidszotten: `lib` or `library` might be a better name than `codebase`

tox.ini
mattbennett: It got ugly making the examples run with multiple python versions. You have to make a testenv that exactly matches the expanded environments so that they don't inherit the default. Like this:

```
[testenv:py27-examples]
pip install --editable .[dev,examples]
make test_examples

[testenv:py34-examples]
pip install --editable .[dev,examples]
make test_examples

...
```

It's not terrible with only two repetitions, but would get worse with more python versions or varieties of commands. FWIW I find the new version easier to understand too.

davidszotten: could we not just move `commands` out of the default section into a `[testenv:lib]`?

davidszotten: bump (got eaten by github when you changed `codebase` to `lib`)

mattbennett: Ah, I wondered what happened to this! I saw the email but couldn't find it...

``[testenv:lib]`` won't be matched by anything. The name has to exactly match one of the expanded  values (e.g. ``py27-latest-lib``).

davidszotten: i see. ok

davidszotten:
:+1:

mattbennett:
Merged as part of #280

PR #255, opened 2015-05-12
nameko run config

kollhof:
A simple way to configure the service runner via a YAML config file.
```
> nameko run --config ./nameko.yaml foobar
```

Motivation:

I am running multiple nameko processes, each with a slightly different config for e.g. the HTTP port.
I wanted to avoid writing my own runner as `nameko run` already provided most of what I needed. 

Current config options are:

```yaml
WEB_SERVER_ADDRESS: 0.0.0.0:8000
AMQP_URI: amqp://guest:guest@localhost
rpc_exchange: nameko-rpc
max_workers: 10
parent_calls_tracked: 10
```


setup.py
```diff
@@ -17,6 +17,7 @@
         "kombu>=3.0.1",
         "mock>=1.0.1",
         "path.py>=6.2",
+        "pyyaml>=3.11",
```
mattbennett: This should be the lowest version of pyyaml we're prepared to work with.

nameko/cli/run.py
```diff
@@ -168,7 +169,14 @@ def main(args):
...
         )
 
-    config = {AMQP_URI_CONFIG_KEY: args.broker}
+    config = {}
+
+    if args.config:
+        with open(args.config) as fle:
+            config = yaml.load(fle)
+    else:
+        config[AMQP_URI_CONFIG_KEY] = args.broker
```
mattbennett: Do we want to keep ``--broker`` as an argument? What about adding ``--web-server-address``?

Obviously requiring every possible config value to be specified as a command line option is ridiculous, but maybe allowing a subset would be helpful. And then which subset?

Is there a world where the runner allows any possible config to be specified on the command line, and it's up to the user to use a config file when the invocation gets silly? It would require us to start registering config keys, but that's something I'd like to do anyway -- each Extension could specify the config it requires and default values; the ServiceContainer could do the same (for ``MAX_WORKERS`` and ``PARENT_CALLS_TRACKED``).

kollhof: I really like the idea of registered config keys, descriptions and defaults. I could then introspect a class and determine what configs I may need to run it and match it up with a preexisting config I already have.

As you said, command line args will end up getting silly, but I did not want to break compatibility. 
One option could be to treat them as deprecated and print out an warning when used.

setup.py
kollhof: 3.10 is the lowest I could install with pip

nameko/cli/run.py
mattbennett: Let's keep ``--broker`` for backwards compatibility. @davidszotten already raised the notion of registering config keys in https://github.com/onefinestay/nameko/issues/223. We can track these ideas there.

mattbennett:
:+1:

mattbennett:
Can you add a small example to http://nameko.readthedocs.org/en/latest/cli.html#running-a-service?

mattbennett:
Great stuff

stchris:
Quick question about this: the 2.1.1 docs advertise the `--config` option as being available, but it doesn't seem to work:

```
nameko: error: unrecognized arguments: --config
```
This is supposed to be in the 2.1.1 release, right?

davidszotten:
due to some accidental confusion, this feature actually landed in `2.1.2`

davidszotten:
where are you seeing this in 2.1.1 docs?

stchris:
The confusing thing is that the latest version docs are labeled 2.1.1, this is where I read about the `--config` option. According to http://nameko.readthedocs.org/en/latest/release_notes.html there is no 2.1.2.

davidszotten:
interesting. thanks for pointing this out. @mattbennett any idea what might have gone wrong with readthedocs (or how to trigger a re-build)?

mattbennett:
@stchris Sorry for the confusion. As @davidszotten says, the feature landed in 2.1.2 by accident and consequently without release notes.

We were also serving the 'latest' docs by default on readthedocs, which is why ``--config`` was documented but not actually available when 2.1.1 was the latest stable release.

And then weirdly readthedocs didn't rebuild 'latest' when 2.1.2 came out. I'm not sure why that happened, but 'latest' should now be correct. I've also changed the default build to 'stable' so the docs should reflect the most recently released version.

Thanks for raising this.

PR #257, opened 2015-05-13
refactor the polling queue consumer

davidszotten:
unroll the nested context managers and drop the coroutine style, to
support changing the timeout on an existing queue consumer


nameko/standalone/rpc.py
```diff
@@ -66,12 +67,17 @@ class PollingQueueConsumer(object):
...
+        channel = self.connection.channel()
         # queue.bind returns a bound copy
-        self.queue = self.queue.bind(self.channel)
-        maybe_declare(self.queue, self.channel)
+        self.queue = self.queue.bind(channel)
+        maybe_declare(self.queue, channel)
+        consumer = Consumer(channel, queues=[self.queue], no_ack=False)
+        consumer.callbacks = [self.on_message]
+        consumer.consume()
+        self.consumer = consumer
```
mattbennett: This does more than set up the queue now. It also creates/replaces the consumer

mattbennett: What happens to the first consumer in the event of a timeout, temporary error or ``KeyboardInterrupt``? Can we end up multiple consumers?

nameko/standalone/rpc.py
```diff
@@ -64,14 +65,27 @@ class PollingQueueConsumer(object):
...
     def __init__(self, timeout=None):
         self.timeout = timeout
+        self.replies = {}
+
+    def _setup_consumer(self):
+        if self.consumer is not None:
+            try:
+                self.consumer.cancel()
+            except socket.error:  # pragma: no cover
+                pass  # already disconnected
```
mattbennett: Might be worth a note here for posterity, explaining that some platforms set ``self.consumer`` to ``None`` on disconnection and some don't -- so it's not possible to reasonably cover this.

Alternatively, we *could* write a test that runs conditionally on the platform.

PR #264, opened 2015-06-26
Custom serialization

mattbennett:
Adds high-level test coverage for alternative serialization formats on top of https://github.com/onefinestay/nameko/pull/244


nameko/messaging.py
```diff
@@ -98,6 +106,7 @@ def spam(self, data):
         """
         self.exchange = exchange
         self.queue = queue
+        self.serializer = serializer
```
noisyboiler: when `setup` is called this value is ignored, and indeed your docstring states it will always be overriden. So why maintain the value on the instance at all? is there a case when `setup` is not called and we determine the correct serializer from this value?

nameko/exceptions.py
```diff
@@ -69,11 +69,11 @@ def safe_for_json(value):
         return value
     if isinstance(value, dict):
         return {
-            safe_for_json(key): safe_for_json(val)
+            safe_for_serialization(key): safe_for_serialization(val)
             for key, val in six.iteritems(value)
         }
     if isinstance(value, collections.Iterable):
-        return list(map(safe_for_json, value))
+        return list(map(safe_for_serialization, value))
```
noisyboiler: as `map` always returns a list, do you need the extra call of `list`?

davidszotten: `map` doesn't return a list on python3

```
>>> map(None, [])
<map object at 0x10240d5f8>
```

nameko/rpc.py
```diff
@@ -194,12 +195,16 @@ def send_response(self, result, exc_info, **kwargs):
...
         # unrecoverable errors (and the message will be requeued for another
         # victim)
+
+        serializer = self.config.get(
+            SERIALIZER_CONFIG_KEY, DEFAULT_SERIALIZER)
+
         try:
-            json.dumps(result)
+            kombu.serialization.dumps(result, serializer)
         except Exception:
```
noisyboiler: can we now `except` Kombu's custom exceptions instead?

nameko/messaging.py
mattbennett: Quite right. I didn't spot this in the review on #244. There's no realistic situation where ``setup()`` isn't called.

davidszotten:
:+1:

nameko/rpc.py
mattbennett: We need to catch everything here, to make sure we avoid the nasty suicide and requeue problem.

If ``serializer`` is a bogus value for example, kombu will probably throw something other than nice serialization error, and we want to catch it regardless.

davidszotten:
:+1:

PR #265, opened 2015-07-01
Use Makefile as source of truth for commands

mattbennett:
This will simplify a number of the changes to ``tox.ini`` in #250. Raised as a separate change for visibility and to slim down that PR.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -25,25 +25,18 @@ deps =
 
     latest: eventlet  # need something
 
+whitelist_externals = make
+
 commands =
     pip install --editable .[dev]
-    flake8 nameko test
-    pylint --rcfile=pylintrc nameko -E
-    coverage run --concurrency=eventlet --source nameko -m \
-        pytest test {posargs:--timeout=30}
-    coverage report --show-missing --fail-under=100
-
+    make test
 
 [testenv:docs]
-changedir = docs
 commands =
-    pip install --editable ..[docs]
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -25,25 +25,18 @@ deps =
 
     latest: eventlet  # need something
 
+whitelist_externals = make
+
 commands =
     pip install --editable .[dev]
-    flake8 nameko test
-    pylint --rcfile=pylintrc nameko -E
```
davidszotten: we ok without specifying the rc file?

tox.ini
```diff
@@ -25,25 +25,18 @@ deps =
...
-    coverage run --concurrency=eventlet --source nameko -m \
-        pytest test {posargs:--timeout=30}
-    coverage report --show-missing --fail-under=100
```
davidszotten: show-missing is handy for debugging failed builds on travis

davidszotten:
i'm sure there was some reason for the way we structured this, but can't recall it, so happy to try this way around

mattbennett:
If I recall correctly we wanted to embrace tox but left the Makefile for convenience. This way we get both with less repetition, and you can dev locally without tox if you want.

Dropping ``show-missing`` and the ``rcfile`` was accidental, I'll add them back again.

davidszotten:
:+1:

mattbennett:
Superseded by https://github.com/onefinestay/nameko/pull/266

PR #266, opened 2015-07-02
move pytest fixtures and config to a plugin

mattbennett:
The motivation here is to avoid reimplementing fixtures and pytest configuration in every application (as in #250).

The original idea was to make the fixtures importable, so you can build your own conftest quickly. By registering the module as a pytest plugin you get them for free. It should be backwards compatible because the config/fixtures in any conftest will override the ones from the plugin.


mattbennett:
Tests are failing on python3 due to the ``eventlet.monkey_patch()`` in the pytest plugin. Still investigating the root cause.

docs/testing.rst
```diff
@@ -73,10 +73,9 @@ The entrypoint waiter is a context manager that does not exit until a named entr
...
 
-Minimal conftest.py
--------------------
-
-Below is a minimal conftest for pytest that includes configuration and fixtures for communicating with RabbitMQ as well as the eventlet monkey patch. It is a good base for service tests if you're using pytest.
+Using py.test
+-------------
 
-.. literalinclude:: examples/testing/conftest.py
+Nameko's test suite uses py.test, and makes some useful configuration and fixtures available for your own tests if you choose to use pytest.
```
davidszotten: maybe call it "py.test" or "pytest"

test/conftest.py
```diff
@@ -1,130 +1,25 @@
...
+def testdir(testdir):
+    """ Wrap pytester's ``testdir`` fixture to include coverage from
+    the generated tests.
+    """
+    rc_path = os.path.join(
+        os.path.dirname(__file__), os.pardir, '.coveragerc'
+    )
+    os.environ['COVERAGE_PROCESS_START'] = os.path.normpath(rc_path)
+
+    testdir.makepyfile(sitecustomize="""
```
davidszotten: don't think you need this if you have `pytest-cov` installed. it adds a `.pth` file

test/conftest.py
```diff
@@ -1,130 +1,25 @@
...
-        except:
-            pass
-
 
 @pytest.yield_fixture
-def predictable_call_ids(request):
-    with patch('nameko.containers.new_call_id', autospec=True) as get_id:
-        get_id.side_effect = (str(i) for i in itertools.count())
-        yield get_id
+def testdir(testdir):
```
davidszotten: nice

.coveragerc
```diff
@@ -0,0 +1,7 @@
+[run]
+concurrency = eventlet
+source = nameko
```
davidszotten: do we want coverage for the tests as well? (catches e.g. duplicate test names)

test/conftest.py
```diff
@@ -1,130 +1,25 @@
...
+    os.environ['COVERAGE_PROCESS_START'] = os.path.normpath(rc_path)
+
+    testdir.makepyfile(sitecustomize="""
+        import coverage; coverage.process_startup()
+    """)
+
+    yield testdir
+
+    ident = uuid.uuid4()
+    dst = os.path.join(testdir._olddir.strpath, '.coverage.{}'.format(ident))
```
davidszotten: not sure which i prefer, but could use `parallel = true` in `coveragerc` instead

test/conftest.py
```diff
@@ -1,130 +1,25 @@
...
+    os.rename(os.path.join(testdir.tmpdir.strpath, '.coverage'), dst)
```
davidszotten: what triggers `coverage combine`?

Makefile
```diff
@@ -11,8 +11,7 @@ pylint:
 	pylint nameko -E
 
 pytest:
-	coverage run --concurrency=eventlet --source nameko -m pytest test
-	coverage report --fail-under=100
+	py.test test --cov
```
davidszotten: do you also want to update `tox.ini`?

docs/testing.rst
mattbennett: Thanks, will do

test/conftest.py
mattbennett: Doesn't appear to work.

I see the ``.pth``, and it contains the following (with the conditional in an ``exec`` so it's on one line):

``` python
import os, sys;

if 'COV_CORE_SOURCE' in os.environ:
    try:
        from pytest_cov.embed import init
        init()
    except ImportError:
        sys.stderr.write("Failed to setup coverage."
                         "Sources: {[COV_CORE_SOURCE]!r}"
                         "Config: {[COV_CORE_CONFIG]!r}"
                         "Exception: {!r}".format(os.environ, exc)))
```

Those environment variables are set. It seems like ``pytest-cov`` is attempting to handle collection of coverage from subprocesses itself, but maybe not working. If we can fix that maybe this entire fixture can go away.

test/conftest.py
mattbennett: Rather than using a uuid? Then you'd have to figure out what the source file was called in order to move it.

test/conftest.py
mattbennett: ``pytest-cov`` I assume

Makefile
mattbennett: Yes, that's why the travis build is broken!

Alternatively we could land https://github.com/onefinestay/nameko/pull/265

.coveragerc
mattbennett: I think we should if possible. Will need a bit of work to get it up to 100% though. Separate PR I think

davidszotten:
looks good. we probably want a beta run against the larger projects at ofs

mattbennett:
Agreed. I'm going to merge this into the "fix-examples" branch to address the final comments there, then cut an RC branch with everything that's marked as "ready"

mattbennett:
Incorporated into #250

PR #271, opened 2015-07-25
Fix get_container import in the service runner example

miguelfrde:
`get_container` is in `nameko.testing.utils` not in `nameko.utils`.


mattbennett:
Thanks for the contribution! This is actually duplicate of #247 and has been incorporated into #250

PR #273, opened 2015-07-25
Improved error message if the RabbitMQ management API is not reachable.

davidweterings:
During EuroPython 2015 sprint people got weird errors during setup, this pull request raises a friendlier, more clear exception if the RabbitMQ management API is not reachable.

### DIFF
--- a/nameko/testing/rabbit.py
+++ b/nameko/testing/rabbit.py
@@ -1,6 +1,6 @@
 import json
 
-from requests import HTTPError, Session
+from requests import HTTPError, Session, ConnectionError
 from six.moves.urllib.parse import quote  # pylint: disable=E0611
 
 __all__ = ['Client', 'HTTPError']
@@ -17,6 +17,7 @@ class Client(object):
         self._base_url = '{}/api'.format(uri)
         self._session = Session()
         self._session.headers['content-type'] = 'application/json'
+        self._verify_api_connection()
 
     def _build_url(self, args):
         args = map(_quote, args)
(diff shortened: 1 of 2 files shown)


nameko/testing/rabbit.py
```diff
@@ -30,7 +31,13 @@ def _request(self, method, *args, **kwargs):
         json_data = kwargs.pop('json', None)
         if json_data is not None:
             kwargs['data'] = json.dumps(json_data)
-        result = self._session.request(method, url, **kwargs)
+
+        try:
+            result = self._session.request(method, url, **kwargs)
+        except ConnectionError:
+            raise Exception('Connection error for the RabbitMQ management HTTP'
```
davidszotten: as discussed, let's use `six.raise_from` like `nameko/amqp.py:21`

davidweterings:
Exception now uses six.raise_from!

test/testing/test_utils.py
```diff
@@ -247,3 +248,13 @@ def wait_for_event(self):
     with eventlet.Timeout(1):
         gt.wait()
     assert container._worker_pool.free() == max_workers
+
+
+def test_rabbit_connectipdateon_refused_error():
```
davidszotten: are there some stray characters in this function name?

test/testing/test_utils.py
```diff
@@ -247,3 +248,13 @@ def wait_for_event(self):
...
+
+    bad_port_uri = 'http://guest:guest@localhost:11111'
```
davidszotten: i guess we don't need credentials here?

davidszotten: do we need to worry about someone having this port open? we could use the trick from `test/web/conftest.py` to ask the os for a free port, and then close it and use that?

or maybe we could use one of the low, [well known](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Well-known_ports) but unassigned ports, e.g. `4`

davidweterings:
Yikes, newbie mistakes! Both are fixed now.

davidweterings:
The port is now also changed to 4.

davidszotten:
awesome! thanks

PR #275, opened 2015-08-06
Add support for custom error payload

kooba:
It would be nice to allow to define custom payload when raising exceptions within HTTP handler. Also added documentation around expected_exceptions.

### DIFF
--- a/docs/built_in_extensions.rst
+++ b/docs/built_in_extensions.rst
@@ -107,6 +107,35 @@ The HTTP entrypoint is built on top of `werkzeug <http://werkzeug.pocoo.org/>`_.
     Content-Length: 0
     Date: Fri, 13 Feb 2015 14:58:48 GMT
 
+Exceptions can be raised safely if added to ``expected_exceptions`` or if subclass from :class:`~nameko.exceptions.HttpError`:
+
+.. literalinclude:: examples/http_exceptions.py
+
+.. code-block:: shell
+
+    $ nameko run http_exceptions
+    starting services: http_exceptions_service
+
+.. code-block:: shell
+
+    $ curl -i http://localhost:8000/expected_exception
+    HTTP/1.1 400 BAD REQUEST
+    Content-Type: text/plain; charset=utf-8
+    Content-Length: 41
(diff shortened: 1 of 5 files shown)


davidszotten:
Hi,

thanks for the contribution!

I might be missing something, but what does this add over just

```
@http('GET', '/expected_custom_exception')
def expected_custom_exception(self, request):
    return 400, json.dumps({
        'error': 'INVALID_REQUEST',
        'description': 'This is invalid request.',
        'code': 400
    })
```

like your example, the content type is `text/plain` but should probably be `application/json`, which can be achieved by returning a 3-tuple `(status_code, headers dict, response body)` instead of just `(status code, response body)`

kooba:
Primary reason for raising vs just returning would be ability to capture and alert on these outside of main code with something like [nameko-sentry](https://github.com/mattbennett/nameko-sentry) 

I agree application/json should be returned in this instance.

davidszotten:
ok, i see what you mean (though one might wonder why your service wants to notify sentry about caller errors. do your webapps log 404s or invalid forms to sentry?)

going along with this though, dumping json into exceptions feels to me like mixing responsibilities. i think a better solution would be to open up for customising the way exceptions are turned into http responses (currently https://github.com/kooba/nameko/blob/custom-error-payload/nameko/web/handlers.py#L94). in fact, an early version of the web handler had a much more generic implementation (see e.g. https://github.com/kooba/nameko/blob/1a416524d52107fc93831bc42076bf5c14588da1/nameko/web/handlers.py#L69) we scaled it back until we could get a better feel for what we might want out of this type of feature. 

we might not need something quite this generic (protocol, vs say just a method in charge of this that you can subclass `HttpRequestHandler` and override to customise)

mattbennett:
Sentry is a specific example. Any ``DependencyProvider`` can implement ``worker_result`` and it seems reasonable that an http entrypoint returning a 4xx/5xx status should follow the "error" path:

```python
def worker_result(worker_ctx, result=None, exc_info=None):
    if exc_info is not None:
        # process worker error
```

I like the suggestion of an overridable method that transforms an exception into a response.

mattbennett:
Superseded by #291

PR #280, opened 2015-09-24
v2.2.0 rc

mattbennett:
RC branch for 2.2.0 release. cc @noisyboiler


setup.py
```diff
@@ -24,13 +33,14 @@
     ],
     extras_require={
         'dev': [
-            "coverage==4.0a1",
+            "coverage==4.0b3",
```
davidszotten: 4.0 is now out

mattbennett: Thanks. I added this to #250 and cherry-picked it here. If you're happy with that branch can you give it a :+1: ?

PR #288, opened 2015-10-19
Random vhost

mattbennett:
Addresses https://github.com/onefinestay/nameko/issues/278

Outstanding as of 2015-12-16:
* [x] "Yep, agree that the error should identify the connection. I will add."[comment](https://github.com/onefinestay/nameko/pull/288#discussion_r46677404)
* [x] Merge conflicts
* [ ] Change (optional) `RuntimeError` -> `CleanupError`


davidszotten:
seems like a reasonable approach

nameko/testing/pytest.py
```diff
@@ -84,30 +84,51 @@ def rabbit_manager(request):
 
 @pytest.yield_fixture()
 def rabbit_config(request, rabbit_manager):
+    import random
+    import string
     from kombu import pools
-    from nameko.testing.utils import (
-        reset_rabbit_vhost, reset_rabbit_connections,
-        get_rabbit_connections, get_rabbit_config)
+    from nameko.testing.utils import get_rabbit_connections
```
davidszotten: probably reverse this order

nameko/testing/pytest.py
```diff
@@ -84,30 +84,51 @@ def rabbit_manager(request):
...
     amqp_uri = request.config.getoption('AMQP_URI')
 
-    conf = get_rabbit_config(amqp_uri)
+    uri = urlparse(amqp_uri)
+    username = uri.username
+    vhost = uri.path[1:]
 
-    reset_rabbit_connections(conf['vhost'], rabbit_manager)
-    reset_rabbit_vhost(conf['vhost'], conf['username'], rabbit_manager)
+    use_random_vost = vhost == ":random:"
```
davidszotten: i like brackets for emphasis when assigning the outcome of a comparison:

`use_random_vost = (vhost == ":random:")`

nameko/testing/pytest.py
```diff
@@ -84,30 +84,51 @@ def rabbit_manager(request):
...
+
+    if use_random_vost:
+        vhost = "test_{}".format(
+            "".join(random.choice(string.lowercase) for _ in range(10))
+        )
+        amqp_uri = amqp_uri.replace(":random:", vhost)
```
davidszotten: above you parse the url to explicitly compare the value for vhost to `':random:'`, but here your run the replace over the whole string. seems inconsistent.

mattbennett:
@davidszotten this is ready for another round when you have time

nameko/testing/pytest.py
```diff
@@ -84,30 +85,51 @@ def rabbit_manager(request):
...
 
     # raise a runtime error if the test leaves any connections lying around
-    connections = get_rabbit_connections(conf['vhost'], rabbit_manager)
-    open_connections = [
-        conn for conn in connections if conn['state'] != "closed"
-    ]
-    if open_connections:
-        count = len(open_connections)
-        raise RuntimeError("{} rabbit connection(s) left open.".format(count))
+    try:
```
noisyboiler: this is a weird pattern... it's not clear if any of your instructions raise exceptions or if it's just your own `RuntimeError` that you're handling.

Anything that doesn't raise an Exception should be taken out of the `try` block.

Maybe you could remove the `vhosts` explicitly based on the count rather than use the `finally` clause from the `Exception` that you raise yourself? And don't you have to remove the `vhost` if it's a random one or not?

Also, would be useful to know which of the connections were left open in your exception msg.

is there an option to check for `conn['state'] == "open"` because this is ambiguous if there are other states.

does the `rabbit_manager` object not have an API itself to return connections, as it has one for `delete`? Else could you move your helper function logic to the `rabbit_manager`?

`use_random_vhost` appears to be handling a problem rather than solving it.

Why a `RuntimeError` rather than something more specific?

noisyboiler: sorry, that was a lot of questions!

nameko/testing/pytest.py
```diff
@@ -84,30 +85,51 @@ def rabbit_manager(request):
...
+
+    conf = {
+        'AMQP_URI': amqp_uri,
+        'username': username,
+        'vhost': vhost
+    }
 
     yield conf
 
     pools.reset()  # close connections in pools
```
noisyboiler: it would feel better if this action raised something like a `ResetError` rather than raising a `RuntimeError` based on your count.

test/testing/test_pytest_plugin.py
```diff
@@ -24,6 +25,34 @@ def test_rabbit_manager(rabbit_manager):
     assert "/" in [vhost['name'] for vhost in rabbit_manager.get_all_vhosts()]
 
 
+def test_rabbit_config_random_vhost(testdir):
+
```
noisyboiler: why do you leave blank lines in your test cases? you don't in the core modules.

test/testing/test_pytest_plugin.py
```diff
@@ -24,6 +25,34 @@ def test_rabbit_manager(rabbit_manager):
...
+    testdir.makepyfile(
+        """
+        def test_rabbit_config(rabbit_config):
+            assert "specified_vhost" in rabbit_config['AMQP_URI']
+        """
+    )
+    result = testdir.runpytest(
+        "--amqp-uri", "amqp://guest:guest@localhost:5672/specified_vhost"
+    )
+    assert result.ret == 0
```
noisyboiler: i'm unfamiliar with this; what is the `.ret` telling you? is a negative assertion strong enough to tell us that this test is doing something useful?

tox.ini
```diff
@@ -31,7 +31,7 @@ commands =
     lib: pip install --editable .[dev]
     lib: make flake8
     lib: make pylint
-    lib: make test_codebase
+    lib: make test_lib
```
noisyboiler: is this essentially `test_self`? `lib` is more open to interpretation than `codebase`.

nameko/testing/pytest.py
mattbennett: Did you mean to put this comment on a different line? ``pools.reset`` just closes any connections that are held open in the kombu connection pools.

nameko/testing/pytest.py
mattbennett: I think you've misunderstood how try/finally flow control works. I'm not handling the ``RuntimeError`` here.

I want to raise the exception if any connections are left over. This exception will bubble up and stop the execution of the test. The ``finally`` is there because, whether I throw the exception or not, I should always delete any vhost the test created. In the case where the user provided a vhost (i.e. it's not a random one) we assume they don't want it destroyed.

Regarding your other points:

* Yep, agree that the error should identify the connection. I will add.
* Checking ``conn['state'] == "open"`` isn't sufficient because there are other states that should also be considered "lying around"
* ``rabbit_manager.get_connections`` gets connections on all vhosts. ``get_rabbit_connections`` filters that response. 
* On ``RuntimeError`` vs something else... was here before this PR but I'm happy to raise something more specific if we can think of something. Maybe ``CleanupError``?

test/testing/test_pytest_plugin.py
mattbennett: This is how you use the `testdir` fixture. ``runpytest` runs pytest in a subprocess and it will discover any test files you've written into the "testdir".

``result.ret`` is the return value of the subprocess, so "0" means my test passed.

tox.ini
mattbennett: I think they're both as vague as each other really. We already had this debate in another PR and settled on ``lib``, I just missed this reference.

test/testing/test_pytest_plugin.py
mattbennett: Sometimes I do. If the next section is a reasonably big block or particularly distinct I'll usually give it some breathing space.

mattbennett:
@davidszotten @noisyboiler this is ready, finally

test/testing/test_utils.py
```diff
@@ -98,51 +97,6 @@ class ServiceY(object):
...
-    assert vhost in get_active_vhosts()
-
-    rabbit_manager.delete_vhost(vhost)
-    assert vhost not in get_active_vhosts()
-
-    reset_rabbit_vhost(vhost, username, rabbit_manager)
-    assert vhost in get_active_vhosts()
-
-
-def test_reset_rabbit_vhost_errors():
```
davidszotten: could you remind me why we no longer need these?

mattbennett: The `reset_rabbit_vhost` helper doesn't exist anymore. We either create and later delete a random vhost, or use whichever was specified and don't modify it.

davidszotten: ah yes

davidszotten:
this probably warrants a line in the changelog since we are changing default behaviour when using our helpers to test apps

davidszotten:
other than that i'm happy

davidszotten:
:+1:

PR #289, opened 2015-10-23
Added support for 'config' option in nameko shell

astawiarski:
Beside adding the '--config' to arg parser, I also extended tests to verify that 'serializer' attribute in config is picked up and ensured that proper keys from config are asserted
I used separate shell-config.yaml with different endpoint than the default in run-config.yaml since shell tests connect to rabbitmq


mattbennett:
This is great, thanks for the contribution!

I'm happy for the ``AMQP_URI`` in the ``run-config.yaml`` file to change though, to avoid the need for a separate ``shell-config.yaml`` file. Please also rename it to just ``config.yaml``.

davidszotten:
could we share the code that creates the config object based on the yaml file and or command line arguments to keep the two copies from diverging?

astawiarski:
@mattbennett : I made "shell" and "run" more consistent and both are now using config.yaml

@davidszotten : Good idea, I will improve it on separate branch since it may involve more changes (unless you prefer a larger pull here?). Initially I wanted to just add parsing to main.main() , but that won't sit well with current tests that are calling parse_args() directly. So I'll most likely try with argparse.Action .

mattbennett:
@astawiarski Excellent, thanks. To round out this PR please add release notes to CHANGES and also add yourself to the bottom of the contributors file.

Format for unreleased changes is as follows:

```
Version 2.2.1
-------------

Released PENDING

* Release notes
* As bullets

...
```

We follow semantic versioning and this is a backwards-compatible change so the version should be 2.2.1.

Re: @davidszotten's comment. I think ``main.main`` should parse the yaml file and pass the result into ``run.main`` and ``shell.main``. These functions can then add their required defaults if no config object was provided. I've not looked at the test fallout though and I'd say it's of marginal benefit.

astawiarski:
I had a look at unifying that config parser and there is more to it indeed. Tests for `shell` and `run` would need to change. I just updated contributors and changelog.

mattbennett:
:+1: Great stuff. Thank you @astawiarski!

PR #290, opened 2015-10-28
"pyamqp" should be suggest way of defining the protocol for rabbitmq

astawiarski:
Kombu will try to use librabbitmq if available and when using nameko on standalone basis (without eventlet and monkey patching). This will throw exception due to incompatibility between the two.

Using "pyamqp" instead of "amqp" will force Kombu to always use the pure Python implementation and not attempt to use librabbitmq.

"amqp" is an alias for "pymyqp" anyway

### DIFF
--- a/docs/key_concepts.rst
+++ b/docs/key_concepts.rst
@@ -82,6 +82,8 @@ Each worker executes in its own greenthread. The maximum number of concurrent wo
 
 Workers are stateless so are inherently thread safe, but dependencies should ensure they are unique per worker or otherwise safe to be accessed concurrently by multiple workers.
 
+Unfortunately, many C-extensions that are using sockets and that would normally be considered thread-safe may not work with greenthreads. Among them `librabbitmq <https://pypi.python.org/pypi/librabbitmq>`_, `MySQLdb <http://mysql-python.sourceforge.net/MySQLdb.html>` and others.
+
 .. _extensions:
 
 Extensions
--- a/docs/spelling_wordlist.txt
+++ b/docs/spelling_wordlist.txt
@@ -33,6 +33,7 @@ deserialized
 deserializers
 deserializes
 deserializing
+e
 entrypoint
 entrypoints
 eventlet
@@ -41,6 +42,7 @@ freenode
 greenthread
 greenthreads
 homebrew
+i
 init
 initialize
 initialized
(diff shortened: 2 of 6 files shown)


mattbennett:
Great catch.

We should probably draw attension to the C extension issue in http://nameko.readthedocs.org/en/stable/key_concepts.html#concurrency and reference it from the "similar projects" page.

I'm happy to merge this as-is and create an issue to improve the docs re eventlet.

astawiarski:
Added a few more paragraphs. There is probably much more to it, but I hope it will be enough for developers to understand potential issues and not crawl through code with debugger in hand :)

docs/spelling_wordlist.txt
```diff
@@ -33,6 +33,7 @@ deserialized
 deserializers
 deserializes
 deserializing
+e
```
mattbennett: You shouldn't need to do this. These are from "i.e.", right? Travis correctly considers it a word.

Are you developing locally on OS X? If so, the master branch of pyenchant has more up-to-date dictionaries. It hasn't been released though :(

astawiarski: Yes, it was from "i.e", but it was failing on travis and on my local (ubuntu/mint). I can revert those.

docs/key_concepts.rst
```diff
@@ -82,6 +82,8 @@ Each worker executes in its own greenthread. The maximum number of concurrent wo
 
 Workers are stateless so are inherently thread safe, but dependencies should ensure they are unique per worker or otherwise safe to be accessed concurrently by multiple workers.
 
+Unfortunately, many C-extensions that are using sockets and that would normally be considered thread-safe may not work with greenthreads. Among them `librabbitmq <https://pypi.python.org/pypi/librabbitmq>`_, `MySQLdb <http://mysql-python.sourceforge.net/MySQLdb.html>` and others.
```
mattbennett: Please change

"Unfortunately, many C-extensions" => "Note that many C-extensions"
"Among them" => "Among them are"

And then add

"Modules with C-extensions that are not "green" may be wrapped in an eventlet `thread-pool <http://eventlet.net/doc/threading.html>`."

docs/spelling_wordlist.txt
mattbennett: That is curious because we already have "i.e." in the master branch. Maybe something changed on Travis...

jessepollak:
Since progress here seems to have halted, I re-opened these changes as a new PR with the comments address in #361. Can we close this and merge that in? Thanks!

PR #291, opened 2015-11-02
Moved web exception handling to separate method to allow for overriding

kooba:
Also added examples of expected_exceptions and overriding response_from_exception. Moved response_from_result into class for consistency and ability to override.

This PR replaces https://github.com/onefinestay/nameko/pull/275 and is related to https://github.com/onefinestay/nameko/issues/284

Let me know what you think.


davidszotten:
Hi. As discussed on #275 i think this is a good approach. a few comments inline

docs/built_in_extensions.rst
```diff
@@ -108,6 +108,37 @@ The HTTP entrypoint is built on top of `werkzeug <http://werkzeug.pocoo.org/>`_.
     Date: Fri, 13 Feb 2015 14:58:48 GMT
 
 
+Exceptions can be raised safely if added to ``expected_exceptions``. You can also change exceptions formatting by overriding :meth:`~nameko.web.HttpRequestHandler.response_from_exception`:
```
davidszotten: "exception formatting" or possibly "exception's formatting". though maybe explicitly mention that this controls how exceptions are returned to a http caller

i'm not sure "safely" is appropriate here. do we write about `expected_exceptions` elsewhere?

docs/built_in_extensions.rst
```diff
@@ -108,6 +108,37 @@ The HTTP entrypoint is built on top of `werkzeug <http://werkzeug.pocoo.org/>`_.
...
+
+.. code-block:: shell
+
+    $ curl -i http://localhost:8000/expected_custom_exception
+    HTTP/1.1 400 BAD REQUEST
+    Content-Type: application/json
+    Content-Length: 96
+    Date: Thu, 06 Aug 2015 09:53:56 GMT
+
+    "{\"code\": 400, \"description\": \"This is invalid request.\", \"error\": \"INVALID_REQUEST\"}"
```
davidszotten: this looks double encoded. (might be related to a bug i saw elsewhere, when raising the exception. will comment there too)

docs/examples/http_exceptions.py
```diff
@@ -0,0 +1,49 @@
...
+from werkzeug.wrappers import Response
+
+
+class ApplicationError(Exception):
+    pass
+
+
+class HttpError(Exception):
+
+    def __init__(self, payload, status_code=None):
```
davidszotten: it seems slightly strange to me to have the exception class (and by extension, the service code) care about things like http status codes. i would expect that decision to be made in `response_from_exception`

`HttpError`, again seems to focus on mechanism rather than (as common with exception classes) _what_ went wrong.

maybe something like `InvalidAgruments` (not a great name, but along those lines)

docs/examples/http_exceptions.py
```diff
@@ -0,0 +1,49 @@
...
+    name = "http_exceptions_service"
+
+    @http('GET', '/expected_exception', expected_exceptions=ApplicationError)
+    def expected_exception(self, request):
+        raise ApplicationError("Invalid request")
+
+    @http('GET', '/expected_custom_exception',
+          expected_exceptions=HttpError)
+    def expected_custom_exception(self, request):
+        raise HttpError(json.dumps({
```
davidszotten: you are calling `json.dumps` both here and in `response_from_exception`. is this intentional?

docs/examples/test/test_examples.py
```diff
@@ -48,6 +49,33 @@ def test_advanced(self, container_factory, rabbit_config):
         assert res.status_code == 200
         assert res.text == 'payload'
 
+    def test_expected_exception(self, container_factory, rabbit_config):
+
+        from examples.http_exceptions import Service
+
+        container = container_factory(Service, rabbit_config)
```
davidszotten: i don't this (or indeed a bunch of other tests in this module) needs a `rabbit_config`

docs/examples/test/test_examples.py
```diff
@@ -48,6 +49,33 @@ def test_advanced(self, container_factory, rabbit_config):
...
+        container.start()
+
+        res = requests.get(self.base_url + "/expected_custom_exception")
+        assert res.status_code == 400
+        assert res.headers['Content-Type'] == 'application/json'
+        assert json.loads(res.json()) == {
```
davidszotten: this is double-decoding and sounds to me like we're doing something wrong

docs/built_in_extensions.rst
kooba: Changed wording slightly. This is first mention of `expected_exceptions` in the docs.

docs/built_in_extensions.rst
kooba: My bad, this was a mistake in double encoding in the example.

docs/examples/http_exceptions.py
kooba: This is fixed now. Doing it in `response_from_exception` only.

docs/examples/test/test_examples.py
kooba: Changed all HTTP tests to user `web_config` and `web_session`

docs/examples/http_exceptions.py
kooba: Once `response_from_exception` is overridden, client's only option to raise different http errors would be to carry that information in exception. To be more explicit other helper exceptions that inherit from `HttpError` could be created by client like `InvalidAgruments (422)`, `EntityNotFound (404)` or `BadRequest (400)` etc...

docs/built_in_extensions.rst
davidszotten: @mattbennett do you recall if we purposely didn't mention `expected_exceptions` because we weren't ready to commit to it as a public api?

docs/examples/http_exceptions.py
davidszotten: such subclasses sound reasonable. and indeed, i would expect those classes to decide on status codes, not the service code raising the exception

in your example, i would expect something closer to `raise HttpError("This is (an?) invalid request")` or whatever (ideally some more detail, e.g. invalid argument foo), and have the class together with `response_from_exception` construct the payload

btw, how come the status code is also included in the payload?

kooba: I simplified example of raising custom http exception. Adding status code in payload is a common practice (twilio, eventbrite, uber) but I removed it from the example to avoid any confusion.  
Please review.

docs/built_in_extensions.rst
mattbennett: @davidszotten I think that's probably the case. I intended for it to be a public API on all entrypoints, but we only had experience of using it in RPC. This PR obviously solves some of that.

The docs need a bit of a revamp. The "built-in extensions" content is pretty light. I always intended  to fill in the detail on the underlying implementations, and I think we also need somewhere to document these features that are common to many implementations (``expected_exceptions``, ``sensitive_variables``, the whole concept of communication between extensions etc.)

docs/built_in_extensions.rst
```diff
@@ -108,6 +108,37 @@ The HTTP entrypoint is built on top of `werkzeug <http://werkzeug.pocoo.org/>`_.
     Date: Fri, 13 Feb 2015 14:58:48 GMT
 
 
+Expected exceptions can be added to ``expected_exceptions``. This will result in 400 HTTP error to be returned to the client. You can also change exception's formatting by overriding :meth:`~nameko.web.HttpRequestHandler.response_from_exception`:
```
mattbennett: This is a bit self-referential somewhere describing why an exception might be "expected" at all (the name is a bit of an oxymoron). As discussed below I think we should have a separate section of the docs for this, and reference it from here. I am happy to add that as a separate PR.

docs/examples/http_exceptions.py
```diff
@@ -0,0 +1,53 @@
...
+class HttpEntrypoint(HttpRequestHandler):
+    def response_from_exception(self, exc):
+        if isinstance(exc, HttpError):
+            response = Response(
+                json.dumps(exc.payload),
+                status=exc.status_code,
+                mimetype='application/json'
+            )
+            return response
+        else:
```
mattbennett: The ``else`` is unnecessary here.

docs/built_in_extensions.rst
davidszotten: agree that the name isn't great - we should consider alternatives

also, note that for for the rpc entrypoint, the concept of this set of exceptions was purely internal to the service, whereas for the http, we are making it part of the service api (since it affects the http status; something we don't have for rpc atm)

docs/examples/http_exceptions.py
```diff
@@ -0,0 +1,53 @@
...
+
+class HttpError(Exception):
+
+    def __init__(self, message, error, status_code):
+        self.payload = {
+            'error': error,
+            'message': message
+        }
+        self.status_code = status_code
+        super(HttpError, self).__init__(self.payload)
```
mattbennett: It seems unconventional to pass a dict into ``Exception.__init__``. I think
```python
super(HttpError, self).__init__("{}: {}".format(error, message))
```
would be more normal. @davidszotten do you have an opinion on this?

davidszotten: i would probably restructure the whole thing as something like

```python
class HttpError(Exception):
    error_code = 'BAD_REQUEST'  # if we want to allow raising "bare" HttpErrors
    status_code = 400


class InvalidArgumentsError(HttpError):
    error_code = 'INVALID_ARGUMENTS'


def response_from_exception(self, exc):
    if isinstance(exc, HttpError):
        response = Response(
            json.dumps({
                'error': exc.error_code,
                'message': unicode(exc),  # do we need to worry about this failing?
            })  # or dumps failing?
            status=exc.status_code,
            mimetype='application/json'
        )
        return response
    else:
        return HttpRequestHandler.response_from_exception(self, exc)


def service_code():
    raise InvalidArgumentsError("Argument `foo` is required.")
```

docs/built_in_extensions.rst
kooba: Removed mention of expected_exceptions. This will be covered by #293

docs/examples/http_exceptions.py
kooba: Incorporated changes this suggestion.

docs/examples/http_exceptions.py
kooba: Removed.

kooba:
@davidszotten - Changed to use `safe_for_serialization`

davidszotten:
awesome! :+1:  sorry it's been a bit of a slug, but we got there in the end. please update `CHANGES` (for the currently pending version, and add yourself to `contributors`

kooba:
@davidszotten Thank you! This was a great experience. `CHANGES` and `CONTRIBUTORS.txt` updated.

PR #298, opened 2015-12-14
upgrade pylint

davidszotten:
pinned pylint doesn't pin astroid but isn't compatible with the latest version
may as well just upgrade pylint


mattbennett:
:+1:

PR #301, opened 2015-12-16
promote mailing list instead of irc

davidszotten:


docs/spelling_wordlist.txt
```diff
@@ -38,6 +38,7 @@ entrypoints
 eventlet
 extensibility
 freenode
+google
```
mattbennett: Shows how out of date these dictionaries are!

mattbennett:
:+1:

davidszotten:
we need to cut a new release before this shows up on e.g. readthedocs, but have updated the channel topic for now

PR #302, opened 2016-01-05
First do task and then wait for the next one

Sparkycz:
I think this change is needed for testing and observation of Timer service after deploy on production server when there is defined long time between tasks.

### DIFF
--- a/nameko/timer.py
+++ b/nameko/timer.py
@@ -2,7 +2,6 @@ from __future__ import absolute_import
 from logging import getLogger
 import time
 
-from eventlet import Timeout
 from eventlet.event import Event
 
 from nameko.extensions import Entrypoint
@@ -13,11 +12,12 @@ _log = getLogger(__name__)
 class Timer(Entrypoint):
     def __init__(self, interval, eager=False):
         """
-        Timer entrypoint implementation. Fires every :attr:`interval`
-        seconds.
+        Timer entrypoint. Fires every `interval` seconds or as soon as the previous
+        worker completes if that took longer.
 
-        The implementation sleeps first,i.e. does not fire at time 0.
-        If you want to fire at time 0 (fire first) set :attr:`eager`=True
+        The default behaviour is to wait `interval` seconds before firing for the
+        first time. If you want the entrypoint to fire as soon as the service starts,
+        pass `eager=True`.
 
         Example::
 
@@ -33,7 +33,8 @@ class Timer(Entrypoint):
         """
         self.interval = interval
         self.eager = eager
-        self.should_stop = Event()
+        self.should_stop = False
+        self.event = Event()
         self.gt = None
 
     def start(self):
@@ -42,7 +43,7 @@ class Timer(Entrypoint):
 
     def stop(self):
         _log.debug('stopping %s', self)
-        self.should_stop.send(True)
+        self.should_stop = True
         self.gt.wait()
 
     def kill(self):
@@ -51,25 +52,28 @@ class Timer(Entrypoint):
 
     def _run(self):
         """ Runs the interval loop. """
-
         sleep_time = 0 if self.eager else self.interval
 
         while True:
             # sleep for `sleep_time`, unless `should_stop` fires, in which
             # case we leave the while loop and stop entirely
-            with Timeout(sleep_time, exception=False):
-                self.should_stop.wait()
-                break
+            time.sleep(sleep_time)
 
             start = time.time()
 
             self.handle_timer_tick()
+
+            self.event.wait()
+            self.event.reset()
 
             elapsed_time = (time.time() - start)
 
             # next time, sleep however long is left of our interval, taking
             # off the time we took to run
             sleep_time = max(self.interval - elapsed_time, 0)
+
+            if self.should_stop:
+                break
 
     def handle_timer_tick(self):
         args = ()
@@ -78,9 +82,14 @@ class Timer(Entrypoint):
         # Note that we don't catch ContainerBeingKilled here. If that's raised,
         # there is nothing for us to do anyway. The exception bubbles, and is
         # caught by :meth:`Container._handle_thread_exited`, though the
-        # triggered `kill` is a no-op, since the container is alredy
+        # triggered `kill` is a no-op, since the container is already
         # `_being_killed`.
-        self.container.spawn_worker(self, args, kwargs)
+        self.container.spawn_worker(
+            self, args, kwargs, handle_result=self.handle_result)
+
+    def handle_result(self, worker_ctx, result, exc_info):
+        self.event.send(result, exc_info)
+        return result, exc_info
 
 
 timer = Timer.decorator
--- a/nameko/timer.py
+++ b/nameko/timer.py
@@ -11,12 +11,13 @@ _log = getLogger(__name__)
 
 
 class Timer(Entrypoint):
-    def __init__(self, interval):
+    def __init__(self, interval, eager=False):
         """
-        Timer entrypoint implementation. Fires every :attr:`self.interval`
+        Timer entrypoint implementation. Fires every :attr:`interval`
         seconds.
 
-        The implementation sleeps first, i.e. does not fire at time 0.
+        The implementation sleeps first,i.e. does not fire at time 0.
+        If you want to fire at time 0 (fire first) set :attr:`eager`=True
 
         Example::
 
@@ -31,6 +32,7 @@ class Timer(Entrypoint):
 
         """
         self.interval = interval
+        self.eager = eager
         self.should_stop = Event()
         self.gt = None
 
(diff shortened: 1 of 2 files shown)


davidszotten:
this used to be the behaviour, but it was changed for 2.0 to wait first. @mattbennett do you recall the reason?

Sparkycz:
Thanks for the fast response.. I wonder the reason because when i will want the first wait i can check first round and break it by own way

davidszotten:
in the meantime, you could of course subclass the timer to change the behaviour:

```
class MyEagerTimer(Timer):
    def _run(self):
        self.handle_timer_tick()
        super(MyTimer, self)._run()
```

Sparkycz:
There's no rush. I solved it by other way. Thank you ;-)
I just think it's strange..

mattbennett:
This might not be the entire story, but I remember all the "timer" entrypoints firing simultaneously as soon as a service started, which didn't seem sensible.

I think that both behaviours could be considered "correct" and it probably just needs documenting better. Or giving it an `eager=True` parameter.

Sparkycz:
I added parameter eager how Matt wrote. What do you about this change? :)
There is still "sleep first" mode by default..

nameko/timer.py
```diff
@@ -50,7 +52,17 @@ def kill(self):
     def _run(self):
         """ Runs the interval loop. """
 
-        sleep_time = self.interval
+        def _process_task():
```
davidszotten: rather than adding another function here, maybe just bake this into `handle_timer_tick`. However, we might want to think about the sleep time bug #303

Sparkycz:
it's not final solution yet..

nameko/timer.py
```diff
@@ -11,12 +10,13 @@
...
-    def __init__(self, interval):
+    def __init__(self, interval, eager=False):
         """
-        Timer entrypoint implementation. Fires every :attr:`self.interval`
+        Timer entrypoint implementation. Fires every :attr:`interval`
         seconds.
 
-        The implementation sleeps first, i.e. does not fire at time 0.
+        The implementation sleeps first,i.e. does not fire at time 0.
+        If you want to fire at time 0 (fire first) set :attr:`eager`=True
```
mattbennett: I think we should take this opportunity to improve the docs, including explicitly documenting the behaviour you've just fixed.

``` python
"""
Timer entrypoint. Fires every `interval` seconds or as soon as the previous
worker completes if that took longer.

The default behaviour is to wait `interval` seconds before firing for the
first time. If you want the entrypoint to fire as soon as the service starts,
pass `eager=True`.
"""
```

Sparkycz: Ok, I'll update it. thanks

nameko/timer.py
```diff
@@ -31,7 +31,9 @@ def tick(self):
 
         """
         self.interval = interval
-        self.should_stop = Event()
+        self.eager = eager
+        self.should_stop = False
+        self.event = Event()
```
mattbennett: Maybe call this `worker_complete`?

nameko/timer.py
```diff
@@ -76,9 +81,14 @@ def handle_timer_tick(self):
...
         # caught by :meth:`Container._handle_thread_exited`, though the
-        # triggered `kill` is a no-op, since the container is alredy
+        # triggered `kill` is a no-op, since the container is already
         # `_being_killed`.
-        self.container.spawn_worker(self, args, kwargs)
+        self.container.spawn_worker(
+            self, args, kwargs, handle_result=self.handle_result)
+
+    def handle_result(self, worker_ctx, result, exc_info):
+        self.event.send(result, exc_info)
```
mattbennett: You don't actually need to pass `result` and `exc_info` in here. It would be sufficient to send `True`.

nameko/timer.py
```diff
@@ -49,26 +51,29 @@ def kill(self):
...
+            self.event.wait()
+            self.event.reset()
+
             elapsed_time = (time.time() - start)
 
             # next time, sleep however long is left of our interval, taking
             # off the time we took to run
             sleep_time = max(self.interval - elapsed_time, 0)
 
+            if self.should_stop:
```
mattbennett: Not a big deal but this check can go higher in the loop. There's no need to reset the event or work out the new `sleep_time` if you're going to exit the loop.

nameko/timer.py
Sparkycz: of course, thanks

nameko/timer.py
Sparkycz: It can be..

nameko/timer.py
Sparkycz: Yes, but it's needed for calculating of coverage in Travis's tests.. (When i move this lines higher, under the self.event.reset(), the lines 69-74 are not used in tests)

test/test_timers.py
```diff
@@ -8,27 +7,33 @@
 from nameko.testing.utils import wait_for_call
 
 
-def test_provider():
+def test_provider(interval=0.1, eager=False):
```
mattbennett: @davidszotten, is this valid/recommended pytest usage?

test/test_timers.py
```diff
@@ -8,27 +7,33 @@
...
-            timer.stop()
+        timer.handle_result(None, None, None)
+        timer.stop()
 
     # the timer should have stopped and should only have spawned
     # a single worker
-    spawn_worker.assert_called_once_with(timer, (), {})
+    spawn_worker.assert_called_once_with(
+        timer, (), {}, handle_result=timer.handle_result)
     assert timer.gt.dead
```
mattbennett: The tests for this module have always been a mess. Rather than a broad `test_provider` test I'd prefer break down each method and test its code paths individually, and then have a single "end-to-end" test verifying all the pieces work together.

Take a look at https://github.com/mattbennett/nameko-sentry/blob/master/test_nameko_sentry.py to see tests that I've written in this way on a similar extension. Testing the `_run` method separately makes things much simpler.

test/test_timers.py
davidszotten: the usual way to get this behaviour is `@pytest.mark.parametrize`

nameko/timer.py
```diff
@@ -49,26 +52,29 @@ def kill(self):
...
     def _run(self):
         """ Runs the interval loop. """
-
-        sleep_time = self.interval
+        sleep_time = 0 if self.eager else self.interval
 
         while True:
             # sleep for `sleep_time`, unless `should_stop` fires, in which
             # case we leave the while loop and stop entirely
-            with Timeout(sleep_time, exception=False):
```
davidszotten: how come you are changing this? the old behaviour is explained in the comment just above and doesn't work with the replacement.

basically, if the service is stopped while the timer is sleeping, we don't want to wait until it's done sleeping before tearing down the timer entrypoint

nameko/timer.py
```diff
@@ -49,26 +52,29 @@ def kill(self):
...
-                self.should_stop.wait()
-                break
+            time.sleep(sleep_time)
 
             start = time.time()
 
             self.handle_timer_tick()
 
+            self.worker_complete.wait()
```
davidszotten: at this point an eventlet `Semaphore` (remember to init with zero, not the default 1) might be more idiomatic

nameko/timer.py
davidszotten: is that because we are missing a test that goes all the way around this loop?

nameko/timer.py
mattbennett: @davidszotten is correct. The thread will probably spend most of its time asleep, and we have to be able to wake it when `stop` is called. We should have a test for this too, something like:

``` python
def test_stop_while_sleeping(container_factory):

    class Service(object):
        name = "service"

        @timer(interval=5)
        def tick(self):
            pass

    container = container_factory(Service, {})
    container.start()

    # raise a Timeout if the container fails to stop within 1 second
    with eventlet.Timeout(1):
        container.stop()
```

Sparkycz: Uff.. Yes, i understand

I replaced the eventlet.Timeout() by my solution because self.should_stop.wait() wasn't called correctly under "with Timeout(..):".

I will think of an other solution for solve the problem.
thx

jessepollak:
Wanted to follow up on this PR with two questions:

1. @mattbennett and @davidszotten if the comments in this PR were addressed, would this change be pulled into master and shipped? If so, what priority level is the additional functionality here on a scale of 1-10?
2. @Sparkycz do you have any interest in finishing out the work to get this shipped? If not, would you mind letting us know so we can either close the PR or delegate the work to another contributor?

Thanks!

Sparkycz:
wait a few days pls. i'm currently on vacation..

test/test_timers.py
mattbennett: This change should be considered optional to for this PR. I'd prefer to land these fixes rather than force them to sit behind stylistic cleanup of old tests.

mattbennett:
@jessepollak 

Yes, although I've made the test cleanup optional. I think my former-self was being overzealous there (sorry @Sparkycz!) It's a low priority fix.

PR #307, opened 2016-01-29
Mock(spec=foo) -> create_autospec(foo)

davidszotten:
The former doesn't work as expected, see
http://www.szotten.com/david/watch-out-for-mockspec.html


davidszotten:
@mattbennett it looks like [mock](https://pypi.python.org/pypi/mock)'s started moving again. how do we feel about bumping the minimum required version (would fix the issue here)

mattbennett:
@davidszotten, yep I'm fine with bumping the minimum required version.

davidszotten:
i'm having second thoughts on bumping mock, since nameko itself, and not just its test suite depends on mock. not sure i want to mandate mock version in order to use nameko

mattbennett:
I think you can just bump the version of mock required in `[dev]` requirements

davidszotten:
unfortunately not, since one of the issues is in `nameko/testing/pytest.py` which is part of the main codebase

one option (that we've considered previously) is to package the test helpers separately, though that has other drawbacks instead

mattbennett:
How far do we have to move mock to fix this issue? I don't want to yak-shave our way into having to repackage the test helpers. I think the sensible options here are:

1. Abandon this PR, and let the spec remain "loose"
2. Upgrade our mock library to get the new feature

Upgrading seems fine to me. Mock is quite a well behaved library, I don't think there will be many folks out there who are _required_ to use an older version.

davidszotten:
i'd like to investigate the state of mock under the new maintainers before forcing an upgrade. don't mind if we leave this open or close in the mean time

mattbennett:
OK, that seems sensible. Let's leave this open for now

mattbennett:
Mock 2.0.0 is stable and has been in use with our nameko services for some time, so we can bring this up to date and merge it now

davidszotten:
brought up to date. didn't need to bump mock all the way, so left "oldest" where i could

mattbennett:
@davidszotten, we missed the boat here.

Can you bring up to date again? Then it's good to land, right?

mattbennett:
👍

PR #311, opened 2016-03-14
enable reliable broadcast-mode events by making queue names predictable

mattbennett:


davidszotten:
hm. not sure what i think about this. will try to find some time to consider properly

concerned about backwards incompat and "non-safe" default, but also more generally about having semantics that make sense. e.g. if we have distinct use cases for the current and this new mode of operation, maybe we should allow both. or is this "obviously" the way one would like to use broadcast?

mattbennett:
This change doesn't prevent the "old" behaviour so in a sense it does allow both modes of operation. To summarise:

**Old `BROADCAST` behaviour**

* `reliable_delivery` not possible (queues set to exclusive and auto-delete)
* queue names appended with a uuid created on service start 

Therefore any messages sent to a previous instance of the service are not seen by new instances (e.g. after a restart)

**New `BROADCAST` behaviour**

* `reliable_delivery` is possible, but not mandatory
* you can supply a `differentiator` callable to provide the value appended to the queue name
* default differentiator uses `socket.gethostname` so queues will be reused between service restarts

It is therefore possible to use the old behaviour by passing `uuid.uuid4` (or similar) as the `differentiator` and setting `reliable_delivery=False`.

I would argue that both modes of operation are valuable. Semantically, the old behaviour is "broadcast to all service instance _processes_", and the default new behaviour is "broadcast to all service instance _deployments_".

If we have to pick one the new behaviour is more useful, and we would want to keep the ability to provide a differentiator (for example if `socket.gethostname` didn't return something sane or useful) which basically you get the old mode for free anyway. Though we might choose not to document it. 

This change is backwards incompatible in the sense that `reliable_delivery` in broadcast mode now defaults to `True`. Actually this brings it in line with the other modes, which also use reliable delivery by default. #104 has been open for some time to make us "safe" (i.e. non-reliable) by default, but I think this is a separate concern.

davidszotten:
so have been thinking a bit about this, but not sure i can articulate my concerns very well

i think there are a few independent issues (which keep getting muddled in my head). an attempt

1: changing the default behaviour: i guess i'm not convinced that the new behaviour is "better _enough_" than the old one to change the default. esp. given that (#104) we might eventually end up going the other way on this. so maybe we it would be preferable to _enable_ the new behaviour but leave the old default

2: i'm not so keen on this api for supplying a `differentiator`. 
a) it's an argument that only applies/means something for certain combinations of other arguments
b) it feels a bit limiting, and would maybe be better exposed by the extension using a method to produce the queue name (which can be overridden by a subclass for your use-case, either in nameko or in your app/shared lib for your projects)
c) defaulting to `gethostname` can lead to strange behaviour if there are multiple processes running with the same service on the same host. i guess when you call it _deployments_ there's an implicit assumption that this isn't the case

mattbennett:
OK thanks. Let me try to address these concerns:

1. There are two things that are backwards incompatible here:
 * a) `reliable_delivery` for `BROADCAST` mode now defaults to `True`
 * b) the "differentiator" changes from `uuid.uuid4` to `socket.gethostname` (so we go from _process_ broadcast to _deployment_ broadcast).

  For 1.a, we could keep the old behaviour by mapping a `handler_type` to a default value for `reliable_delivery`. But would then have to explain why it's like that, and I don't think there is a good reason, except for maintaining backwards compatibility. We would want to remove the anomaly eventually, so all we're doing is postponing the backwards-incompatible change. We include #104 in these changes too, which would compress the two into a single backwards-incompatible release.

  1.b can be solved by changing the default `differentiator` (or its equivalent) back to `uuid.uuid4`. We can definitely do this, but there's an argument that we're making things less useful for the sake of backwards-compatibility.

2. I understand all of these concerns. I will have a crack at an implementation using a method that can be overridden. This would be more in line with the nameko way of doing things. Being able to change the entire queue name is an interesting idea.

Regarding `gethostname` -- I wanted something that would be unique to the service instance but survive restarts, and for _most people_ I figured hostname was a reasonable guess. It makes for nice readable queue names too ;) It won't always apply, but that's why you can specify it yourself.

It turns out that broadcast queues aren't exclusive, but they should be, which would also let us detect the situation where multiple service instances end up with the same broadcast queue name.

mattbennett:
There's also _hidden option number three_: explicitly add `PROCESS_BROADCAST` and `DEPLOYMENT_BROADCAST` as separate modes. Although in the latter case you have to be able to provide a way to differentiate one deployment from another, since `gethostname` won't always be valid.

mattbennett:
OK, this version is better I think. I've made the following changes:

* Removed `differentiator` in favour of a `broadcast_identity` property, which can be overridden
* Restored the default to a uuid that is created when the service starts
* Added docs explaining this behaviour and some alternatives

The behavioural difference now is that `reliable_delivery` defaults to `True` in broadcast mode, and that we emit a warning in this case rather than raise an exception. Since we used to raise, there can't be any working code out there that will now behave differently.

I also made the broadcast queues exclusive, although it turns out you can't easily detect the rejected connection so I didn't bother going further.

davidszotten:
as discussed, let's raise instead of just warning in the default implementation of `broadcast_identifier` if `reliable_delivery` is set to `True`. 

This stops a naive user from shooting themselves in the foot by just enabling `broadcast` and leaving the rest as default, and also provides a loud warning to anyone upgrading about the backwards incompatible change (allowing them to explicitly set it back if they need)

CHANGES
```diff
@@ -6,13 +6,19 @@ are in form of *headline.major.minor* numbers. Backwards-compatible changes
...
-* Add support for loading configuration file in ``nameko shell`` via ```--config`` option
-* Changed :class:`~nameko.web.handlers.HttpRequestHandler` to allow override how web exceptions are handled
+* Add support for loading configuration file in ``nameko shell`` via ``--
+  config`` option
+* Changed :class:`~nameko.web.handlers.HttpRequestHandler` to allow override
+  how web exceptions are handled
+* Enabled reliable delivery on broadcast events when combined with a custom
+  :py:attr:`~nameko.events.EventHandler.broadcast_identity`. Reliable delivery
+  now defaults to enabled for all handler types. It must be explicitly turned
+  off with broadcast mode unless override the default `broadcast_identity`.
```
davidszotten: since you are bumping the version here, i think you should also add a changelog entry about the bumped minimum eventlet requirements (due to eventlet removing them from pypi)

mattbennett: Quite right. I will do that

nameko/events.py
```diff
@@ -29,14 +29,14 @@ def bar(event_data):
 
 """
 from __future__ import absolute_import
-from logging import getLogger
```
davidszotten: have you switched linter? :p

nameko/events.py
```diff
@@ -183,10 +168,69 @@ def __init__(self, source_service, event_type, handler_type=SERVICE_POOL,
         super(EventHandler, self).__init__(
             queue=None, requeue_on_error=requeue_on_error)
 
+    @property
```
davidszotten: how come this is a `property`?

nameko/events.py
```diff
@@ -183,10 +168,69 @@ def __init__(self, source_service, event_type, handler_type=SERVICE_POOL,
...
     def setup(self):
         _log.debug('starting %s', self)
 
-        # handler_type determines queue name
+        cls = type(self)
+
+        if (
+            self.reliable_delivery and
+            self.handler_type is BROADCAST and
+            cls.broadcast_identifier is EventHandler.broadcast_identifier
```
davidszotten: can't we just raise this _in_ the default implementation of `broadcast_identifier`?

nameko/events.py
```diff
@@ -197,19 +241,24 @@ def setup(self):
...
-                                                      uuid.uuid4().hex)
+                                                      broadcast_identifier)
 
         exchange = get_event_exchange(self.source_service)
 
         # auto-delete queues if events are not reliably delivered
         auto_delete = not self.reliable_delivery
+
+        # broadcast queues are exclusive
+        exclusive = self.handler_type is BROADCAST
```
davidszotten: docs: "Exclusive queues may only be accessed by the current connection, and are deleted when that connection closes.". that sounds incompatible with the example that uses hostname to share (re-use) broadcast queues. am i missing something here?

test/test_events.py
```diff
@@ -23,6 +22,13 @@ def queue_consumer():
         yield mock_ext.return_value
 
 
+@pytest.yield_fixture
+def logger():
+    with patch('nameko.events._log', autospec=True) as patched:
+        yield patched
+    patched.reset_mock()
```
davidszotten: isn't the patch is un-applied at this point? though this whole fixture may no longer be used

nameko/events.py
mattbennett: Hehe, yep. I now use `isort`. It's awesome!

nameko/events.py
mattbennett: Yes, we can. That'd be much better

nameko/events.py
mattbennett: I find myself using a `property` when the alternative method would be `get_foo()` with no arguments.

test/test_events.py
mattbennett: Correct on all counts. I'm surprised `patched.reset_mock` doesn't raise actually...

I will remove this fixture

nameko/events.py
davidszotten: think i prefer `get_foo()` to make it clear(er) to a "caller" that this is a function call. esp. since it might be doing non-trivial stuff

nameko/events.py
mattbennett: Ah. It turns out that `exclusive` implies `auto_delete` even if you request otherwise (and nothing complains if you supply both). I will change it so we only use `exclusive` when reliable delivery is off.

nameko/events.py
mattbennett: I don't think attribute access implies trivial. `property` gets used all the time to hide lazy creation of expensive objects, or provide a helpful interface on top of a moderately complex operation (e.g. https://github.com/onefinestay/nameko/blob/master/nameko/web/server.py#L64)

`obj.foo` is just a little bit cleaner than ``obj.get_foo()``

nameko/events.py
```diff
@@ -197,19 +235,26 @@ def setup(self):
...
                                                       self.event_type,
                                                       service_name,
                                                       self.method_name,
-                                                      uuid.uuid4().hex)
+                                                      broadcast_identifier)
 
         exchange = get_event_exchange(self.source_service)
 
-        # auto-delete queues if events are not reliably delivered
+        # queues are auto-deleted unless reliable delivery is requested
```
davidszotten: this comment (and the one for `exclusive`) don't add much info beyond the line of code they decorate. could we change it to explain _why_ we want this behaviour (or drop, though maybe at least include the point about `exclusive` silently overwriting the `auto_delete` behaviour)

nameko/events.py
```diff
@@ -183,10 +168,63 @@ def __init__(self, source_service, event_type, handler_type=SERVICE_POOL,
...
+        value out of a configuration file ::
+
+            @property
+            def broadcast_identifier(self):
+                return self.config['SERVICE_IDENTIFIER']  # or similar
+
+        Broadcast queues are exclusive to ensure that `broadcast_identifier`
+        values are unique.
+        """
+        if self.reliable_delivery and self.handler_type is BROADCAST:
```
davidszotten: i guess we don't need to check `handler_type` in here

nameko/events.py
mattbennett: Ah yes. I will try to improve it

nameko/events.py
mattbennett: Actually you do, because `self.broadcast_identifier` can be inspected at any time regardless of the handler type (and is, in fact, during `bind`)

davidszotten: ah. though that's quite surprising behaviour i think, and unless we can figure out how to `getmembers` or similar without executing descriptors, i'd (independently of previous concerns) prefer making this a regular method

mattbennett: I don't think it's strange to check both conditions are met before raising the error. Otherwise you're relying on it only being called in specific circumstances.

It doesn't matter if it's a method or a property. This would be weird:

```
>>> event_handler = EventHandler("svc", "evt")
>>> event_hander.get_broadcast_handler()
Traceback:
...
EventHandlerConfigurationError: You are using the default broadcast identifier
which is not compatible with reliable delivery...
```

It is less likely that you'd intentionally call `get_broadcast_handler` than you would `getmembers`, but I'm not sure that matters.

mattbennett: Alternatively we could add

``` python
if self.handler_type is not BROADCAST:
    return None
```

before the check. I think I prefer that. It's reasonable that `broadcast_identifier` is undefined outside of broadcast mode.

davidszotten: i think the strangeness is caused by a problem "higher up", namely trying to have a single class for both "service pool" and "broadcast" events (ignoring singleton for now)

assuming any change to that is out of scope for now

the initial implementation had a kwarg that only made sense for certain values of handler_type. having a method is less bad, but we still have this method that only make sense if the class was initialised with a certain parameter value

we are extending `EventHandler` and added a protocol which means it calls `broadcast_handler` if it's in broadcast mode. i think calling that method "out of turn" is fine to be "undefined behaviour". returning `None` as you suggest is fine

however, i think my (new) objection to the property decorator is that because we use `getmembers`, the method (or a new implementation in user code) gets called at unexpected times. so regardless of whether we can get this particular implementation of `broadcast_handler` to handle that, i'd prefer to stay away from descriptors until we can make our inspection not invoke them

mattbennett: I see your point about sharing a class for "service pool" and "broadcast" events. It's that way because they're (even now) very similar in implementation. It is only the queue name, and now the `exclusive` value, that changes.

I would advocate a separate class if it helped from an API perspective, but I don't think it does. Broadcast makes sense to me as an extension of the default "service pool" behaviour; they're both "event handlers"; if we made them separate implementations we'd need to give them distinct names, and I can't think of good ones.

If a single implementation with multiple modes is appropriate then I don't think it's "bad" in any sense to have an attribute, property or method that doesn't always apply. As you say, it's fine to have undefined behaviour when calling "out of turn".

I have converted `broadcast_handler` to be method rather than a property, for consideration, but I don't think it's really an improvement. It feels less idiomatic without benefit to me.

With regard to `getmembers` and descriptors... Is it really better to say "don't use descriptors" than it is to say "make sure descriptors are safe to be invoked at any time"? I think latter is more pythonic.

davidszotten: "don't use" vs "make sure is safe" is a fair point. i'm still kind of hoping we can handle it so users don't have to, but haven't done much research and at a glance it wasn't obvious if/how it could be done. 

"is safe" might be slightly harder to explain but i feel less strongly about it now i guess


re shared class with multiple modes i think that's a recipe for bad apis. we were already in that boat before this pr and i'm not suggesting we change that (as part of this at least). nor do i have any great suggestions for names (`@service_event` / `@broadcast_event` maybe?). _concepts_ and apis should drive object design, not code reuse. anyway. all this for another time

nameko/events.py
noisyboiler: agreed! :)

nameko/events.py
mattbennett: As far as I can see, the only control you get over `getmembers` is the `predicate` argument, which evaluates each item. That would invoke the descriptors if you're inspecting an instance of a class.

For the `bind` case specifically we can avoid the problem by inspecting the class rather than the instance, but we also call `getmembers` in `iter_extensions` and that isn't a simple change. I would really like to remove `iter_extensions` and daisy-chain calls to `setup`, `start`, `stop` etc. by making extensions responsible for looking after their subextensions, but that is a bigger issue for another time. I guess the summary is that's not an insurmountable problem.

In the mean time, "is safe" is nebulous but there is at least a concrete description: the descriptor on an extension will be called during container creation (after the extension's `__init__`, before `setup`) and may be called at any point afterwards.

As for healthy APIs... I completely agree that they should be driven by concepts not code reuse. I guess I like the `handler_type` part of the existing API - it makes it clear that the value changes how the handler behaves, as opposed to `broadcast_event`, which might suggest that the event type is different. That's a discussion for another time though.

What would you like me to do about this PR? I'm still fairly keen on using a descriptor with a caveat in the docs and an intention to improve it later.

davidszotten: as i said i'm less strongly against the descriptor now so if you much prefer it i'm happy to go back to it

CHANGES
```diff
@@ -6,15 +6,21 @@ are in form of *headline.major.minor* numbers. Backwards-compatible changes
...
 Released PENDING
 
+* Add support for loading configuration file in ``nameko shell`` via ``--
+  config`` option
+* Changed :class:`~nameko.web.handlers.HttpRequestHandler` to allow override
+  how web exceptions are handled
+* Enabled reliable delivery on broadcast events when combined with a custom
+  :py:attr:`~nameko.events.EventHandler.broadcast_identity`. Reliable delivery
+  now defaults to enabled for all handler types. It must be explicitly turned
+  off with broadcast mode unless override the default `broadcast_identity`.
```
davidszotten: unless _ override?

CHANGES
```diff
@@ -6,15 +6,21 @@ are in form of *headline.major.minor* numbers. Backwards-compatible changes
 increment the minor version number only.
 
 
-Version 2.2.1
+Version 2.3.0
 -------------
 
 Released PENDING
 
+* Add support for loading configuration file in ``nameko shell`` via ``--
```
davidszotten: awkward line break

nameko/events.py
davidszotten: do you just run it (on save?) or also use for validation?

davidszotten:
:+1: modulo spelling

CHANGES
mattbennett: unless _you_ override. I will fix, thanks.

CHANGES
mattbennett: Oh that is awkward. No idea how that got there!

PR #313, opened 2016-03-28
Configure logging via dictConfig and config.yaml

mattbennett:
Adds django-style configuration of the python logging framework via the config file


docs/cli.rst
```diff
@@ -24,17 +24,32 @@ and providing a simple YAML configuration file:
...
-
     WEB_SERVER_ADDRESS: '0.0.0.0:8000'
-
     rpc_exchange: 'nameko-rpc'
-
     max_workers: 10
-
     parent_calls_tracked: 10
 
+    LOGGING:
```
davidszotten: could we get away with a simpler example here? this increases what we describe as a "simple yaml config file" from 5 to 20 lines

mattbennett: I have trimmed it down some. It assumes you know about about how `dictConfig` works but I think that was probably the case beforehand anyway.

davidszotten:
👍 

changelog

PR #318, opened 2016-04-30
add call_async alias for MethodProxy.async

andriykohut:
https://github.com/onefinestay/nameko/issues/317


davidszotten:
hey, thanks for the report. i remember thinking about this when pep 492 was accepted. i wonder if we want to go a step further and start issuing deprecation warnings for the old name, since we'll probably have to drop it when `async` becomes a keyword

that doesn't have to be done as part of this, but wanted to raise the issue while i remembered

what i would like to see from this pr are
- consensus on the name (since `async` will eventually have to go). i'm fine with `call_async` i think. @mattbennett any thoughts?
- a test to codify the new name
- the docs updated  (only docs/examples/async_rpc.py i think)

thanks again!

andriykohut:
Oh, good point! I'm on it.

andriykohut:
Hmm, one of the jobs failed, https://travis-ci.org/onefinestay/nameko/jobs/127208999#L381 but this doesn't seem related, not sure how I can trigger that job again

mattbennett:
Seems like a fine name and approach to me. You'll need to add a test which covers the `.async` case (and ideally verifies the call to `warnings`) to get to 100% coverage.

davidszotten:
i think only maintainers can trigger re-build. have done so. iirc, the log disappears when builds are re-reiggered, so posted in a [gist](https://gist.github.com/davidszotten/1349148ac1dbdd43750430b1c51bab14) if anyone wants to investigate

nameko/rpc.py
```diff
@@ -350,10 +351,16 @@ def __call__(self, *args, **kwargs):
         reply = self._call(*args, **kwargs)
         return reply.result()
 
-    def async(self, *args, **kwargs):
+    def call_async(self, *args, **kwargs):
         reply = self._call(*args, **kwargs)
         return reply
 
+    def async(self, *args, **kwargs):  # pragma: no cover
```
davidszotten: can we not actually test this instead as matt [suggests](https://github.com/onefinestay/nameko/pull/318#issuecomment-216233850)?

andriykohut: @davidszotten Is there way to run tests locally? I can't find any explanation in the docs. It clearly expected that rabbitmq should be running on localhost, not sure how to pass it different amqp uri

davidszotten: sure. the tests use [`py.test`](http://pytest.org/latest/), are you familiar with that?

`py.test --help` shows our custom options, including `--amqp-uri=AMQP_URI   The AMQP-URI to connect to rabbit with.` (admittedly not that easy to find if you don't know what you're looking for)

you can also let `tox` handle it all for you (`pip install tox; tox`). actually, i think `tox` calls a makefile, so you can probably write `make test` (or look in the `Makefile` for other targets to run subtests)

to install test dependencies you can install `nameko[dev]`, e.g. `pip install -e .[dev]` 

will try to write some of this up and add to the docs

andriykohut: Hmm, I already tried that (`py.test --amqp-uri amqp://192.168.99.100:5672`), seems like it's hardcoded somewhere (or default value is used). I've ended up just `sed`-ing every `localhost` to my `docker-machine` ip, will try to figure out what's going on when I'll get a chance

mattbennett: You'll also need to use the `--rabbit-ctl-uri` flag, which specifies the management interface of the rabbit broker (you'll have to enable to management plugin too, if it's not already)

andriykohut:
I've added the test, mind taking a look?

nameko/rpc.py
davidszotten: can we drop the `no cover` now?

andriykohut: Done!

davidszotten:
cool. please also add a note in the changelog (backwards compat so can just bump the minor version) and add yourself to the bottom of contributors (if you want). see e.g. https://github.com/onefinestay/nameko/commit/41013e8ab4683f4cfeb6108f4f1b7c8897b72e6e for an example

andriykohut:
Looks like rebuild required for this job: https://travis-ci.org/onefinestay/nameko/jobs/127563677#L365

docs/spelling_wordlist.txt
```diff
@@ -136,3 +136,4 @@ webapp
 websocket
 websockets
 zato
+favor
```
davidszotten: afraid the docs use british english ("favour")

CHANGES
```diff
@@ -5,6 +5,12 @@ Here you can see the full list of changes between nameko versions. Versions
 are in form of *headline.major.minor* numbers. Backwards-compatible changes
 increment the minor version number only.
 
+Version 2.3.1
+-------------
+
+Released PENDING
+
+* Deprecate ``MethodProxy.async`` in favor of ``MethodProxy.call_async``
```
davidszotten: maybe add something like "in preparation for ``async`` becoming a keyword"

andriykohut:
🚀

davidszotten:
please see comment regarding spelling

docs/spelling_wordlist.txt
andriykohut: Ah, ok, I assumed both spelling will work.

davidszotten:
👍

davidszotten:
Thanks for much for the contribution!

andriykohut:
Nice! Thanks for helping me out with this

PR #324, opened 2016-05-17
remove "protected" managed threads

mattbennett:
Plus: simpler mechanism and better logging for worker vs managed threads


Makefile
```diff
@@ -1,5 +1,8 @@
 .PHONY: test docs
 
+COVERAGE_OPTS := --rcfile $(CURDIR)/.coveragerc
```
davidszotten: i thought we only needed to specify this for `pytest-cov` (it's already the default). do we definitely need this explicitly?

circle.yml
```diff
@@ -0,0 +1,19 @@
+machine:
```
davidszotten: how come we are adding another ci tool?

nameko/containers.py
```diff
@@ -171,8 +170,8 @@ def __init__(self, service_cls, config, worker_ctx_cls=None):
         self.started = False
         self._worker_pool = GreenPool(size=self.max_workers)
 
-        self._active_threads = {}
-        self._protected_threads = set()
+        self._worker_threads = WeakKeyDictionary()
```
davidszotten: how come these need to be weakkey?

nameko/containers.py
```diff
@@ -353,41 +348,55 @@ def spawn_worker(self, entrypoint, args, kwargs,
...
         _log.debug('spawning %s', worker_ctx)
-        gt = self._worker_pool.spawn(self._run_worker, worker_ctx,
-                                     handle_result)
-        self._active_threads[gt] = entrypoint
-        gt.link(self._handle_thread_exited)
+        gt = self._worker_pool.spawn(
+            self._run_worker, worker_ctx, handle_result
+        )
+        self._worker_threads[worker_ctx] = gt
+        self.worker_created(worker_ctx)
```
davidszotten: probably want to call `link` before the `worker_created` hook in case the latter yields

nameko/containers.py
```diff
@@ -353,41 +348,55 @@ def spawn_worker(self, entrypoint, args, kwargs,
...
-            self._active_threads[gt] = MANAGED_THREAD
-        else:
-            self._protected_threads.add(gt)
-        gt.link(self._handle_thread_exited)
+        if not isinstance(extension, Extension):
+            warnings.warn(
+                "The signature of `spawn_managed_thread` has changed. "
+                "The `protected` kwarg is now deprecated, and extensions "
+                "should pass themselves as the second argument for better "
+                "logging. See :meth:`nameko.containers.ServiceContainer."
```
davidszotten: passing `self` seems a bit ugly. also, if this is just for logging, do we want to be more explicit and call it out as that? an extension using multiple managed threads could then request them with different info, e.g. `MyExt.method_foo`

nameko/containers.py
```diff
@@ -353,41 +348,55 @@ def spawn_worker(self, entrypoint, args, kwargs,
...
+        gt = eventlet.spawn(fn)
+        self._managed_threads[gt] = extension
+        gt.link(self._handle_managed_thread_exited, extension)
         return gt
 
+    def worker_created(self, worker_ctx):
+        """ Called after a worker is created.
+
+        This is a hook for `ServiceContainer` subclasses to track worker
+        creation.
```
davidszotten: maybe a sentence about when/for what one should (and shouldn't!) use this

nameko/containers.py
```diff
@@ -421,6 +430,9 @@ def _run_worker(self, worker_ctx, handle_result):
                 _log.debug('handling result for %s', worker_ctx)
 
                 with _log_time('handled result for %s', worker_ctx):
+                    # TODO: retry handle_result once, catch any exception
```
davidszotten: this sounds like fairly surprising behaviour to me

nameko/containers.py
```diff
@@ -436,44 +448,43 @@ def _run_worker(self, worker_ctx, handle_result):
...
+        num_workers = len(self._worker_threads)
 
-        if num_active_threads:
-            _log.warning('killing %s active thread(s)', num_active_threads)
-            for gt, extension in list(self._active_threads.items()):
-                if extension is not MANAGED_THREAD:
-                    _log.warning('killing active thread for %s', extension)
+        if num_workers:
+            _log.warning('killing %s active workers(s)', num_workers)
+            for worker_ctx, gt in self._worker_threads.copy().items():
```
davidszotten: i think the canonical way to iterate over dict while mutating is `for key, val in list(dict.items()):`

test/test_container.py
```diff
@@ -351,7 +331,7 @@ def raise_error():
         raise Exception('foobar')
 
     container.start()
-    container.spawn_managed_thread(raise_error)
+    container.spawn_managed_thread(raise_error, Extension())
```
davidszotten: isn't the second argument optional?

Makefile
mattbennett: We also need it for the `coverage merge` command to work (which is what `pytest-cov` does internally).

circle.yml
mattbennett: Circle has an amazing SSH feature that lets you connect to a box after a failed run. I added this so that I could live-debug what turned out to the the eventlet-pytest-py3 compatibility issue (on my local box it just hangs).

Circle doesn't parallelise builds using tox though, so the test run is slower. I'm happy to remove this again, but I will keep it in a gist for the next time!

nameko/containers.py
mattbennett: The `WorkerContext` has a reference to the `ServiceContainer`, and these dicts are on the container. I _think_ we're disciplined enough about the worker lifecycle to make these non-weak references, but I'm not sure.

I'm don't know what best practice is here -- should we make them weakrefs just in case, or is that sloppy?

nameko/containers.py
mattbennett: Happy to do this, but why does that matter?

nameko/containers.py
mattbennett: Great suggestions

nameko/containers.py
mattbennett: Will do

nameko/containers.py
davidszotten: in case `gt` gets scheduled and completes/dies before you get to call `link`

nameko/containers.py
mattbennett: Hmm, yes. I think we can scratch this.

I wanted to apply some kind of retry policy for replies, so that (for example) an intermittent connection error would be overcome. That is clearly the responsibility of the entrypoint though.

Makefile
davidszotten: again, since `.coveragerc` is the default file location i think coverage looks for it even if we don't specify. seems to work for me locally (even via tox). Does it fail for you if you drop the explicit `COVERAGE_OPTS`?

circle.yml
davidszotten: Yea, ssh'ing into ci is nice, and afik travis doesn't offer it. travis is still fairly defacto standard so would prefer to stick to (just) that for now

was this just more convenient or were you not able to reproduce the issue (e.g. in a local vm)?

nameko/containers.py
mattbennett: Is it? In Py3 `dict.items()` will give you back a `dict_items` object, so if you `copy()` you're also iterating over a `dict_items` (which is a generator, I think) rather than reading them all into a list first.

test/test_container.py
mattbennett: Optional but raises a `DeprecationWarning`

nameko/containers.py
mattbennett: Ah yes, of course. Thanks!

nameko/containers.py
davidszotten: except you've just called `copy` which copies the data anyway.

(`dict.items()` returns a _view_, which is linked to the original dict so that it updates dynamically)

circle.yml
mattbennett: Just more convenient that a local VM. I'll remove the circle.yml file

nameko/containers.py
davidszotten: given that they weren't before, and we use `link` pretty well, i think i'd prefer to avoid weakrefs

davidszotten: on a separate note, from a quick look it seems we only put the container in the `worker_ctx` to get to `config` and `service_name`. might be better just putting those in. something for another time

test/test_container.py
davidszotten: might change as discussed above, but if we're keeping it optional, could we not skip the warning if the value is `None`?

Makefile
mattbennett: Yep, works fine. I will remove this

circle.yml
mattbennett: For posterity: https://gist.github.com/mattbennett/aa7cf713c97e689c56257991231471e5

nameko/containers.py
mattbennett: OK, will remove. I completely agree on removing `container` from the `WorkerContext`. Bound extensions all have access to it anyway, we don't need to pass the reference around everywhere.

nameko/containers.py
mattbennett: I guess `copy` is just as bad as `list`. Happily, removing the WeakKeyDictionaries makes this problem go away.

davidszotten: i don't understand that. are we not still mutating the dictionary inside the loop?

test/test_container.py
mattbennett: Yep, this has solved itself :)

nameko/containers.py
mattbennett: Struggling to come up with something that's helpful here. It's a hook for you to do what you want with... I don't think there's anything specific you really _shouldn't_ do

nameko/containers.py
mattbennett: Hmm, maybe this works by accident. We kill the greenthreads here, but `_handle_x_thread_exited` isn't called until we yield. You would think that the log call would/could yield though.

I will add the `list` back for safety.

test/test_container.py
```diff
@@ -118,6 +120,11 @@ def logger():
         yield patched
 
 
+@pytest.fixture(autouse=True)
```
davidszotten: i recently realised that these are silent by default, and wondering whether we want to change that. until recently django used to use a subclass to un-hiden deprecation warnings https://github.com/django/django/blob/1.9/django/utils/log.py#L55 (was changed in https://code.djangoproject.com/ticket/25999 to aid their LTS process which involves third parties conditionally importing old stuff which would trigger these warnings). i don't think we have that problem, so we might want to consider doing this

mattbennett: The "default" behaviour which django enables there also only shows the warning once, so subsequent trips through the same code path don't trigger it. To fix the currently failing tests I need to set it to "always" mode.

I have added it as a top-level fixture (rather than in `pytest_configure`) so that it can be disabled by overriding if needs be.

davidszotten: my comment wasn't so much about this test, as about our desired behaviour in general (e.g. when using `run`. we recently started adding deprecation warnings but users might not be seeing them

mattbennett:
@davidszotten this is ready for another round of review.

Apologies for all the commits. I have been back and forth *a lot* with various versions of pytest, eventlet, coverage and pytest-cov.

The root cause of all the failures above was upgrading to a recent pytest, which has issues with eventlet on python3. I've summarised these in a separate repo (https://github.com/mattbennett/eventlet-pytest-bug) but the short answer is that pytest > 2.7.3 is incompatible with eventlet > 0.17.4.

Along the way I also learned that (irrespective of python version) we can't use coverage on its own (i.e. without pytest-cov) because coverage from tests executed by the `testdir` subprocess aren’t stored unless they return with `0` status code, and nor can we use pytest-cov > 2.1.0 because it doesn't honour the `coveragerc` file when used with the `testdir` pytest plugin. For posterity, I created and then reverted a couple of commits to demonstrate these failures.

There is at least one bug to raise against eventlet/pytest. It may also be considered a bug that coverage isn’t collected in all cases when using `testdir`, and that the behaviour in pytest-cov changed since 2.1.0. I’ll try to raise these too.

In the mean time I don’t think there’s anything blocking this PR from landing, if you’re happy.

nameko/containers.py
```diff
@@ -353,41 +347,58 @@ def spawn_worker(self, entrypoint, args, kwargs,
...
         return gt
 
+    def worker_created(self, worker_ctx):
+        """ Called after a worker is created.
+
+        This is a hook for `ServiceContainer` subclasses to track worker
+        creation.
+        """
+
+    def worker_destroyed(self, worker_ctx):
```
davidszotten: was thinking a bit more about these. if the main use-case you have in mind is a better hook for something like the entrypoint waiter, we might want to keep this private for now, (and only add the teardown one).

mattbennett: That is the main use-case and I'm happy to make that change. I'll wait 'til you see my proposed implementation for the entrypoint waiter though, since there might be other things to discuss.

mattbennett: I have backed them out because `_worker_teardown` is as good as `worker_destroyed` for my (only) use-case

test/test_service_runner.py
```diff
@@ -274,7 +275,7 @@ def raises():
     runner = runner_factory(rabbit_config, Service)
 
     container = get_container(runner, Service)
-    container.spawn_managed_thread(raises)
+    container.spawn_managed_thread(raises, Extension())
```
davidszotten: no longer needer, right?

nameko/messaging.py
```diff
@@ -195,8 +196,7 @@ def start(self):
             self._starting = True
 
             _log.debug('starting %s', self)
-            self._gt = self.container.spawn_managed_thread(self.run,
-                                                           protected=True)
+            self._gt = self.container.spawn_managed_thread(self.run, self)
```
davidszotten: should stop passing `self`. pass string or nothing i guess

nameko/timer.py
```diff
@@ -36,7 +37,7 @@ def tick(self):
 
     def start(self):
         _log.debug('starting %s', self)
-        self.gt = self.container.spawn_managed_thread(self._run)
+        self.gt = self.container.spawn_managed_thread(self._run, self)
```
davidszotten: passing `self`, as above

nameko/web/server.py
```diff
@@ -86,8 +84,7 @@ def start(self):
                                      self._wsgi_app,
                                      protocol=HttpOnlyProtocol,
                                      debug=False)
-            self._gt = self.container.spawn_managed_thread(
-                self.run, protected=True)
+            self._gt = self.container.spawn_managed_thread(self.run, self)
```
davidszotten: `self`

test/test_container.py
```diff
@@ -560,3 +550,43 @@ class AnonymousService():
...
+):
+
+    def wait():
+        Event().wait()
+
+    container.spawn_managed_thread(wait, *args, **kwargs)
+
+    # TODO: pytest.warns is not supported until pytest >= 2.8.0, whose
+    # `testdir` plugin is not compatible with eventlet on python3 --
+    # see https://github.com/mattbennett/eventlet-pytest-bug
```
davidszotten: fyi i think the fix is https://github.com/eventlet/eventlet/pull/257 . trying to push that along. (my journey's in https://github.com/eventlet/eventlet/pull/314)

test/test_service_runner.py
mattbennett: Correct. I will fix all of these

test/test_container.py
mattbennett: Amazing. I'm enjoying watching those tickets progress

mattbennett:
@davidszotten I think this can land now, if you're happy

davidszotten:
yep. this is fine. (changelog entry?) 👍

nameko/containers.py
```diff
@@ -385,20 +384,6 @@ def spawn_managed_thread(self, fn, protected=None, identifier=None):
...
         return gt
 
-    def worker_created(self, worker_ctx):
-        """ Called after a worker is created.
-
-        This is a hook for `ServiceContainer` subclasses to track worker
-        creation.
-        """
-
-    def worker_destroyed(self, worker_ctx):
```
davidszotten: hm. i thought we were keeping this one (for use in the entrypoint waiter?

mattbennett: it turns out that `worker_teardown` is just as effective (the patched method is run before the hook)

PR #325, opened 2016-05-20
Better entrypoint waiter

mattbennett:
Includes #324, which provides the `worker_destroyed` hook


nameko/testing/services.py
```diff
@@ -76,33 +75,12 @@ def catch_container_errors(gt):
...
-
-    def wait(self):
-        self.done.acquire()
+class EntrypointWaiterTimeout(Exception):
+    pass
 
 
 @contextmanager
-def entrypoint_waiter(container, entrypoint, timeout=30):
+def entrypoint_waiter(container, method_name, timeout=30, callback=None):
```
davidszotten: i'm not a huge fan of this `callback` api. do you have other use-cases in mind than multiple calls?

mattbennett: Sorry, I should have put a note on this PR. It's a work on progress.

I think I'm going to scrap it for an `even-better-entrypoint-waiter` PR...

PR #327, opened 2016-05-24
Add dict-like access to shell RPC proxy

gianchub:
Hello,

this is to add the ability of calling services whose name is not a valida Python name from the RPC proxy when using the nameko shell helper.

CC:
@mattbennett @davidszotten


nameko/standalone/rpc.py
```diff
@@ -308,6 +317,12 @@ def __getattr__(self, name):
                 self._worker_ctx, name, self._reply_listener)
         return self._proxies[name]
 
+    def __getitem__(self, name):
+        """Enable dict-like access on the proxy. """
+        if not isinstance(name, string_types):
```
davidszotten: nice! if you use `getattr(self, name)` instead of calling `self.__getattr__` you get the type check for free

test/standalone/test_rpc_proxy.py
```diff
@@ -362,6 +362,37 @@ def test_cluster_proxy(container_factory, rabbit_manager, rabbit_config):
...
+    container_factory, rabbit_manager, rabbit_config
+):
+    container = container_factory(FooService, rabbit_config)
+    container.start()
+
+    with ClusterRpcProxy(rabbit_config) as proxy:
+        assert proxy['foobar'].spam(ham=3) == 3
+
+
+@pytest.mark.parametrize('name', [
```
davidszotten: if we switch to `getattr` above we can probably drop this test entirely

CHANGES
```diff
@@ -5,6 +5,15 @@ Here you can see the full list of changes between nameko versions. Versions
 are in form of *headline.major.minor* numbers. Backwards-compatible changes
 increment the minor version number only.
 
+Version 2.3.2
+-------------
+
+Released PENDING
+
+* Add dictionary access to ``standalone.rpc.ClusterProxy`` to allow the proxy
+  to call services whose name is not legal in Python (e.g. name has a
```
davidszotten: maybe "is not a legal _identifier_ in python"

davidszotten:
thanks for the pr! a few small comments

also, feel free to add yourself to `CONTRIBUTORS.txt` as part of this pr

nameko/standalone/rpc.py
gianchub: Ah yes, that's much better!

test/standalone/test_rpc_proxy.py
gianchub: Removed.

gianchub:
Thank you so much David,
all comments addressed, ready for another pass.

fab

davidszotten:
👍

PR #334, opened 2016-06-14
add retry to rabbit connection check

mattbennett:
Not entirely sure how I feel about introducing https://pypi.python.org/pypi/retry as a dependency, but this is a helpful improvement to stop intermittent test failures


mattbennett:
@davidszotten what do you think about introducing this library as a dependency?

It's only conditionally required if users also have pytest installed. I'm not sure if setuptools has the ability  to specify rules like that.

davidszotten:
for the pretty small part of that already fairly small library, you could probably just reimplement

mattbennett:
@davidszotten I have reimplemented the most useful bits of that library

davidszotten:
i'm sorry if i wasn't very clear but my point definitely wasn't that we should provide our own general  `@retry` utility that we have to maintain. if a user of nameko needs that, _they_ can just use `retry` from pypi. what did i mean was, since you are only using `@retry` with all the defaults (and in particular, no backoff, max_delay or allowed_exceptions), that we'd just implement a very simple decorator that takes no arguments and provides just the behaviour we need. given that we're decorating an internal function use by our pytest plugin, we probably don't even need `wrapt` (though i mind less depending on it)

mattbennett:
No, I completely agree. I've removed it for a use-case specific version.

davidszotten:
👍

PR #336, opened 2016-06-17
Catch IOError in PollingQueueConsumer

aheaume:
Hi folks,

We're getting `IOError: Socket closed` while waiting for synchronous RPC calls to complete. The disconnections are expected (HAProxy timing out after 60s) but the error type isn't (`IOError` instead of `socket.error`)

Here's a typical stacktrace

```
 File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
   R = retval = fun(*args, **kwargs)
 File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
   return self.run(*args, **kwargs)
 File "/data/backoffice/forcity_core/api_simulation/scenario/tasks.py", line 69, in compile_timeline_via_broke                                              r
   ret.reply_event.queue_consumer._setup_consumer()
 File "/usr/local/lib/python2.7/dist-packages/nameko/standalone/rpc.py", line 80, in _setup_consumer
   self.consumer.cancel()
 File "/usr/local/lib/python2.7/dist-packages/kombu/messaging.py", line 461, in cancel
   cancel(tag)
 File "/usr/local/lib/python2.7/dist-packages/amqp/channel.py", line 1633, in basic_cancel
   (60, 31),  # Channel.basic_cancel_ok
 File "/usr/local/lib/python2.7/dist-packages/amqp/abstract_channel.py", line 67, in wait
   self.channel_id, allowed_methods, timeout)
 File "/usr/local/lib/python2.7/dist-packages/amqp/connection.py", line 241, in _wait_method
   channel, method_sig, args, content = read_timeout(timeout)
 File "/usr/local/lib/python2.7/dist-packages/amqp/connection.py", line 330, in read_timeout
   return self.method_reader.read_method()
 File "/usr/local/lib/python2.7/dist-packages/amqp/method_framing.py", line 189, in read_method
   raise m
IOError: Socket closed
```

Catching `IOError` in `PollingQueueConsumer._setup_consumer` is enough to fix the problem.

Do you think this is worth merging? Or should we find another solution?

Thanks,
Alex

/cc @manfred-chebli


davidszotten:
since we are just trying to reset the consumer, ignoring ioerrors as well seems fine. i'd like to understand when we get those instead of the`socket.error` that we are already catching. if this turns out to be what we want, we can probably catch both together `except (Foo, Bar)` and amend the comment

aheaume:
[`transport.py` in py-ampqlib](https://github.com/barryp/py-amqplib/blob/master/amqplib/client_0_8/transport.py) will raise `IOError` when it can't read from or write to the socket. I guess this is cannot be considered a "socket error" since the socket never actually failed, it's just closed.

In fact, according to [the python 2 socket docs](https://docs.python.org/2/library/socket.html#socket.error) `socket.error` has been a child of `IOError` since 2.6. In [the python 3 socket docs](https://docs.python.org/3.6/library/socket.html#socket.error) it is a subclass of `OSError` since 3.3 (of which `IOError` is an alias, I believe).

davidszotten:
cool, thanks for the investigation and the patch! 

(looks like you have a pyflakes error causing the tests to fail)

aheaume:
For some reason I couldn't get the Travis log until now. I'll fix the line length and squash my commits.

davidszotten:
cool! if you fancy, you can amend this pr to add yourself to the list of contributors in the project root
👍

aheaume:
Done, thanks @davidszotten

jessepollak:
We are running into this issue a fair amount both in the services we run and with clef/flask-nameko - is there an ETA for a next release that will include this fix?

PR #337, opened 2016-06-22
Enable confirms for all amqp publishers

mattbennett:
Enable confirms for all AMQP publishers so we can honour our guarantee of "at least once" delivery, even in unreliable networks.
- Refactors the various users of kombu's pools into nameko.amqp
- Creates a nicer pattern for mocking these helpers
- Defaults confirms to true, but allows subclasses to disable it

This PR also adds a number of functional tests using https://github.com/Shopify/toxiproxy, which demonstrate how the AMQP components behave under various connection errors, usually:
- When the socket is closed
- When the socket times out
- When the connection is recovered
- When the connection is recovered and a retry policy is in place

Publish confirms are required for publishers to detect a connection that is broken after being established.


nameko/amqp.py
```diff
@@ -47,3 +51,33 @@ def verify_amqp_uri(amqp_uri):
...
+    conn = Connection(amqp_uri, transport_options=transport_options)
+
+    with producers[conn].acquire(block=True) as producer:
+        yield producer
+        try:
+            returned = producer.channel.returned_messages.get_nowait()
+        except Empty:
+            pass
+        else:
+            raise UndeliverableMessage(returned)
```
davidszotten: i don't think the event dispatcher wants to raise here. there might not be anyone subscribed to the events yet. we use the confirm to get the server to "acknowledge" our message, but returned messages are also fine in this case

mattbennett: It only raises when the `mandatory` flag is set on the message. It would be unusual to set mandatory delivery for an event dispatcher, but less so for a normal publisher. And reasonable to raise in any case if it is set.

test/standalone/test_rpc_proxy.py
```diff
@@ -430,3 +431,124 @@ def __setitem__(self, key, value):
     for _, message in fake_replies.messages:
         consumer_tags.add(message.delivery_info['consumer_tag'])
     assert len(consumer_tags) == 1
+
+
+@pytest.mark.skipif(
+    spawn.find_executable('toxiproxy-server') is None,
```
mattbennett: Can we mark this as "do not skip" on Travis only?

mattbennett:
I'm surprised `verify_amqp_uri` doesn't give a better error message in this case though

mattbennett:
@davidszotten @iky @timbu 

This is ready for review now. The toxiproxy tests were useful for me to understand how the different failure modes were handled. I think they're useful to keep as functional demonstrations, but I could be persuaded otherwise.

mattbennett:
Bump on this please chaps.

test/standalone/test_rpc_proxy.py
```diff
@@ -410,3 +413,121 @@ def __setitem__(self, key, value):
...
+            # expect the write to raise a BrokenPipe or, if it succeeds,
+            # the socket to be closed on the subsequent confirmation read
+            "Broken pipe" in str(exc_info.value) or
+            "Socket closed" in str(exc_info.value)
+        )
+
+        toxiproxy.enable()
+
+        # consumer discovers its socket was closed
+        # TODO: can we not reestablish the connection earlier?
```
davidszotten: i think this is the next chance the proxy has to run code. also, note that you caught the `IOError`, which we wouldn't expect a regular caller to do, right?

test/test_amqp.py
```diff
@@ -49,3 +54,147 @@ def test_other_error(uris):
...
+        with get_producer(amqp_uri, confirms) as producer:
+            producer_ids.append(id(producer))
+            transport_options = producer.connection.transport_options
+            assert isinstance(producer, Producer)
+            assert transport_options['confirm_publish'] is confirms
+
+        with get_producer(amqp_uri, confirms) as producer:
+            producer_ids.append(id(producer))
+            assert len(set(producer_ids)) == 1
+
```
davidszotten: maybe check that with different values for `confirm` we get different producers

test/test_messaging.py
```diff
@@ -411,3 +381,264 @@ def spawn_managed_thread(method):
...
+    def use_confirms(self, request):
+        with patch.object(Publisher, 'use_confirms', new=request.param):
+            yield request.param
+
+    @pytest.yield_fixture
+    def publisher_container(
+        self, request, container_factory, tracker, rabbit_config
+    ):
+        retry = False
+        if "enable_retry" in request.keywords:
```
davidszotten: don't think this would work if we run pytest with `--strict`, which i thought we were (we aren't) and think we should

test/test_rpc.py
```diff
@@ -648,3 +654,364 @@ def never_stops():
...
+        def kill_greenthreads():
+            for gt in gts:
+                try:
+                    gt.kill()
+                except GreenletExit:  # pragma: no cover
+                    pass
+            del gts[:]
+
+        class ThreadSafeRpcProxy(object):
+            """ The ServiceRpcProxy is not thread-safe, so we can't create it
```
davidszotten: how come? do we need to fix this?

nameko/messaging.py
```diff
@@ -105,30 +101,36 @@ def spam(self, data):
...
-        conn = Connection(self.amqp_uri)
-        return producers[conn].acquire(block=True)
-
-    def setup(self):
+    @property
+    def amqp_uri(self):
+        return self.container.config[AMQP_URI_CONFIG_KEY]
 
-        self.amqp_uri = self.container.config[AMQP_URI_CONFIG_KEY]
+    @property
```
mattbennett: `@property` makes them read-only on the instance, so you can only override them with a subclass. I _think_ that's a helpful restriction, but I could be persuaded otherwise.

nameko/messaging.py
```diff
@@ -105,30 +101,36 @@ def spam(self, data):
...
+    def use_confirms(self):
+        return True
```
mattbennett: Yep, I will add these

nameko/rpc.py
```diff
@@ -351,6 +367,33 @@ def __call__(self, *args, **kwargs):
         reply = self._call(*args, **kwargs)
         return reply.result()
 
+    @property
```
mattbennett: Yes. It seems this would be helpful (https://github.com/onefinestay/nameko/issues/374)

I'll make this enhancement as part of this PR

.travis.yml
```diff
@@ -1,11 +1,15 @@
...
 language: python
 python:
   - "2.7"
 
 services:
     - rabbitmq
 
+before_install:
+    - wget -O toxiproxy-2.0.0.deb https://github.com/Shopify/toxiproxy/releases/download/v2.0.0/toxiproxy_2.0.0_amd64.deb
+    - sudo dpkg -i toxiproxy-2.0.0.deb
```
davidszotten: i didn't mean _using_ the library (i saw you commented about it not being ready). i meant using their approach to travis setup

mattbennett: Ahh OK! I'll see what I can do...

nameko/rpc.py
```diff
@@ -387,19 +422,22 @@ def _call(self, *args, **kwargs):
...
+        # functionality and therefore :class:`UnknownService` will never
+        # be raised (and the caller will hang).
+
         routing_key = '{}.{}'.format(self.service_name, self.method_name)
 
         exchange = get_rpc_exchange(container.config)
 
-        with producers[conn].acquire(block=True) as producer:
-
+        def publish(producer):
```
mattbennett: Actually not sure. I've unrolled it so it's more inline with the other producers

test/conftest.py
```diff
@@ -1,10 +1,151 @@
...
+
+@pytest.yield_fixture
+def mock_connection():
+    with patch('nameko.amqp.connections') as patched:
+        yield patched[ANY].acquire().__enter__()
+
+
+@pytest.fixture(scope='session')
+def toxiproxy_host():
+    return "127.0.0.1"
```
mattbennett: Good points. I've turned them into constants in the module.

test/standalone/test_rpc_proxy.py
```diff
@@ -410,3 +413,121 @@ def __setitem__(self, key, value):
...
+
+    def test_timeout(self, service_rpc, toxiproxy):
+        toxiproxy.timeout()
+
+        with pytest.raises(IOError) as exc_info:
+            service_rpc.echo(1)
+        assert "Socket closed" in str(exc_info.value)
+
+    def test_recover_from_down(self, service_rpc, toxiproxy, use_confirms):
+        if not use_confirms:
```
mattbennett: Yeah I'm not convinced this is good.

I was trying to demonstrate that these cases aren't recoverable unless you have publish confirms enabled. In contrast the other cases behave the same way regardless of confirms.

I'll change it so we don't try to run them when without confirms enabled, and put the context into a docstring.

test/test_messaging.py
mattbennett: Strict mode actually works fine if you register `enable_retry` as a mark in an ini file.

Is there a better way to control the behaviour of a fixture like this? Also do you want me to enable strict mode as part of this PR?

test/test_amqp.py
mattbennett: This passes:

``` python
    def test_different_producers(self, rabbit_config):
        amqp_uri = rabbit_config['AMQP_URI']
        producer_ids = []

        # get a producer
        with get_producer(amqp_uri, True) as confirmed_producer:
            producer_ids.append(id(confirmed_producer))
            assert len(set(producer_ids)) == 1

        # get a producer with the same parameters
        with get_producer(amqp_uri, True) as confirmed_producer:
            producer_ids.append(id(confirmed_producer))
            assert len(set(producer_ids)) == 1

        # get a producer with different parameters
        with get_producer(amqp_uri, False) as unconfirmed_producer:
            producer_ids.append(id(unconfirmed_producer))
            assert len(set(producer_ids)) == 2
```

But I'm not sure it's a valid test. Publish confirms are a parameter of the connection -- isn't it internal to kombu whether you get the same or different producer instances back in this case?

test/test_rpc.py
mattbennett: Eventlet throws `Second simultaneous read` errors. I don't think we need to fix it.

The `ServiceRpcProxy` is synchronous and blocking so it won't interfere with whatever concurrency framework a client app uses. It assumes that it's being used either in synchronous code or safe concurrent code, whereas here we're sharing a single instance between multiple threads.

nameko/messaging.py
davidszotten: don't feel that strongly. python tends to take the "we're all consenting adults" approach as opposed to making things hard to change. not sure if this is something to worry about in practice. i guess regular attrs is a bit less code, but don't mind

test/conftest.py
```diff
@@ -1,10 +1,151 @@
...
+            'upstream': upstream
+        })
+    )
+
+    # create proxied uri for publisher
+    proxy_uri = amqp_uri.replace(str(rabbit_port), str(proxy_port))
+
+    toxic_name = '{}_timeout'.format(proxy_name)
+
+    class Controller(object):
```
davidszotten: ok

test/test_amqp.py
davidszotten: my understanding was that they have to be different since they have different connection params (fixed a bug in kombu a few years ago when it wasn't doing this properly causing issues in nameko)

test/test_messaging.py
davidszotten: unless it's faff, why not

test/test_amqp.py
mattbennett: Ah yes that rings bells. I'll add this

test/test_messaging.py
mattbennett: Cool. Will enable strict mode.

Makefile
```diff
@@ -11,10 +11,10 @@ pylint:
 	pylint --rcfile=pylintrc nameko -E
 
 test_lib:
-	BRANCH=$(ENABLE_BRANCH_COVERAGE) py.test test --cov --cov-config=$(CURDIR)/.coveragerc
+	BRANCH=$(ENABLE_BRANCH_COVERAGE) py.test test --strict --cov --cov-config=$(CURDIR)/.coveragerc
```
davidszotten: one benefit is that i get a good error message when i misspell parameterize

mattbennett: :) I nearly raised a pull request to enable both US and UK spellings...

nameko/standalone/rpc.py
```diff
@@ -48,10 +48,17 @@ def wait(self):
         if self.exception:
             raise self.exception
 
-        if self.queue_consumer.consumer.connection is None:
+        if self.queue_consumer.stopped:
             raise RuntimeError(
                 "This consumer has been stopped, and can no longer be used"
             )
+        # TODO: seems unnecessarily unhelpful -- can't we try to reconnect?
```
mattbennett: It seems not. You have to re-establish the consumer (and its exclusive reply queue) before sending any request, otherwise the reply queue may not exist when the response is published. The `MethodProxy` is used by both the standalone and dependency versions of the RPC proxy, so you can't use check the consumer connection before sending a request.

.travis.yml
```diff
@@ -6,6 +6,11 @@ python:
 services:
     - rabbitmq
 
+before_install:
+    - wget -O toxiproxy-server https://github.com/Shopify/toxiproxy/releases/download/v2.0.0/toxiproxy-server-linux-amd64
+    - chmod +x toxiproxy-server
+    - export PATH=$PATH:$PWD/
```
davidszotten: this seems a bit dangerous (what other binaries might end up in the working dir?). probably fine for travis, but better practice might be to put this binary into a subdirectory and just add that

nameko/amqp.py
```diff
@@ -47,3 +51,33 @@ def verify_amqp_uri(amqp_uri):
     transport = TestTransport(connection.transport.client)
     with transport.establish_connection():
         pass
+
+
+class UndeliverableMessage(Exception):
```
davidszotten: thoughts?

.travis.yml
mattbennett: OK, I'll make a subdirectory here

mattbennett:
I'd like to bump the `connection_params` changes into a separate PR because it deserves its own consideration. Given that, this is ready for a final review.

mattbennett:
Although it will (gently) clash with #383 so we may want to land one before the other.

nameko/amqp.py
```diff
@@ -47,3 +51,33 @@ def verify_amqp_uri(amqp_uri):
...
+
+@contextmanager
+def get_producer(amqp_uri, confirms=True):
+    transport_options = {
+        'confirm_publish': confirms
+    }
+    conn = Connection(amqp_uri, transport_options=transport_options)
+
+    with producers[conn].acquire(block=True) as producer:
+        yield producer
```
davidszotten: just wanted to check we are happy with the behaviour if multiple messages are sent while inside this cm

nameko/standalone/rpc.py
davidszotten: replace the (code) comment with this?

test/conftest.py
```diff
@@ -1,9 +1,150 @@
...
+        ['toxiproxy-server', '-port', str(port), '-host', host],
+        stdout=subprocess.PIPE
+    )
+    time.sleep(0.2)  # allow server to start
+    yield "{}:{}".format(host, port)
+    server.terminate()
+
+
+@pytest.yield_fixture
+def toxiproxy(toxiproxy_server, rabbit_config, free_port):
```
davidszotten: hm. every user of the `free_port` fixture will be given the same port. which probably isn't what we want

mattbennett: Quite right. It needs to be a factory fixture (or just a util function)

nameko/amqp.py
mattbennett: Hmm. As it's implemented here you can send multiple undeliverable messages and be notified about only one of them, and then only after you exit the contextmanager.

It might be better to push this check down in to the different implementations.

mattbennett: I've done this and added explicit tests about how to use the mandatory flag. It's nicer.

mattbennett:
@davidszotten, this is ready for another round.

nameko/standalone/events.py
```diff
@@ -48,5 +51,23 @@ def dispatch(service_name, event_type, event_data):
                     routing_key=routing_key,
                     retry=retry,
                     retry_policy=retry_policy,
+                    mandatory=mandatory,
                     **kwargs)
+
+                if mandatory:
+                    if not use_confirms:
+                        warnings.warn(
```
davidszotten: a warning is fine for now, but i think it would be better to choose either error or nothing

PR #340, opened 2016-07-06
add master eventlet build

mattbennett:
As suggested in https://github.com/eventlet/eventlet/pull/314

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ env:
   - TOX_ENV=docs
 
 matrix:
+  fast_finish: true
   allow_failures:
     - env: TOX_ENV=py27-mastereventlet-lib
     - env: TOX_ENV=py33-mastereventlet-lib
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,15 +21,24 @@ env:
   - TOX_ENV=py27-oldest-lib
   - TOX_ENV=py27-pinned-lib
   - TOX_ENV=py27-latest-lib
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -28,6 +28,8 @@ deps =
 whitelist_externals = make
 
 commands =
+    mastereventlet: pip install --upgrade https://github.com/eventlet/eventlet/archive/master.zip
```
davidszotten: how come you are using the zip rather than git?

.travis.yml
```diff
@@ -21,16 +21,25 @@ env:
...
   - TOX_ENV=py34-oldest-lib
   - TOX_ENV=py34-pinned-lib
   - TOX_ENV=py34-latest-lib
+  - TOX_ENV=py34-mastereventlet-lib
   - TOX_ENV=py27-examples
   - TOX_ENV=py34-examples
   - TOX_ENV=docs
 
+matrix:
+  allow_failures:
```
davidszotten: might also want `fast_finish: true` here

tox.ini
mattbennett: Apparently it's much faster, presumably because it doesn't have to clone the whole repo. See http://stackoverflow.com/a/24811490/128749

.travis.yml
mattbennett: Nice!

tox.ini
davidszotten: ah, i see

PR #341, opened 2016-07-12
Even better entrypoint waiter

mattbennett:
Alternative entrypoint waiter implementation that:

* Gives you access to the result returned (or exception raised)
* Adds the ability to wait for a specific result
* Doesn't fire until the worker is torn down (without using wait_for_worker_idle so fixes onefinestay#165)

Originally raised and reviewed on https://github.com/mattbennett/nameko/pull/2


mattbennett:
@davidszotten --

@timbu and @iky and I have been discussing this PR, in particular your concerns about the similarity of the `wait_for_call` and `entrypoint_waiter` callback signatures.

They do similar things (both "wait for a method call") but the `entrypoint_waiter` is tailored to a special case and has to check much more than just that the service method was executed -- it waits for the method to be executed, the result to be handled and worker teardown to complete. We couldn't find a way to one a special case of the other.

But there is a way to make their callback signatures the same...

The argument against having the worker context in the `entrypoint_waiter` callback is that most test writers will not need it. Having said that is is useful for testing nameko internals, as I have done in https://github.com/mattbennett/nameko/pull/3/files#diff-c6900903a76fe66c078314cc5973ee04R655

So we could have a basic  documented `callback` argument with the same signature as the `wait_for_call` callback, and an another optional `worker_callback` (or similar) that gets the worker context instead of args and kwargs.

I prefer having two separate signatures, but I will push something like the above for your consideration.

At the risk of adding another chef, do you have any thoughts on this @jessepollak? Is it confusing to have have both of these APIs?

Would it be less confusing if we spun `wait_for_call` out into a standalone library, so the `entrypoint_waiter` was more obviously implemented on top of it?

jessepollak:
@mattbennett I don't think I have enough context on this PR to have a strong opinion! sorry!

davidszotten:
i agree that the former is probably better. clear apis with messy or confusing implementations are better than the other way around

mattbennett:
OK great, I've reverted those changes.

I think this is ready to land, if you're happy.

davidszotten:
👍

PR #342, opened 2016-07-15
customise_dependencies - More control over "replaced" dependencies when testing.

timbu:
During testing, we don't always want dependencies replaced with Mock(). 
E.g. 
-  When dependency providers provide primitive values
-  When you want to use a "fake" dependency with more controlled behaviour.

`replace_dependencies` now lets you supply the dependencies you want as kwargs.


davidszotten:
nice idea! api-wise, did you consider just expanding the existing `replace_dependencies` to accept kwargs as well as args?

timbu:
thanks @davidszotten :) - I wasn't quite sure which was the best way to go... I can definitely see an advantage in having only one method (as both are essentially doing the same thing). But the downside is we will make the docstring quite complex by accepting both args and kwargs  - e.g. the return values are needed in the `args` case, but are irrelevant (unusable really, as they are an ordered list based on unordered kwargs) in the `kwargs` case .
Am happy to change if you think kwargs is the way to go.

davidszotten:
hm. that's a good point. though i guess we could keep the same behaviour with respect to `args` and return values (so only returning the replacements corresponding to args, since the caller already owns the ones corresponding to `kwargs`)

it does make things a bit more complicated (and means we can't use kwargs to tweak behaviours in any way should we want to in the future; not sure if that's an issue)

i _think_ my preference is to expand the existing command rather than introducing a new, very similar one, but i'm open to other arguments against

timbu:
cool - will do 👍

mattbennett:
Can we not change the API to just accept named arguments? The named dependency would be replaced by the argument value, or a `MockDependencyProvider` if the value is `None`.

We could maintain backwards compatibility for people using positional arguments and raise a `DeprecationWarning` to notify of the changed API

davidszotten:
given that we got this far with the args only api, i guess that's a pretty common case so i think it would be nice to keep it

mattbennett:
Not sure I follow your argument.

With the changes in this PR, `replace_dependencies` has two slightly different APIs. One is basically an extension of the other. My suggestion is that, rather than maintaining and documenting both, we transition to the one that is most capable.

davidszotten:
i think there also an issue around the return value (because kwargs aren't ordered)

mattbennett:
Hmm, yes that is different. I would be OK with a different return format for the deprecated API though -- i.e. return a list if you provide args only, and a dict otherwise.

mattbennett:
Just making a note that we discussed this in person and decided that supporting both `args` and `kwargs` APIs was the best option. Will review in due course.

nameko/testing/services.py
```diff
@@ -218,29 +218,62 @@ def cm_to_inches(self, cms):
...
-    so that calls to the replaced dependencies can be inspected. Return a
-    single object if only one dependency was replaced, and a generator
-    yielding the replacements in the same order as ``names`` otherwise.
+    Any dependencies named in the `dependencies` args will be replaced with a
+    :class:`MockDependencyProvider` that provides a ``MagicMock()`` dependency.
+
+    The dependency provided by the ``MockDependencyProvider`` can be customised
+    by using the `dependency_map` kwargs. Each kwarg name is the dependency
+    name, and the value specifies the custom dependency that should be
+    provided.
```
mattbennett: I would rewrite these two paragraphs as:

> Dependencies named in `*dependencies` will be replaced with a :class:`MockDependencyProvider`, which injects a `MagicMock` instead of the dependency.
> 
> Alternatively, you may use keyword arguments to name a dependency and provide the replacement value that the `MockDependencyProvider` should inject.

nameko/testing/services.py
```diff
@@ -261,43 +294,45 @@ def cm_to_inches(self, cms):
...
 
         # assert that the dependency was called as expected
-        maths_rpc.divide.assert_called_once_with(100, 2.54)
+        mock_maths_rpc.divide.assert_called_once_with(100, 2.54)
 
-    """
-    if container.started:
-        raise RuntimeError('You must replace dependencies before the '
-                           'container is started.')
+    **Kwargs Usage**
```
mattbennett: I think we can keep the original heading as **Usage** and just give this an introductory sentence:

> Providing a specific replacement by keyword:

nameko/testing/services.py
```diff
@@ -261,43 +294,45 @@ def cm_to_inches(self, cms):
...
 
-    dependency_names = {dep.attr_name for dep in container.dependencies}
+    ::
 
-    missing = set(dependencies) - dependency_names
-    if missing:
-        raise ExtensionNotFound("Dependency(s) '{}' not found on {}.".format(
-            missing, container))
+        mock_maths_rpc = MagicMock()
```
mattbennett: This isn't the most compelling example because it doesn't achieve anything different from the other API. How about the following?

```python
stub_maths_rpc = RpcProxy('stub-maths')
```

You'd need to drop the assertion about the mock too. Maybe just state what you would do, rather than demonstrate it...

mattbennett:
@timbu there are just a couple of outstanding comments on this

nameko/testing/services.py
timbu: I've gone with a Stub class - let me know what you think :)

mattbennett:
@timbu thanks! happy for me to fix the spelling and add a changelog note in this branch?

mattbennett:
👍

PR #344, opened 2016-08-02
allow custom classes to be specified in config

mattbennett:
* Allows custom `ServiceContainer` and `WorkerContext` classes to be specified via config values
* Deprecates the `kwarg` interface for specifying these because they cannot be used in conjunction with the `nameko run` command


mattbennett:
@davidszotten this is ready for a round of review when you have a chance

nameko/containers.py
```diff
@@ -134,8 +144,16 @@ def __init__(self, service_cls, config, worker_ctx_cls=None):
         self.service_cls = service_cls
         self.config = config
 
-        if worker_ctx_cls is None:
-            worker_ctx_cls = WorkerContext
+        if worker_ctx_cls is not None:
+            warnings.warn(
```
davidszotten: we are starting to add a whole bunch of deprecation warnings. we might want some kind of plan for how long we keep the old behaviours around, and maybe add a grep-able comment marker to tell ourselves when (which version) these can be tidied up

davidszotten:
curious what your use-case for replacing these is

nameko/containers.py
mattbennett: Definitely agree with this. Something like two major releases or 6 months, whichever comes last?

I'll create document somewhere to describe which deprecation warnings we have and when they were introduced, so we know when they're eligible for removal.

mattbennett:
Our use-case is for a custom `ServiceContainer` class.

We deploy our services to multiple regions and need slightly different behaviour in each, but don't want to maintain different branches or codebases. For example, it's only valid to write to the database in one region.

The custom container supports an "entrypoint blacklist", a configuration level way to disable certain entrypoints, so we can deploy the same codebase in all regions and tweak the behaviour via config.

davidszotten:
ok. do you also use a custom worker context? (does that definitely need to be configurable. ofs used that to transition to nameko, but have you come across use-cases beyond that?)

nameko/utils.py
```diff
@@ -1,10 +1,11 @@
-import sys
 import inspect
 import re
+import sys
+from pydoc import locate
```
davidszotten: interesting. seems to have nice behaviour around missing target vs target having bad imports. could maybe use this in cli.run (separate pr)

nameko/containers.py
davidszotten: i think that info is better kept in the code (e.g. as suggested above)

test/testing/test_pytest_plugin.py
```diff
@@ -136,6 +136,113 @@ def test_container_factory(container_factory, rabbit_config):
     assert get_rabbit_connections(vhost, rabbit_manager) == []
 
 
+def test_container_factory_with_custom_container_cls(testdir):
+
+    testdir.makeconftest(
```
davidszotten: why not make an actual module?

mattbennett: as opposed to a dynamically generated one? seems nicer to have it all contained within this test, rather than having a file floating around somewhere else.

mattbennett:
We don't have a custom `WorkerContext`, but we can't get rid of it without another change:

`WorkerContext.context_keys` specifies which bits of the context data get passed from call to call, and the only way to set it is with a subclass. It's an odd API though. I'm not sure why we need a distinction between `data` on the worker context and `context_data`, which is the subset that matches the keys.

test/testing/test_pytest_plugin.py
davidszotten: no, i mean instead of `makeconftest` with a fake module, you could `makepyfile` with an actual module

mattbennett: Ha, yes that is a better option!

mattbennett:
@davidszotten as discussed in the [offline meetup](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/nameko-dev/WZ-xRYd2-vw), I've removed `WorkerContext.context_keys` and the ability to (directly) customise the worker context class.

nameko/containers.py
```diff
@@ -50,72 +50,59 @@ def get_container_cls(config):
...
+        self._parent_call_id_stack = self.data.pop(
+            CALL_ID_STACK_CONTEXT_KEY, []
         )
-        n = -self.parent_calls_tracked
-        self.call_id_stack = self.parent_call_stack[n:]
-        self.call_id_stack.append(self.call_id)
-        try:
-            self.immediate_parent_call_id = self.parent_call_stack[-1]
-        except IndexError:
-            self.immediate_parent_call_id = None
```
davidszotten: this drops a few "public" attributes on the worker context. is that intentional? don't know if it still works, but e.g. `nameko-sentry` looks to make use of `immediate_parent_call_id`

nameko/containers.py
```diff
@@ -50,72 +50,59 @@ def get_container_cls(config):
...
-        """ Return a tuple of keys describing data kept on this WorkerContext.
-        """
+    @property
+    def call_id_stack(self):
+        if self._call_id_stack is None:
+            parent_calls_tracked = self.container.config.get(
+                PARENT_CALLS_CONFIG_KEY, DEFAULT_PARENT_CALLS_TRACKED
+            )
+            stack_length = parent_calls_tracked + 1
+
```
davidszotten: probably overkill to go via a deque. e.g.

```
stack = self._parent_call_id_stack[-parent_calls_tracked:] + [self.call_id]
```

nameko/containers.py
```diff
@@ -50,72 +50,59 @@ def get_container_cls(config):
...
-        key_data[CALL_ID_STACK_CONTEXT_KEY] = self.call_id_stack
-        return key_data
-
-    @classmethod
-    def get_context_data(cls, incoming):
-        data = {k: v for k, v in six.iteritems(incoming)
-                if k in cls.context_keys}
+    @property
+    def context_data(self):
+        data = self.data.copy()
```
davidszotten: how come we re-build this every time?

mattbennett: Context data can be modified through the course of a worker executing

nameko/containers.py
mattbennett: That's quite a lot less readable. Also conceptually, the size-limited stack is exactly a deque.

nameko/containers.py
mattbennett: Ah, thank you. It was intentional to drop most of these, but I hadn't noticed that `immediate_parent_call_id` is read by `nameko-sentry` (which is in use in several places).

After adding it back, only `self.parent_call_stack` and `self.unique_id` have been removed, leaving (I think) a public API.

mattbennett:
@davidszotten this is ready for another round

nameko/containers.py
davidszotten: ok. don't feel that strongly. noted it because a) you only interact with the deque once (here) and b) you leave the attribute as a deque but convert it to a list every time it's read

nameko/containers.py
```diff
@@ -374,7 +373,8 @@ def spawn_managed_thread(self, fn, protected=None, identifier=None):
                 "The `protected` kwarg is now deprecated, and extensions "
                 "can pass an idenifier as a keyword argument for better "
                 "logging. See :meth:`nameko.containers.ServiceContainer."
-                "spawn_managed_thread`.", DeprecationWarning
+                "spawn_managed_thread`. This warning will be removed in "
+                "version 2.6.0.", DeprecationWarning
```
davidszotten: oh, i meant as a comment but i guess this works

davidszotten:
👍

PR #348, opened 2016-08-21
adaptable wsgi application i.e exposing wsgi call to the service method

eightnoteight:
Hi,
nameko is great thing to have multiple services in a project.
I'm used to flask api  a lot, which provides features such as templates, elegant Request, Response classes and a great number of extensions. 

as nameko  is just exposing werkzeug''s api, it would be very cool if the WsgiApp interface is exposed so that the users can just plug any popular WsgiApp like flask, tornado etc,..

But as I'm a beginner in contributing to this project, I wanted to start with an example implementation, take some pointers and then add the feature to the project.

I have tested the concurrency level of this example doubting any break between flask and eventlet, but it was fine.

testing cmd
```
ab -n 1000 -c 100 -p file.txt -T "multipart/form-data; multipart/form-data; boundary=------------------------d4d522c09e67e301" -X 127.0.0.1:8001 http://localhost:8000/upload
```
[file.txt](https://github.com/onefinestay/nameko/files/428736/file.txt)


jessepollak:
Interesting example! We've actually been discussing similar WSGI-extensibility in the mailing list [here](https://groups.google.com/forum/#!topic/nameko-dev/gFAnGc2tx-E) as well!

eightnoteight:
closing in favour of #352

Issue #349, opened 2016-08-21
nameko.exceptions.RemoteError: UnicodeEncodeError 'ascii' codec can't encode character u'\u6bb5' in position 0: ordinal not in range(128)


dxfmichaelgithub:
nameko support non ascii character?

>>> n.rpc.greeting_service.hello("段")
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/nameko/rpc.py", line 352, in __call__
    return reply.result()
  File "/usr/lib/python2.7/site-packages/nameko/rpc.py", line 338, in result
    raise deserialize(error)
RemoteError: UnicodeEncodeError 'ascii' codec can't encode character u'\u6bb5' in position 0: ordinal not in range(128)

davidszotten:
Hi,

This is not a problem with nameko, but a bug in the example `GreetingService`: 

https://github.com/onefinestay/nameko/blob/master/docs/examples/helloworld.py#L8 calls ` "Hello, {}!".format(name)`, which fails with non-ascii `name`s. You can fix it by prefixing the string with `u`, (so `"uHello, {}!".format(name)`) or by calling `from __future__ import unicode_literals` at the top of the `helloworld.py`

Leaving this open, as we should get our examples fixed

mattbennett:
Fixed by #350

PR #350, opened 2016-08-22
make examples work with unicode inputs

mattbennett:
Addresses #349


davidszotten:
i'm still on the fence about this. e.g. `u` prefix vs `unicode_literals`, or even just "assuming" py3

mattbennett:
I am in favour of assuming py3. Seems fitting for an example to use the best version of the language.

I will rework these.

mattbennett:
@davidszotten happy for this to land?

davidszotten:
👍

PR #351, opened 2016-08-25
Add secure connection support

hedin:
Allows to set SSL parameters in config 

```
SSL:
  ca_certs: '/path/to/cacert.pem'
  certfile: '/path/to/client_cert.pem'
  keyfile: '/path/to/client_key.pem'
```

They will be passed to kombu's connections across the library

### todo before merging

- [x] decide on config key name
- [ ] documentation
- [ ] end-to-end ci test


davidszotten:
Hi,

Thanks for the pull request!

A few initial thoughts:
`SSL` is a very generic key for something that only applies to the amqp extension. Maybe `AMQP_SSL` or `AMQP_SSL_CONFIG` or something along those lines?

(unsure about this, but maybe) do we want to somehow group together the hostname (currently passed as an arg, but could be a kwarg instead) and ssl parameters (internally) as`kombu_connection_params`, `amqp_connection_params` or similar to have them controlled from a single place?

test/test_amqp.py
```diff
@@ -21,6 +21,19 @@ def uris(rabbit_config):
     }
 
 
+@pytest.fixture
+def ssl_config(rabbit_config):
```
davidszotten: not sure having this as a fixture is worth it for the one place where it's used

nameko/amqp.py
```diff
@@ -38,12 +39,20 @@ class TestTransport(Transport):
     Connection = ConnectionTester
 
 
-def verify_amqp_uri(amqp_uri):
-    connection = Connection(amqp_uri)
+def verify_amqp_uri(amqp_uri, config=None):
```
mattbennett: It's odd to pass the `amqp_uri` and the whole `config` object (which contains the AMQP URI data).

I would prefer if the arguments here were `amqp_uri` and `ssl_params` (or just `ssl`)

nameko/amqp.py
```diff
@@ -38,12 +39,20 @@ class TestTransport(Transport):
...
 
     transport = TestTransport(connection.transport.client)
     with transport.establish_connection():
         pass
+
+
+def get_ssl_params(config):
+    params = config.get('SSL')
+    if params and isinstance(params, dict):
+        params.update({'cert_reqs': ssl.CERT_REQUIRED})
```
davidszotten: maybe we should at least look for this key as well to make it possible for users to set the behaviour if they want? (e.g. params['cert_reqs'] = params.get('cert_reqs', CERT_REQUIRED)` or similar

or is this never the case?

davidszotten:
is there a good way to end-to end test this (on e.g. travis)?

nameko/amqp.py
davidszotten: i know config is still up in the air, but eventually we might even move to something like a dict with all amqp config, e.g.

```
amqp:
  url: amqp://localhost,
  ssl: 
    ca_certs: ...
    ...
```

(or even)

```
amqp:
  default:
    url: amqp://localhost,
    ssl: 
      ca_certs: ...
      ...
  other:
    url: ...
```

mattbennett: I like this idea, but still think that `verify_amqp_uri` should either take the whole config object, or the uri and ssl params separately.

mattbennett:
It seems not: https://github.com/travis-ci/travis-ci/issues/6379

We seem to be fighting with Travis more and more. Time to switch to CircleCI?

mattbennett:
It's actually pretty easy to run travis builds against a dockerized rabbitmq (e.g. https://github.com/nameko/nameko-amqp-retry/blob/master/.travis.yml), so end-to-end tests are quite possible here. It's a one-liner to start a rabbit broker and pass the certs in.

@hedin are you interested in picking this PR back up again?

roblofthouse:
We're very much interested in getting this into Nameko as we require secure service to service communication.

hedin:
Ok, I'll reup it in a couple of days

jakedahn:
I'm also interested in the functionality provided by this PR -- I'm willing and happy to pitch in to help get these changes merged into master.

@hedin @mattbennett What are the remaining items that need to be figured out before this pr is mergeable?

mattbennett:
@jakedahn apologies for the delay. I think it just needs an end-to-end test on CI, which probably means switching over to a dockerized rabbit rather the Travis service.

I will take a look in the next couple of days and confirm exactly what's needed.

nameko/constants.py
```diff
@@ -3,6 +3,7 @@
 RPC_EXCHANGE_CONFIG_KEY = 'rpc_exchange'
 SERIALIZER_CONFIG_KEY = 'serializer'
 HEARTBEAT_CONFIG_KEY = 'HEARTBEAT'
+SSL_CONFIG_KEY = 'ssl'
```
davidszotten: while we aren't exactly in a good place at the moment, i'd like to at least consider a more specific key here. we are currently often conflating rpc and amqp and have e.g. a top level key `heardbeat` which only applies to the rabbit broker, but unlike "heartbeat", ssl might confuse users of the web extension.

whilst fixing existing ones may require breaking backwards compat and is a job for later, i thought this worth bringing up before adding to this

hedin: Something like AMQP_SSL_CONFIG_KEY?

davidszotten: i'm less worried about the name of the constant (though your suggestion is good) than the value of the key that needs to go into `config.yml`, currently just `ssl`

hedin: Oh I get it. Will `AMQP_SSL_CONFIG_KEY = 'amqp_ssl'` be ok?

davidszotten: sounds good to me!

mattbennett:
I'm happy for this to land when the two remaining items on the todo list in the description are done.

mattbennett:
These tests will fail against the built-in Travis RabbitMQ broker because it's not configured for SSL.

I have spun up a dockerized RabbitMQ with SSL enabled using the following steps:

1. Generate ca, server and client certs following https://www.rabbitmq.com/ssl.html
2. Gather certs and keys from the server and ca directories into a directory (`ssl`)
3. Run dockerized rabbit identifying the appropriate certs/keys:

```
docker run -it --rm -v "$PWD"/ssl:/ssl -e RABBITMQ_SSL_CACERTFILE=/ssl/cacert.pem -e RABBITMQ_SSL_CERTFILE=/ssl/servercert.pem -e RABBITMQ_SSL_KEYFILE=/ssl/serverkey.pem  --hostname rabbit --name rabbit -p 15672:15672 -p 5672:5672 -p 15671:15671 -p 5671:5671 -e RABBITMQ_DEFAULT_USER=guest -e RABBITMQ_DEFAULT_PASS=guest rabbitmq:3-management
```

Unfortunately, in the official RabbitMQ docker image, enabling SSL disables the normal ports (https://github.com/docker-library/rabbitmq/issues/79) so it can't be used here as a trivial drop-in replacement for the built-in broker.

I guess the next step would be to create our own Dockerfile to run a containerised RabbitMQ that runs with SSL and non-SSL ports enabled simultaneously.

hedin:
Yep! I had to use my local non-docker rabbitmq server for testing.

mattbennett:
@hedin I have created a dockerized rabbitmq for testing -- https://hub.docker.com/r/nameko/nameko-rabbitmq/. The image includes self-signed certs and keys for the server, client and CA under /mnt/certs.

You should be able to use this to write an end-to-end test on CI.

hedin:
@mattbennett thanks for this container it works perfectly. Should I write more tests than just testing SSL connection?
If so, could you please tell me what tests you want to see. I'm just starting my way in testing.

gianchub:
Hello all, we're very interested in this functionality to land in nameko. What is the current state of things for this PR? Is anyone on it? Would the author be happy to accept contributions so that this can finally land?

mattbennett:
Sorry, I didn't reply to @hedin above. I would like to see an end-to-end test over SSL. Just a simple service talking AMQP to an SSL enabled rabbit.

It should be simple to add now we have a docker container that will accept SSL and non-SSL connections.

gianchub:
Thanks @mattbennett :) Please do let us know if we can help getting this in quickly.

hedin:
Hi, @mattbennett . While I'm working on updating my fork and adding tests, could you please update certificates in docker image. They are expired:
```bash
openssl s_client -connect localhost:5671 -cert clientcert.pem -key clientkey.pem -CAfile cacert.pem -tls1_2
...
Verify return code: 10 (certificate has expired)
```

mattbennett:
@hedin new images with updated certs are available now. Valid for 10 years this time ;)

hedin:
Hi, @mattbennett . I've added simple rpc over ssl test.

https://github.com/hedin/nameko/blob/a8f386c7a7dbb639781abe762043808cd5f55c06/test/test_rpc.py#L1389-L1397

test/test_rpc.py
```diff
@@ -1384,3 +1384,14 @@ def method(self):
 
         with eventlet.Timeout(1):
             [reply.result() for reply in replies]
+
+
+def test_rpc_over_ssl(container_factory, runner_factory, rabbit_ssl_config):
+    container = container_factory(ExampleService, rabbit_ssl_config)
+    container.start()
+
+    with entrypoint_hook(container, 'echo') as echo:
```
mattbennett: `entrypoint_hook` triggers the entrypoint without actually using whatever transport it's connected to, so this isn't using AMQP.

You need to generate a real AMQP message, for example with the standalone RPC proxy:

``` python
with ServiceRpcProxy("service", rabbit_config) as proxy:
    assert proxy.echo("foo") == "foo"
``` 

The proxy can connect using SSL or not. In fact, it might be nice to test both.

hedin: Thanks for comment! My code doesn't work now on real messaging. Going to find why

hedin:
Ok, I found more places to pass ssl_params and it works now.

> The proxy can connect using SSL or not. In fact, it might be nice to test both.

Should I update the test to use both ssl and non-ssl connection at the same time?

mattbennett:
This PR actually enables SSL for all the AMQP extensions, not just RPC. So for completeness we should probably have the following tests:

test_rpc.py:
• test_rpc_over_ssl: handling an rpc request in an ssl-connected service
• test_rpc_proxy_over_ssl: using an rpc proxy over ssl
standalone/test_rpc.py:
 • test_rpc_proxy_over_ssl: same as above but using standalone proxy
test_events.py:
 • test_event_handler_over_ssl: handling an event in an ssl-connected service
 • test_event_dispatcher_over_ssl: dispatching an event in an ssl-connected service
standalone/test_events.py:
 • test_event_dispatcher_over_ssl: same as above but using standalone dispatcher
test_messaging.py:
 • test_consume_over_ssl: handling a message in an ssl-connected service
 • test_publish_over_ssl: publishing a message from an ssl-connected service

In each case there are two channels of communication, client->broker and broker->service, and it's possible for one or both legs to be SSL'd. I think it's sufficient to test just the appropriate leg in each test.

In the interest of shipping this change soon, I'd be happy to add these on top of your commits, if that's OK with you?

hedin:
Sure!

mattbennett:
Commits on top of this branch are in #524. Thanks again @hedin for raising this PR.

mattbennett:
@hedin @roblofthouse @gianchub @jakedahn @pimiela @bweaver-cxp

This functionality is now (finally) available in a pre-release build on PyPI: 2.9.1rc0. I don't have the ability to test it extensively, so could one or more of you please try and report back if everything seems good and stable?

Thanks @hedin for the original PR.

bweaver-cxp:
Sorry for the delay; the project we're doing in nameko has been put on hold for several months, so it took a while to carve out some time to run that system again. My testing wasn't extensive either, but this build worked nicely under the testing I did do.

mattbennett:
Thanks for the feedback @bweaver-cxp

PR #357, opened 2016-09-13
Connection pool for multithreaded gevent-monkeypatched apps (e. g. using Nameko with Django)

and3rson:
Hi,

I've recently discovered Nameko and instantly fell in love with it. Great work!

The only thing I am missing was connection pooling mechanism: spamming server with RPC calls sometimes led to GEvent exceptions about using same socket by multiple greenlets.

So I decided to implement a simple connection pooling class to use Nameko with my Django apps.

Usage is fairly simple:

```
from nameko.standalone.rpc import RPCProxyPool

pool = RPCProxyPool(pool_size=8)

# Within django view or model:
with pool.next() as rpc:
    rpc.mailer.send_mail(foo='bar')
```

It uses gevent-powered Queue for pooling connections. Once there are no more free connections to reuse, it will block until a connection becomes available. Later we can implement dynamic pool size to create more connections if the pool is drained for a longer time.

I tested my code with siege, and here's what happened:
- Creating one ClusterRPCProxy per request: 10~20 RPS + driving RabbitMQ nuts (~50% availability).
- Creating global shared ClusterRPCProxy for all requests: ~50 RPS with periodic crashes (~70% availability)
- Using RPCProxyPool: ~110 RPS (100% availability).

I hope this code helps. I'm open to any comments to make it even better.

### DIFF
--- a/nameko/standalone/rpc.py
+++ b/nameko/standalone/rpc.py
@@ -16,6 +16,7 @@ from nameko.exceptions import RpcConnectionError, RpcTimeout
 from nameko.rpc import ServiceProxy, ReplyListener
 from nameko.constants import (
     SERIALIZER_CONFIG_KEY, DEFAULT_SERIALIZER)
+from gevent.queue import Queue
 
 
 _logger = logging.getLogger(__name__)
@@ -327,3 +328,49 @@ class ClusterRpcProxy(StandaloneProxyBase):
     def __init__(self, *args, **kwargs):
         super(ClusterRpcProxy, self).__init__(*args, **kwargs)
         self._proxy = ClusterProxy(self._worker_ctx, self._reply_listener)
+
+
+class RPCProxyPool(object):
+    """
+    Connection pool for Nameko RPC cluster.
+
+    Pool size can be customized by passing `pool_size` keyword argument to the constructor.
+    Default size is 4.
+
+    Example usage:
+
+        pool = RPCProxyPool()
+        # ...
+        with pool.next() as rpc:
+            rpc.mailer.send_mail(foo='bar')
+
+    This class is thread-safe and designed to work with GEvent.
+    """
+    class RPCContext(object):
+        def __init__(self, pool, config):
+            self.pool = pool
+            self.proxy = ClusterRpcProxy(config)
+            self.rpc = self.proxy.start()
+
+        def __enter__(self):
+            return self.rpc
+
+        def __exit__(self, *args, **kwargs):
+            self.pool._put_back(self)
+
+    def __init__(self, pool_size=4):
+        self.queue = Queue()
+        for i in xrange(pool_size):
+            ctx = RPCProxyPool.RPCContext(self)
+            self.queue.put(ctx)
+
+    def next(self):
+        """
+        Fetch next connection.
+
+        This method is thread-safe.
+        """
+        return self.queue.get()
+
+    def _put_back(self, ctx):
+        self.queue.put(ctx)
--- a/nameko/standalone/rpc.py
+++ b/nameko/standalone/rpc.py
@@ -331,13 +331,12 @@ class ClusterRpcProxy(StandaloneProxyBase):
 
 
 class RPCProxyPool(object):
-    """
-    Connection pool for Nameko RPC cluster.
+    """ Connection pool for Nameko RPC cluster.
 
     Pool size can be customized by passing `pool_size` keyword argument to the constructor.
     Default size is 4.
 
-    Example usage:
+    *Usage*
 
         pool = RPCProxyPool()
         # ...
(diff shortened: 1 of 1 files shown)


mattbennett:
There is some overlap here with https://github.com/clef/flask-nameko. Connection pooling is clearly a use-case we need to support. The performance improvement is very significant!

We intentionally made the standalone package independent of eventlet to avoid clashes with whatever concurrency pattern the client is using. 

This is a nice PR but it might be better as a separate library, as flask-nameko is, unless there's a non-eventlet based pool implementation we can use here. I think that would be the best solution, if possible.

@davidszotten thoughts?

and3rson:
Hey @mattbennett, thanks for the response!

Dependency on `eventlet` is not necessary, it can simply be replaced by built-in `Queue` implementation which can be monkey-patched by `gevent` or whatever the client is using. I just used eventlet because it seemed like nameko is using it as well.

davidszotten:
for production usage of the standalone proxy, i agree that a connection pool is required to get reasonable performance. perhaps to the extent that we should ship one in nameko. i think the library from clef linked above has a slightly nicer implementation, e.g. only instantiating proxies on demand.

if we do provide one, it would be nice to have enough hooks to e.g. tie in to the django/flask request cycle

there are some further considerations, e.g. around call ids that i think the version of this i wrote for ofs (never got open sourced iirc)

and3rson:
I'm thinking of probably moving this out into some new `django-nameko` package. Some Django support would be pretty convenient, I think.

mattbennett:
This could be enhanced to instantiate proxies only on demand, no? With that change (and using non-eventlet primitives) this would be an enhancement of nameko's out-of-the-box offering. It would be even better if it pooled connections inside the existing API, rather than adding a new one.

A new `django-nameko` package would be cool too, especially if there are Django specific additions to make. That's probably the path of least resistance would be a nice outcome.

and3rson:
@mattbennett Right. Currently it blocks if the pool is drained. We can make it spawn more proxies in this case.

I'm currently packaging it within `django-nameko` package in order to do some good testing in real projects, will be in touch.

mattbennett:
Great stuff, thanks. Looking forward to the results!

and3rson:
In case you're interested, here's the plugin for Django: https://github.com/and3rson/django-nameko

jessepollak:
@and3rson nice work! I wrote the `flask-nameko` library, which is super similar. @davidszotten and @mattbennett - since there's so much overlap between `flask-nameko` and `django-nameko`, it seems like we should definitely try to consolidate the logic / call style into something standard across the two of them (at least at the underlying implementation level).

I'd much prefer to have a connection pooled proxy merged into `nameko` than maintain an outside project - what are the major blockers / constraints to make that happen? @mattbennett I know you mentioned you'd want it to fall within the existing APIs, can you flesh that idea out a little bit more?

Thanks!

mattbennett:
@jessepollak there is some overlap here with our recent [mailing list thread](https://groups.google.com/forum/#!topic/nameko-dev/woyoV9Iz5mk).

I think the ideal API here has the following qualities:

* Is backwards compatible with the existing proxy (i.e. no change in the API, or having to interact with the pool)
* Connections are pooled and re-used rather than creating a new one for each use
* The pool supports a minimum number of connections (to be created at startup) and a maximum number (and blocks or raises when exhausted), and others are created on demand.
* Stale connections are detected and recovered transparently

and3rson:
@jessepollak @mattbennett Thanks for your feedback, guys!
I'm willing to cooperate on this, so let me know if I can make some help.

jessepollak:
@mattbennett just to clarify, do you think this should replace the `ClusterRpcProxy` or should it be a new class?

mattbennett:
@jessepollak if it can be backwards compatible I think we should consider it an enhancement to the existing `ClusterRpcProxy`.

If you decide to experiment with a different AMQP library as per the mailing list, that would be best as a PoC.

albertmenglongli:
@and3rson , your django-nameko is cool as a client to call rpc, any idea  how to provide as rpc service in django using nameko, thanks in advance.

lewisoaten:
Did this endeavour die off? A pool would be pretty useful!

ArtikUA:
Great pool request!
Nameko should have pools anyway: it's very costful to create and close new connection for any request
Otherwise, everyone should create similar realization on their own project

julianser:
To echo what was said earlier, great pool request!

However, many people probably want to change the next() function to this:

```
    def next(self, timeout=None):
        """ Fetch next connection.
        This method is thread-safe.
        """
        return self.queue.get(timeout=timeout)
```

This will ensure that it never times out. On the other hand, setting "timeout=False" by default (as it was before) ensures that it times out immediately if the queue is empty. See also the original queue get() function:

https://github.com/python/cpython/blob/3.7/Lib/queue.py#L153

PR #362, opened 2016-09-19
Basic support for environment variables.

kooba:
This PR adds basic support for environment variables in configuration file.

After playing around with different regexes and yaml constructors I came to conclusion that basic (strings only) support for env vars in config file will cover 80% of use cases. 

In https://github.com/onefinestay/nameko/issues/223 @fabiocerqueira mentioned it would be nice to add support for other data types but implementation was getting too hairy. If we really need to add support for that, the simplest way would be to process config file before loading it with `yaml` package. This could be easily accomplished by bringing in `jinja` templates and using their custom filters functionality. I didn't feel bringing-in dependency like that would be justified for this functionality. 

Additionally looking around I didn't see many project providing such a support (https://docs.docker.com/compose/compose-file/#/variable-substitution) 

All the logic is in `nameko.utils` and initialised in `nameko.cli.main` please let me know if that's desired. 

There is a catch that if value is quoted, explicit `!env_var` resolver has to be used. Possibly should document that. https://github.com/onefinestay/nameko/pull/362/files#diff-85c7da7d97a696e8017da5742a876475R212

I had regexes set on constants and added to `constants.py` byt at the end decided to add them inline.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -24,7 +24,8 @@ and providing a simple YAML configuration file:
 .. code-block:: yaml
 
     # foobar.yaml
-    AMQP_URI: 'amqp://guest:guest@localhost'
+
+    AMQP_URI: 'pyamqp://guest:guest@localhost'
     WEB_SERVER_ADDRESS: '0.0.0.0:8000'
     rpc_exchange: 'nameko-rpc'
     max_workers: 10
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -62,7 +62,7 @@ To run your service and set environment variables for it to use:
 
     $ RABBITMQ_USER=user RABBITMQ_PASSWORD=password RABBITMQ_HOST=host nameko run --config ./foobar.yaml <module>[:<ServiceClass>]
 
-If you need to quote the values in your YAML file, the explicit !env_var resolver is required:
+If you need to quote the values in your YAML file, the explicit ``!env_var`` resolver is required:
 
 .. code-block:: yaml
 
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -60,6 +60,12 @@ Run your service with environment variables set:
 
     $ RABBITMQ_USER=user RABBITMQ_PASSWORD=password RABBITMQ_HOST=host nameko run --config ./foobar.yaml <module>[:<ServiceClass>]
 
+If you need to quote your yaml values explicit resolver ``!env_var`` is required
+
+.. code-block:: yaml
+
+    # foobar.yaml
+    AMQP_URI: !env_var "amqp://${RABBITMQ_USER:guest}:${RABBITMQ_PASSWORD:password}@${RABBITMQ_HOST:localhost}"
 
 Interacting with running services
 ---------------------------------
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -43,6 +43,24 @@ and providing a simple YAML configuration file:
 The ``LOGGING`` entry is passed to :func:`logging.config.dictConfig` and should conform to the schema for that call.
 
 
+Environment variable substitution
+---------------------------------
+YAML configuration files have basic support for environment variables.
+You can use bash style syntax: ``${ENV_VAR}`` optionally you can provide default values ``${ENV_VAR:default_value}``
+
+
+.. code-block:: yaml
+
+    # foobar.yaml
+    AMQP_URI: amqp://${RABBITMQ_USER:guest}:${RABBITMQ_PASSWORD:password}@${RABBITMQ_HOST:localhost}
+
+Run your service with environment variables set:
+
+.. code-block:: shell
+
+    $ RABBITMQ_USER=user RABBITMQ_PASSWORD=password RABBITMQ_HOST=host nameko run --config ./foobar.yaml <module>[:<ServiceClass>]
+
+
 Interacting with running services
 ---------------------------------
 
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -46,7 +46,8 @@ The ``LOGGING`` entry is passed to :func:`logging.config.dictConfig` and should 
 Environment variable substitution
 ---------------------------------
 YAML configuration files have basic support for environment variables.
-You can use bash style syntax: ``${ENV_VAR}`` optionally you can provide default values ``${ENV_VAR:default_value}``
+You can use bash style syntax: ``${ENV_VAR}``
+Optionally you can provide default values ``${ENV_VAR:default_value}``
 
 
 .. code-block:: yaml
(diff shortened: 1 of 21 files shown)


davidszotten:
a few general comments: will try to look at code in more depth later:

type casting. i think that's fine to leave up to the user (extension author)
i'm not a big fine of `utils.py` (or `helper`, `misc` etc). just becomes a grab bag of unrelated stuff. can this live in `cli.config` or something similar? (or even just in the one module that uses it)

nameko/utils.py
```diff
@@ -206,3 +208,22 @@ def import_from_path(path):
...
+
+def _replace_env_var(match):
+    env_var, _, default = match.groups()
+    return os.environ.get(env_var, default)
+
+
+def _env_var_constructor(loader, node):
+    value = loader.construct_scalar(node)
+    return re.compile(
+        r'\$\{([a-zA-Z][^}:\s]+)(:([^}]+))?\}', re.MULTILINE | re.DOTALL
```
davidszotten: how come we want multiline here?

kooba:
I had it directly on `main` module initially but somehow decided to move it to `utils`. Will move it back there.

nameko/utils.py
kooba: You're right, seems redundant here, we are only processing one line at a time.

nameko/cli/main.py
```diff
@@ -19,9 +22,27 @@ def setup_parser():
     return parser
 
 
+def _replace_env_var(match):
+    env_var, _, default = match.groups()
```
mattbennett: What's the second match here? Can't we have the regex discard it?

nameko/cli/main.py
```diff
@@ -19,9 +22,27 @@ def setup_parser():
...
+    return os.environ.get(env_var, default)
+
+
+def _env_var_constructor(loader, node):
+    value = loader.construct_scalar(node)
+    return re.compile(
+        r'\$\{([a-zA-Z][^}:\s]+)(:([^}]+))?\}'
```
mattbennett: I think I would prefer these defined as variables, although in this module rather the constants.py. That allows you to give them readable names, e.g.

``` python
env_var_matcher = re.compile(r'\$\{([a-zA-Z][^}:\s]+)(:([^}]+))?\}')
quoted_env_var_matcher = re.compile(r'.*\$\{.*\}.*')) 
```

mattbennett: Also worth following raymondh's advice here https://twitter.com/raymondh/status/766654913526890496?lang=en-gb

docs/cli.rst
```diff
@@ -43,6 +43,24 @@ and providing a simple YAML configuration file:
...
+    # foobar.yaml
+    AMQP_URI: amqp://${RABBITMQ_USER:guest}:${RABBITMQ_PASSWORD:password}@${RABBITMQ_HOST:localhost}
+
+Run your service with environment variables set:
+
+.. code-block:: shell
+
+    $ RABBITMQ_USER=user RABBITMQ_PASSWORD=password RABBITMQ_HOST=host nameko run --config ./foobar.yaml <module>[:<ServiceClass>]
+
+
```
mattbennett: I think you should mention the `!env_var` resolver here for quoted values.

kooba:
Comments addressed @mattbennett

nameko/cli/main.py
```diff
@@ -1,10 +1,40 @@
...
 from . import backdoor, run, shell
 
+ENV_VAR_MATCHER = re.compile(
+    r"""
+        \$\{        # match characters ${ literally
+
+        ([^}:\s]+)  # 1st capturing group:
+                    # matches any character except } or :
+
+        (?::)?      # non-capturing group
```
davidszotten: i guess this doesn't need to be a group at all? could just be `:?`

nameko/cli/main.py
```diff
@@ -1,10 +1,40 @@
...
+                    # matches between zero and one times character : literally
+
+        ([^}]+)     # 2nd capturing group
+                    # matches any character except }
+
+        ?\}         # match character } literally
```
davidszotten: since the question mark applies to the previous group, maybe keep it with that?

nameko/cli/main.py
```diff
@@ -1,10 +1,40 @@
...
+    """, re.X
```
davidszotten: i think `re.VERBOSE` is more helpful for the reader than `re.X`

test/cli/test_main.py
```diff
@@ -52,3 +54,51 @@ def test_flag_action(param, value):
...
+            'FOO: http://${BAR:foo}/${FOO:bar}',
+            {'BAR': 'bar', 'FOO': 'foo'},
+            {'FOO': 'http://bar/foo'}),
+    ])
+    def test_environment_vars_in_config(
+        self, yaml_config, env_vars, expected_config
+    ):
+        setup_yaml_parser()
+
+        for val in ['FOO', 'BAR']:
```
davidszotten: might use `patch.dict` here to avoid leaking side-effects

kooba:
Comments addressed @davidszotten

nameko/cli/main.py
```diff
@@ -1,10 +1,31 @@
...
+import yaml
 
 from nameko.exceptions import CommandError, ConfigurationError
 from . import backdoor, run, shell
 
+ENV_VAR_MATCHER = re.compile(
+    r"""
+        \$\{       # match characters `${` literally
+        ([^}:\s]+) # 1st group: matches any character except `}` or `:`
+        :?         # matches between zero and one times character `:` literally
```
mattbennett:
```
# matches the literal `:` character zero or one times
```
would be a clearer way to say this

docs/cli.rst
```diff
@@ -43,6 +43,30 @@ and providing a simple YAML configuration file:
 The ``LOGGING`` entry is passed to :func:`logging.config.dictConfig` and should conform to the schema for that call.
 
 
+Environment variable substitution
+---------------------------------
+YAML configuration files have basic support for environment variables.
+You can use bash style syntax: ``${ENV_VAR}`` optionally you can provide default values ``${ENV_VAR:default_value}``
```
mattbennett: I would say
```
You can use bash style syntax: ``${ENV_VAR}``

Optionally you can also provide default values: ``${ENV_VAR:default_value}``
```

docs/cli.rst
```diff
@@ -43,6 +43,30 @@ and providing a simple YAML configuration file:
...
+
+
+.. code-block:: yaml
+
+    # foobar.yaml
+    AMQP_URI: amqp://${RABBITMQ_USER:guest}:${RABBITMQ_PASSWORD:password}@${RABBITMQ_HOST:localhost}
+
+Run your service with environment variables set:
```
mattbennett:
```
To run your service and set environment variables for it to use:
```

docs/cli.rst
```diff
@@ -43,6 +43,30 @@ and providing a simple YAML configuration file:
...
+
+.. code-block:: shell
+
+    $ RABBITMQ_USER=user RABBITMQ_PASSWORD=password RABBITMQ_HOST=host nameko run --config ./foobar.yaml <module>[:<ServiceClass>]
+
+If you need to quote your yaml values explicit resolver ``!env_var`` is required
```
mattbennett: If you need to quote the values in your YAML file, the explicit ``!env_var`` resolver is required:

kooba:
@mattbennett copy changes addressed

mattbennett:
Thanks @kooba and @fabiocerqueira!

PR #363, opened 2016-10-06
require 100% coverage of test files

mattbennett:
Adds a requirement to reach 100% coverage of test files, to avoid mistakes like https://github.com/onefinestay/nameko/pull/337/commits/9c3a247139151184930b5afd0f90eba63360c361


mattbennett:
@davidszotten can you give these changes a once-over and make sure they seem fine?

.coveragerc
```diff
@@ -2,7 +2,14 @@
...
 data_file = /tmp/.coverage
+source =
+    nameko
+    test
 
 [report]
 show_missing = true
 fail_under = 100
+omit =
+    conftest.py
```
davidszotten: how come we are omitting these?

test/test_container.py
```diff
@@ -602,7 +597,7 @@ class Service(object):
 
             @dummy
             def method(self):
-                pass
+                pass  # pragma: no cover
```
davidszotten: how come we need these?

test/test_events.py
```diff
@@ -319,17 +316,6 @@ def make(base, prefixes):
     return make
 
 
-def test_event_handler_event_type():
```
davidszotten: is this not a test?

test/test_events.py
```diff
@@ -350,9 +336,7 @@ def test_service_pooled_events(rabbit_manager, rabbit_config,
                            properties=dict(content_type='application/json'))
 
     # a total of two events should be received
-    with eventlet.timeout.Timeout(EVENTS_TIMEOUT):
```
davidszotten: is this not protecting a race?

test/test_events.py
```diff
@@ -467,22 +445,29 @@ def test_broadcast_events(rabbit_manager, rabbit_config, start_containers):
 
 def test_requeue_on_error(rabbit_manager, rabbit_config, start_containers):
     vhost = rabbit_config['vhost']
-    start_containers(RequeueingHandler, ('requeue',))
+    container, = start_containers(RequeueingHandler, ('requeue',))
```
davidszotten: i like brackets around this kind of tuple unpacking, for clarity

`(foo,) = ...`

.coveragerc
mattbennett: Actually there is no good reason. I will remove these

test/test_container.py
mattbennett: The entrypoints need to be defined for various reasons (e.g. here the class would not be considered a service without one) but they're never actually executed.

test/test_events.py
mattbennett: It is, but a pointless one. It should have been deleted rather than just updated when we removed the `Event` superclass.

test/test_events.py
mattbennett: I will add these

davidszotten:
i'm actually a little surprised we weren't already doing this. i've started requiring 100% coverage of tests even for projects that don't achieve this for the library code (yet), to avoid that kind of mistake (or duplicate test names etc)

test/test_events.py
mattbennett: They assumed that the events would not be received immediately and we'd need to wait until they were processed.

As it happens, `rabbit_manager.publish` appears to always deliver the message before returning, so they're not required. I'll add resilience if this isn't reliably the case, but I'd rather not unless I have to.

mattbennett:
I was surprised too -- I thought we already required it.

davidszotten:
👍

PR #366, opened 2016-10-10
remove nameko_doc because it's outdated and unsupported

mattbennett:


davidszotten:
do you folks not use anything like this?

mattbennett:
We don't.

I can't even figure out the API here, so I think it best to drop this code. If something like this feature is useful it could easily (and better) be developed as a standalone package.

davidszotten:
sounds good to me

PR #369, opened 2016-10-14
Fix ws cleanup

### LINKED ISSUE #367: cleanup_websocket bug
It seems there is a bug in cleanup_websocket (websocket.py) which fails to remove a socket_id from the subscription list. 

```
def cleanup_websocket(self, socket_id):
    con = self.hub.connections.pop(socket_id, None)
    if con is not None:
        for channel in con.subscriptions:
            subs = self.hub.subscriptions.get(channel)
            if subs:
                subs.discard(channel) <-- This should be socket_id
```

Channel is being discarded from the list of socket_ids instead of the socket_id that is disconnecting.

davidszotten:
fix #367

### DIFF
--- a/nameko/testing/pytest.py
+++ b/nameko/testing/pytest.py
@@ -287,7 +287,9 @@ def websocket(web_config_port):
             '127.0.0.1', web_config_port)
         gr = eventlet.spawn(ws_app.run_forever)
         active_sockets.append((gr, ws_app))
-        return wait_for_sock()
+        socket = wait_for_sock()
+        socket.app = ws_app
+        return socket
 
     try:
         yield socket_creator
--- a/nameko/web/websocket.py
+++ b/nameko/web/websocket.py
@@ -157,7 +157,7 @@ class WebSocketHubProvider(DependencyProvider):
             for channel in con.subscriptions:
                 subs = self.hub.subscriptions.get(channel)
                 if subs:
-                    subs.discard(channel)
+                    subs.discard(socket_id)
 
 
 class WebSocketHub(object):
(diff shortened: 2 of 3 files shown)


nameko/testing/pytest.py
```diff
@@ -287,7 +287,9 @@ def socket_creator():
             '127.0.0.1', web_config_port)
         gr = eventlet.spawn(ws_app.run_forever)
         active_sockets.append((gr, ws_app))
-        return wait_for_sock()
+        socket = wait_for_sock()
+        socket.app = ws_app
```
mattbennett: If we're going to attach the `ws_app` to the `socket` object, perhaps `make_virtual_socket` should pass it in at [construction time](https://github.com/onefinestay/nameko/blob/master/nameko/testing/websocket.py#L45)?

Also, I found it confusing that this variable is called `socket`, assuming it to be the underlying socket object. It is really a websocket proxy (although "proxy" is also a heavily used word with other meanings).

davidszotten: pass it in? not sure i follow. also, as i mentioned in the commit message, don't like the api but it maintains backwards compat. 

called it `socket` since it's the return value from `socket_creator`

mattbennett: `make_virtual_socket` instantiates the `socket` object and the `ws_app` object, and returns both. I was suggesting to make the `app` an explicit member of the `Socket` class, rather than attaching it here.

I also think `socket_creator` is also a confusing name, but let's not muddy the waters of this PR.

davidszotten: if we prefer we could reduce this pr to just fixing the bug (without the test and the changing fixture)

mattbennett: I'm happy to delegate all these changes to another PR. I would prefer to land this with the test though.

mattbennett:
Just to be clear, I'm happy to land these changes as they are 👍

PR #370, opened 2016-10-18
regression test for #368

### LINKED ISSUE #368: OSError: raw readinto() returned invalid length -1 (should have been between 0 and 8192)
After running service for a while, this error occurred. Not sure if you can reproduce without incoming requests, but it happens to me every time.

Here's my code

``` python
class FetchService:
    name = "fetch_userdata_service"

    @timer(interval=86400)
    def update_token(self):
        # update token every day
        if os.path.isfile(TOKEN_FILE):
            os.remove(os.path.join(current_path, TOKEN_FILE))
            print("Old token removed")
        client.login_in_terminal(username, password)
        client.save_token('token.pkl')
        print("New token saved")

    @http('GET', '/ctrl/stop')
    def stop_service(self, request):
        runner.kill()

    @http('GET', '/<string:userhash>')
    def get_method(self, request, userhash):
        print("request userhash: " + userhash)
        u = client.people(userhash)
        # some lines ignored
        return json.dumps(data, separators=(',', ':'), ensure_ascii=False)

if __name__ == '__main__':
    logger.info("\n\nNameko Service Start...\n")
    runner = ServiceRunner(config={})
    runner.add_service(FetchService)
    runner.start()
    runner.wait()
```

The `client` is something similar to an HTTP client that make requests. And here's error log:

``` bash
INFO:__main__:

Nameko Service Start...

INFO:nameko.runners:starting services: fetch_userdata_service
ERROR:nameko.containers:<ServiceContainer [fetch_userdata_service] at 0x7fce53b92ac8> thread exited with error
  File "/usr/local/lib/python3.5/dist-packages/nameko/containers.py", line 477, in _handle_thread_exited
    gt.wait()
  File "/usr/local/lib/python3.5/dist-packages/eventlet/greenthread.py", line 175, in wait
    return self._exit_event.wait()
  File "/usr/local/lib/python3.5/dist-packages/eventlet/event.py", line 125, in wait
    current.throw(*self._exc)
  File "/usr/local/lib/python3.5/dist-packages/eventlet/greenthread.py", line 214, in main
    result = function(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/eventlet/wsgi.py", line 719, in process_request
    proto.__init__(sock, address, self)
  File "/usr/lib/python3.5/socketserver.py", line 681, in __init__
    self.handle()
  File "/usr/lib/python3.5/http/server.py", line 424, in handle
    self.handle_one_request()
  File "/usr/local/lib/python3.5/dist-packages/eventlet/wsgi.py", line 330, in handle_one_request
    self.raw_requestline = self.rfile.readline(self.server.url_length_limit)
OSError: raw readinto() returned invalid length -1 (should have been between 0 and 8192)
INFO:nameko.containers:killing <ServiceContainer [fetch_userdata_service] at 0x7fce53b92ac8> due to raw readinto() returned invalid length -1 (should have been between 0 and 8192)
WARNING:nameko.containers:killing 2 managed thread(s)
WARNING:nameko.containers:killing managed thread `run`
WARNING:nameko.containers:killing managed thread `<unknown>`
INFO:nameko.runners:stopping services: fetch_userdata_service
CRITICAL:__main__:Uncaught exception
  File "service.py", line 89, in <module>
    runner.wait()
  File "/usr/local/lib/python3.5/dist-packages/nameko/runners.py", line 115, in wait
    SpawningProxy(self.containers, abort_on_error=True).wait()
  File "/usr/local/lib/python3.5/dist-packages/nameko/utils.py", line 179, in spawning_method
    return list(fail_fast_imap(pool, call, self._items))
  File "/usr/local/lib/python3.5/dist-packages/nameko/utils.py", line 148, in fail_fast_imap
    eventlet.getcurrent().throw(*exc_info)
  File "/usr/local/lib/python3.5/dist-packages/nameko/utils.py", line 127, in handle_result
    thread_result = finished_thread.wait()
  File "/usr/local/lib/python3.5/dist-packages/eventlet/greenthread.py", line 175, in wait
    return self._exit_event.wait()
  File "/usr/local/lib/python3.5/dist-packages/eventlet/event.py", line 125, in wait
    current.throw(*self._exc)
  File "/usr/local/lib/python3.5/dist-packages/eventlet/greenthread.py", line 214, in main
    result = function(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/nameko/utils.py", line 176, in call
    return getattr(item, name)(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/nameko/containers.py", line 321, in wait
    return self._died.wait()
  File "/usr/local/lib/python3.5/dist-packages/eventlet/event.py", line 121, in wait
    return hubs.get_hub().switch()
  File "/usr/local/lib/python3.5/dist-packages/eventlet/hubs/hub.py", line 294, in switch
    return self.greenlet.switch()
  File "/usr/local/lib/python3.5/dist-packages/nameko/containers.py", line 477, in _handle_thread_exited
    gt.wait()
  File "/usr/local/lib/python3.5/dist-packages/eventlet/greenthread.py", line 175, in wait
    return self._exit_event.wait()
  File "/usr/local/lib/python3.5/dist-packages/eventlet/event.py", line 125, in wait
    current.throw(*self._exc)
  File "/usr/local/lib/python3.5/dist-packages/eventlet/greenthread.py", line 214, in main
    result = function(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/eventlet/wsgi.py", line 719, in process_request
    proto.__init__(sock, address, self)
  File "/usr/lib/python3.5/socketserver.py", line 681, in __init__
    self.handle()
  File "/usr/lib/python3.5/http/server.py", line 424, in handle
    self.handle_one_request()
  File "/usr/local/lib/python3.5/dist-packages/eventlet/wsgi.py", line 330, in handle_one_request
    self.raw_requestline = self.rfile.readline(self.server.url_length_limit)
OSError: raw readinto() returned invalid length -1 (should have been between 0 and 8192)
```

So you might think there's some bug in my code, but the service would work normally for hours, receiving requests and returning data without any problem. And suddenly it just crashed.

mattbennett:
Fixes #368 

@davidszotten the fix required a change to a test which seemed to explicitly expect a `socket.error` to take down the container (called `test_other_error`). What was the intent of that test?

### DIFF
--- a/nameko/web/server.py
+++ b/nameko/web/server.py
@@ -82,11 +82,14 @@ class WebServer(ProviderCollector, SharedExtension):
     def process_request(self, sock, address):
         try:
             self._serv.process_request((sock, address))
-        except OSError:
+        except OSError as exc:
             # OSError("raw readinto() returned invalid length")
-            # can be raised when a client disconnects very early.
+            # can be raised when a client disconnects very early as a result
+            # of an eventlet bug: https://github.com/eventlet/eventlet/pull/353
             # See https://github.com/onefinestay/nameko/issues/368
-            pass
+            if "raw readinto() returned invalid length" in str(exc):
+                return
+            raise
 
     def start(self):
         if not self._starting:
--- a/nameko/web/server.py
+++ b/nameko/web/server.py
@@ -76,8 +76,17 @@ class WebServer(ProviderCollector, SharedExtension):
             sock, addr = self._sock.accept()
             sock.settimeout(self._serv.socket_timeout)
             self.container.spawn_managed_thread(
-                partial(self._serv.process_request, (sock, addr))
+                partial(self.process_request, sock, addr)
             )
+
+    def process_request(self, sock, address):
+        try:
+            self._serv.process_request((sock, address))
+        except OSError:
+            # OSError("raw readinto() returned invalid length")
+            # can be raised when a client disconnects very early.
+            # See https://github.com/onefinestay/nameko/issues/368
+            pass
 
     def start(self):
         if not self._starting:
--- a/test/web/test_server.py
+++ b/test/web/test_server.py
@@ -40,6 +40,27 @@ def test_broken_pipe(
     assert web_session.get('/').text == ''
 
 
+def test_other_socket_error(
+    container_factory, web_config, web_config_port, web_session
+):
+    container = container_factory(ExampleService, web_config)
+    container.start()
+
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.connect(('127.0.0.1', web_config_port))
+
+    with patch.object(BaseHTTPServer.BaseHTTPRequestHandler, 'finish') as fin:
+        fin.side_effect = socket.error('boom')
+        s.sendall(b'GET / \r\n\r\n')
+        s.recv(10)
+        s.close()
+
+    # takes down container
+    with pytest.raises(socket.error) as exc:
+        container.wait()
+    assert 'boom' in str(exc)
+
+
 def test_client_disconnect_os_error(
     container_factory, web_config, web_config_port, web_session
 ):
@@ -59,30 +80,6 @@ def test_client_disconnect_os_error(
 
     # server should still work
     assert web_session.get('/').text == ''
-
-
-def test_other_error(
-    container_factory, web_config, web_config_port, web_session
-):
-    container = container_factory(ExampleService, web_config)
-    container.start()
-
-    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-    s.connect(('127.0.0.1', web_config_port))
-
-    class OtherError(Exception):
-        pass
-
-    with patch.object(BaseHTTPServer.BaseHTTPRequestHandler, 'finish') as fin:
-        fin.side_effect = OtherError('boom')
-        s.sendall(b'GET / \r\n\r\n')
-        s.recv(10)
-        s.close()
-
-    # takes down container
-    with pytest.raises(OtherError) as exc:
-        container.wait()
-    assert 'boom' in str(exc)
 
 
 @pytest.mark.parametrize(['source', 'result'], [
--- a/test/web/test_server.py
+++ b/test/web/test_server.py
@@ -40,7 +40,6 @@ def test_broken_pipe(
     assert web_session.get('/').text == ''
 
 
-@pytest.mark.xfail(reason="https://github.com/onefinestay/nameko/issues/368")
 def test_client_disconnect_os_error(
     container_factory, web_config, web_config_port, web_session
 ):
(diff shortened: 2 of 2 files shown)


test/web/test_server.py
```diff
@@ -43,6 +40,28 @@ def test_broken_pipe(
...
+    """ Regression for https://github.com/onefinestay/nameko/issues/368
+    """
+    container = container_factory(ExampleService, web_config)
+    container.start()
+
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.connect(('127.0.0.1', web_config_port))
+
+    with patch.object(HttpOnlyProtocol, 'handle_one_request') as handle:
+        handle.side_effect = OSError('raw readinto() returned invalid length')
```
mattbennett: It's difficult to demonstrate that something *doesn't* crash when the crash cause is non-deterministic.

I have something that seems to work though. Do you think the following is preferable?

``` python
import eventlet
import pytest
import requests
from eventlet.wsgi import Server

from nameko.testing.waiting import wait_for_call
from nameko.web.handlers import http


@pytest.fixture
def container(container_factory):

    class Service(object):
        name = "service"

        @http('GET', '/answer')
        def answer(self, request):
            return "42"

    container = container_factory(Service, {})
    container.start()

    return container


def test_foo(container):

    run = [True]

    def wait_for_os_error(args, kwargs, res, exc_info):
        if exc_info and exc_info[0] == OSError:
            del run[:]
            return True

    with eventlet.Timeout(1):
        with wait_for_call(Server, 'process_request', callback=wait_for_os_error):
            while run:
                gt = eventlet.spawn(requests.get, "http://localhost:8000/answer")
                eventlet.sleep(0.01)
                gt.kill()

    # still works
    assert requests.get("http://localhost:8000/answer").text == "42"

```

mattbennett: You might need to show me...

mattbennett:
For a regression test that we don't expect to live very long, I'm not sure it's that helpful

davidszotten:
the eventlet release cadence is very jerky. and even if they release a version with this fix, are we going to set that as the minimum required version for nameko?

mattbennett:
Good point. We wouldn't want to do that until the release has proved itself to be stable.

Do you have something in mind?

davidszotten:
not yet. but given the bug, i think what we need is a socket with `closed` (or `is_closed`, don't recall now) `False`, but that raises an error with one of the error codes considered `SOCKET_CLOSED` https://github.com/davidszotten/eventlet/blob/_recv_loop_socket_closed/eventlet/greenio/base.py#L71

test/web/test_server.py
davidszotten: actually, now that we understand this, and know this is (c)python complaining about a protocol violation by eventlet, i'm happy to just simulate that exact response

nameko/web/server.py
```diff
@@ -76,9 +76,21 @@ def run(self):
...
+    def process_request(self, sock, address):
+        try:
+            self._serv.process_request((sock, address))
+        except OSError as exc:
+            # OSError("raw readinto() returned invalid length")
+            # can be raised when a client disconnects very early as a result
+            # of an eventlet bug: https://github.com/eventlet/eventlet/pull/353
+            # See https://github.com/onefinestay/nameko/issues/368
+            if "raw readinto() returned invalid length" in str(exc):
+                return
```
davidszotten: no cleanup needed here?

mattbennett: I don't think so. The managed thread will exit without error and the server will continue waiting for another `accept()`

PR #371, opened 2016-10-21
restore resilience to race conditions in events integration tests

mattbennett:
Fixes the flakey tests introduced by https://github.com/onefinestay/nameko/pull/363

### DIFF
--- a/test/test_events.py
+++ b/test/test_events.py
@@ -1,9 +1,9 @@
 import itertools
+import time
 from collections import defaultdict
 
 import pytest
 from mock import Mock, create_autospec, patch
-
 from nameko.containers import WorkerContext
 from nameko.events import (
     BROADCAST, SERVICE_POOL, SINGLETON, EventDispatcher, EventHandler,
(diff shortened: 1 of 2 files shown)


test/test_events.py
```diff
@@ -332,8 +333,14 @@ def test_service_pooled_events(rabbit_manager, rabbit_config,
...
-    rabbit_manager.publish(vhost, exchange_name, 'eventtype', '"msg"',
-                           properties=dict(content_type='application/json'))
+    rabbit_manager.publish(
+        vhost, exchange_name, 'eventtype', '"msg"',
+        properties=dict(content_type='application/json')
+    )
+    # can't use the entrypoint waiter here because we don't know
+    # which of the "foo" containers will pick up the message,
+    # so just wait for events to propagate
+    time.sleep(.1)
```
davidszotten: could we make `events` a queue and `wait` on it instead of `sleep`ing?

mattbennett: It's a bad test, because we need to verify that exactly two events are received -- not one or three. The only way to do that is give them a fair chance to propagate through the system. For absolute certainty I should also be checking that all the queues are empty.

For this little irritation I'm not sure it's worth all the extra effort to refactor.

davidszotten:
👍

PR #372, opened 2016-10-25
add branch coverage as an allowed failure

mattbennett:


davidszotten:
i don't see the new job in this travis build

mattbennett:
I forgot to add the new env to the global list

davidszotten:
👍  (note that the coverage might be different for different python versions, but it's probably not worth worrying about until we get one of them sorted)

Issue #355, opened 2016-09-09
nameko shell with bpython broken since bpython==0.15


mattbennett:
https://github.com/bpython/bpython/issues/484 switches bpython embed to use curtsies, which seems to be incompatible with eventlet.

`nameko shell` doesn't actually need the eventlet monkeypatch, but it's applied because we import the docstring from `nameko.cli.run` when [setting up the argument parser](https://github.com/onefinestay/nameko/blob/master/nameko/cli/main.py#L13).

As a workaround we could continue to use the old bpython shell [as described here](https://github.com/bpython/bpython/issues/484#issuecomment-229127190)

cc @davidszotten

mattbennett:
Fixed by #376

PR #376, opened 2016-11-06
refactor cli commands to avoid eventlet.monkey_patch

### LINKED ISSUE #375: Nameko shell doesn't work when iPython is installed
With just nameko installed everything works fine, but when ipython is also installed into the same environment, `nameko shell` returns a prompt but won't accept input. I tested this against both python 2.7x and 3.5x with the same results.

davidszotten:
fixes #375


mattbennett:
Also fixes #355

PR #381, opened 2016-11-16
update travis release conf + repo links

davidszotten:


setup.py
```diff
@@ -19,7 +19,7 @@
     long_description=readme,
     author='onefinestay',
     author_email='nameko-devs@onefinestay.com',
```
mattbennett: Should probably drop this. Email isn't a required value

PR #383, opened 2016-11-21
enable heartbeat (by default) for queueconsumer

mattbennett:
Kombu does actually support _consumer_ heartbeats in more recent versions, if used with py-amqp.

This PR also removes an override that was lifted (in 2013) from Kombu for (I think) reasons of optimisation. The test suite certainly passes without it, so I think it can safely be dropped.


nameko/messaging.py
```diff
@@ -388,34 +388,6 @@ def on_consume_ready(self, connection, channel, consumers, **kwargs):
             else:
                 callback()
 
-    def consume(self, limit=None, timeout=None, safety_interval=0.1, **kwargs):
-        """ Lifted from Kombu.
-
-        We switch the order of the `break` and `self.on_iteration()` to
```
davidszotten: is this no longer an issue? how long is that timeout and would stopping have to wait for it without this?

nameko/messaging.py
```diff
@@ -331,7 +328,10 @@ def connection(self):
             return   # don't cache a connection during introspection
 
         if self._connection is None:
-            self._connection = Connection(self.amqp_uri)
+            heartbeat = self.container.config.get(
+                HEARTBEAT_CONFIG_KEY, DEFAULT_HEARTBEAT
+            )
+            self._connection = Connection(self.amqp_uri, heartbeat=heartbeat)
```
davidszotten: is there any overhead to heartbeats?

nameko/messaging.py
```diff
@@ -331,7 +328,10 @@ def connection(self):
             return   # don't cache a connection during introspection
 
         if self._connection is None:
-            self._connection = Connection(self.amqp_uri)
+            heartbeat = self.container.config.get(
```
davidszotten: are these just ignored with older (or non-forked) amqp?

mattbennett: If this works with our oldest supported kombu then I think we're fine. It's ignored for non-pyamqp transports and has been supported there since 0.9.1.

nameko/messaging.py
mattbennett: There is some extra network traffic but no performance implication as far as I'm aware. Services will probably get faster because broken connections are discovered earlier.

nameko/messaging.py
mattbennett: The default timeout is 0.1 seconds so I don't think there's anything to worry about there. If `timeout` is never set as per the comment it has no effect anyway.

@kollhof can you remember why you did this?

mattbennett:
The test suite passed for me locally (OS X, Python 3.4.3) but the failures here do seem to be real. I will investigate.

nameko/messaging.py
davidszotten: ignored (as opposed to explodes) is great

test/test_errors.py
```diff
@@ -85,10 +85,10 @@ def test_handle_result_error(container_factory, rabbit_config):
             # proxy.task() will hang forever because it generates an error
             proxy.task.call_async()
 
-        with eventlet.Timeout(1):
```
davidszotten: this was taking longer than a second?

did we make stuff slower or is this travis having issues?

mattbennett: Hard to say. I added a timer context manager around this locally and `container.wait` only ever took ~0.005 seconds.

mattbennett: But I also can't get it to fail locally...

davidszotten: hm. looking at travis, builds historically take ~3 minutes. this pr is around 9. worth checking if this is travis having issues or this pr

mattbennett: Travis is variable. Re-running master now is taking 4-5 mins: https://travis-ci.org/nameko/nameko/builds/177243811

There is some small impact here on tear-down times, but I think it is just `0.1` seconds. It becomes apparent in here because we do it a _lot_.

Do we really care that services take slightly longer to stop? I could retrofit the call to `conn.heartbeat_check()` into our lifted method, but I'd rather remove the override. We wouldn't justify _adding_ the override for a small improvement in test-run time.

mattbennett: It turns out that `safety_interval` was also changed when we lifted this code. Kombu sets it to `1`, which explains the more-than-expected slowdown.

We've added a second to `QueueConsumer.stop`, which is used in about 120 tests. This is unfortunate but I don't think it justifies keeping this lifted code. I'm also concerned about the 10x difference in timeout interval -- I'm uncomfortable being so far away from value set/expected by kombu for no apparent reason.

mattbennett:
On reflection I think the best resolution for now is to restore the `consume` override which processes `on_iteration` _before_ breaking the loop. It's unfortunate that it requires an overridde, but when the `QueueConsumer` is stopping, switching the order is not a material change.

I have changed the `safety_interval` timeout back to the kombu default though, because I think it has the potential to introduce strange networking behaviour.

I'd like to land this PR as it is so we get a) "default" timeout parameters and b) consumer heartbeats. We can assess whether there's a better way to use the `ConsumerMixin` (or replace it altogether) separately later.

nameko/messaging.py
```diff
@@ -331,10 +331,42 @@ def connection(self):
...
+                HEARTBEAT_CONFIG_KEY, DEFAULT_HEARTBEAT
+            )
+            self._connection = Connection(self.amqp_uri, heartbeat=heartbeat)
 
         return self._connection
 
+    def consume(self, limit=None, timeout=None, safety_interval=1, **kwargs):
```
davidszotten: did you move this intentionally? makes the diff harder to see

mattbennett: Sorry, not intentional. Have moved it back again

davidszotten:
i guess once the toxiproxy lands we can add integration tests for the heartbeat

test/test_errors.py
mattbennett: I don't think we should put them back. We use `pytest-timeout` to catch hanging tests now, which is what these were originally used for.

Tracking the average test execution time would be a much better way to spot performance issues. It probably wouldn't be too difficult to send stats to a free tier on something like https://plot.ly.

davidszotten: ah, forgot about pytest-timeout. though i can't see where we switch it on..

davidszotten: sounds a bit long for an individual test but we can tweak later

davidszotten:
👍 

needs changelog

PR #385, opened 2016-11-30
support listening on unix socket or file descriptor

### LINKED ISSUE #319: Pls allow passing fd to Nameko for http endpoints
Pls allow passing fd to Nameko for http endpoints, so it can be started by circus

asteven:
I need the nameko webserver to listen on a unix socket.
While at it I also added support for listening on a file descriptor (fixes #319). This is useful when running under a process manager that also handles sockets for the application e.g. circus, uwsgi.

I am not sure how to update some of the tests for this change.
Namely there are references to variables called: web_config, web_config_port, web_session in the following files:
```
nameko/testing/pytest.py
test/web/test_server.py
test/web/test_websocket.py
```

I do not understand how this works. Please advise.


davidszotten:
Hi Steven,

Thank for the patch! Will try to find time to get back to you in the next few days. Apologies for the delay. 

Best,
D

davidszotten:
`web_config` and friends are [pytest "fixtures"](http://docs.pytest.org/en/latest/fixture.html), shared test helpers. nameko ships a pytest plugin `nameko/testing/pytest.py` where some of these are defined and can be used not only by the nameko test suite, but in test suites of apps and libraries that have nameko installed

nameko/cli/commands.py
```diff
@@ -72,6 +72,10 @@ def init_parser(parser):
             help='RabbitMQ broker url')
 
         parser.add_argument(
+            '--listen',
```
davidszotten: `listen` sounds a little generic to me, maybe `web-server-listen` (that's a bit long, so open to other suggestions)

nameko/constants.py
```diff
@@ -1,5 +1,5 @@
 AMQP_URI_CONFIG_KEY = 'AMQP_URI'
-WEB_SERVER_CONFIG_KEY = 'WEB_SERVER_ADDRESS'
+WEB_SERVER_CONFIG_KEY = 'WEB_SERVER_URI'
```
davidszotten: while this may be a better key, i'm not sure it's better _enough_ to justify breaking backwards compatibility

nameko/cli/commands.py
asteven: Maybe --listen-http ? Don't care to much how this is called.

nameko/constants.py
asteven: I don't break backcompat. I just changed the default config key to the new value.
In the code that interprets this I check for the old key.

asteven:
Have long  moved on and implemented this in my own package.

PR #386, opened 2016-12-08
Faster tests with xdist

mattbennett:


mattbennett:
It turns out that xdist requires eventlet 0.20.1 on py3. I don't want to force such a recent release as our oldest supported version just to get faster tests, parking this for now.

PR #394, opened 2017-01-13
Perform the whole worker lifecycle in the worker thread

mattbennett:
There are two changes here, in two commits:

1. Move dependency injection into the container, rather than delegated to the dependency provider. 

2. Remove the use of the spawning set during `worker_setup`, `worker_result` and `worker_teardown`, so these all happen in the same (worker) thread.

The motivation for this change is https://github.com/mattbennett/nameko-sentry/pull/7 (specifically the last commit). Running the worker lifecycle methods in the same thread means that thread-locals are shared, which Raven relies upon.

The downside is that the lifecycle of an individual worker may potentially be slower -- for example, if there is more than one dependency with an onerous and I/O bound setup or teardown, they will no longer be run in parallel.

On the other hand this change avoids spawning lots of probably useless threads, since most dependencies don't implement the worker lifecycle methods.

In practice I think the performance change will be negligible.

### DIFF
--- a/nameko/containers.py
+++ b/nameko/containers.py
@@ -432,14 +432,17 @@ class ServiceContainer(object):
             setattr(worker_ctx.service, provider.attr_name, dependency)
 
     def _worker_setup(self, worker_ctx):
-        self.dependencies.all.worker_setup(worker_ctx)
+        for provider in self.dependencies:
+            provider.worker_setup(worker_ctx)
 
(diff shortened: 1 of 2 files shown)


davidszotten:
this seems fine. at some point we might want to use a sorted data structure to make the order more predictable, but let's leave that for another day

mattbennett:
Awesome, thanks. +1 for making the order more predictable too.

PR #395, opened 2017-01-15
Added config dependency provider

vmikki:


test/test_dependencies.py
```diff
@@ -0,0 +1,20 @@
...
+
+from nameko.dependencies import Config
+
+
+class TestConfig:
+
+    @pytest.yield_fixture(autouse=True)
+    def patch_container(self):
+        with patch.object(Config, 'container'):
+            yield
```
mattbennett: This is quite an "internal" looking test. For such a simple dependency provider it would be better just to demonstrate that it works.

You could simply duplicate the test you have above to verify your example. That is "real-world" usage, therefore less brittle, and would show an example service and how to use it next to each other.

vmikki:
Thank for the review! I have addressed your comments and modified the example so it more looks like a "real-life usage". I have also added myself to the contributors.

docs/examples/test/test_examples.py
```diff
@@ -281,3 +282,31 @@ def test_websocket_rpc(self, container_factory, web_config, websocket):
...
+        from examples.config_dependency_provider import Service
+
+        config = empty_config.copy()
+        config["FOO_FEATURE_ENABLED"] = True
+
+        container = container_factory(Service, config)
+        container.start()
+
+        with entrypoint_hook(container, "foo") as foo:
+            assert "foo" == foo()
```
davidszotten: we normally avoid [yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions) in python (here and below)

mattbennett: I didn't know this had a name!

vmikki: I came from PHP, it's good to know that it should be avoided in python, thanks!

davidszotten: it's not harmful, but it reads a bit odd and doesn't add anything, since python uses different operators for comparison and assignment, and e.g. `if a = 2:` is a syntax error

davidszotten:
may be worth a note that this is _read only_ access to the config

mattbennett:
Good shout. I'll add this in the RC branch

Issue #374, opened 2016-11-01
Setting and getting message properties manually


sleighsoft:
Is it possible to set and retrieve message properties in pub-sub manually?

gitricko:
+1

mattbennett:
This is a pretty regularly requested feature. It's been discussed in https://github.com/nameko/nameko/issues/292 previously.

#337 improves the situation slightly (by standardising the AMQP implementations to some extend) and I have some work in progress on top of that which exposes more message and delivery options. So watch this space.

mattbennett:
There is a PR for this now: #456

mattbennett:
Added in #456

PR #400, opened 2017-01-28
🚧  Expose delivery options

mattbennett:
🚧  work in progress 🚧 

This exposes delivery options and message properties to publishers, as requested in https://github.com/nameko/nameko/issues/374. It is backwards compatible.

https://github.com/nameko/nameko/issues/292 is related, though the examples there relate to RPC, which is not tackled here. As a next step I would like to split `Publisher.publish` out into a standalone class, and refactor all the remaining message publishers (RPC proxy and responder, and standalone RPC proxy and event dispatcher) to use it. That would make for cleaner APIs and make #292 easier to implement. I've raised this PR as an interim to help with the review.


mattbennett:
Having started on the "next step" mentioned in the comment above, it might make sense to hold off the review here, other than a glance to get the feel of it.

mattbennett:
Closing in favour of #456

PR #401, opened 2017-01-29
Push coverage to coveralls

mattbennett:


davidszotten:
don't have a coveralls account, but i think there's a tickbox to disable these comments (to reduce the notification spam) could you try to disable please?

mattbennett:
I'm not convinced that coveralls adds much. I may try codecov.io instead eventually but will close this for now.

PR #405, opened 2017-02-06
standalone rpc: non blocking queue consumer

scarchik:
Following up on the [discussion](https://groups.google.com/forum/#!topic/nameko-dev/hGpYx7ka1lI), I prepare change with non-blocking standalone rpc queue consumer. It should help communicate other software with nameko (like: flask) and dealing with connections. 

**Why I need that?** : I want use flask generated REST API with swagger codegen. Flask should provide API gateway to access to microservices on nameko. Flask provide better tools set for REST API creation.

**How it use?** : MultiReplyListener provide non-blocking wait for replies from rabbitmq (like: ReplyListener)
```
from nameko.standalone.rpc import ClusterRpcProxy, MultiReplyListener

with ClusterRpcProxy(rabbit_config, reply_listener_cls=MultiReplyListener) as proxy:
    proxy.foobar.spam()
```

Thanks for the any feedback!

### DIFF
--- a/nameko/standalone/rpc.py
+++ b/nameko/standalone/rpc.py
@@ -207,14 +207,16 @@ class StandaloneProxyBase(object):
 
     def __init__(
         self, config, context_data=None, timeout=None,
-        worker_ctx_cls=WorkerContext, reply_listener_cls=SingleThreadedReplyListener
+        worker_ctx_cls=WorkerContext,
+        reply_listener_cls=SingleThreadedReplyListener
     ):
         container = self.ServiceContainer(config)
 
         self._worker_ctx = worker_ctx_cls(
             container, service=None, entrypoint=self.Dummy,
             data=context_data)
-        self._reply_listener = reply_listener_cls(timeout=timeout).bind(container)
+        self._reply_listener = reply_listener_cls(
+            timeout=timeout).bind(container)
 
     def __enter__(self):
         return self.start()
--- a/nameko/standalone/rpc.py
+++ b/nameko/standalone/rpc.py
@@ -3,10 +3,13 @@ from __future__ import absolute_import
 import logging
 import socket
 
+import eventlet
+from eventlet import event
 from amqp.exceptions import ConnectionError
 from kombu import Connection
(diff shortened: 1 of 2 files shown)


mattbennett:
Thanks for raising this.

One of the requirements for the "standalone" features is that they should be usable without having to run the eventlet event loop -- so they can be used by clients with other concurrency models. Unfortunately this implementation needs eventlet.

What does this give you over https://github.com/clef/flask-nameko? I guess there's a guarantee of exactly one connection, rather than using the pool?

scarchik:
I wanted to use flask-nameko, but after testing I found out that it not what I need. It create too many connections, not much better than use ClusterRpcProxy standalone. 
In system with large number of users make a connection for each call is wasteful. 

With current change I can use one ClusterRpcProxy which have 1 connect to RabbitMQ to send messages and many lazy connections (they created by kombu) for waiting replies. Advantages best be seen on long polling queries.

mattbennett:
@scarchik, sorry to leave this PR hanging.

I understand the advantage of using this over flask-nameko now. As I mentioned, the "standalone" package is for things that will work with a vanilla interpreter, so I don't want to add the `MultiQueueConsumer` there.

Do you want to raise a smaller PR that just allows the `reply_listener_cls` to be specified, so you can use this in your own application without also having to subclass `StandaloneProxyBase`?

scarchik:
Ok, I already change this request as you say, it be useful.

Thank you for your comments.
PS: Does it make sense to create pypi package (like: nameko_proxy) with this code or it not useful for other?

mattbennett:
Great, thank you. Your branch is failing the linter check at the moment with lines >79 chars.

I think it could be useful as a PyPI package, for people who're using eventlet-based clients to call nameko services.

scarchik:
Hi, @mattbennett,
Looks like this change had been accidentally deleted in https://github.com/nameko/nameko/commit/1e2e3298a5e6ca05cd8cef3ae89ef4ef6f7c0c82.

Can we submit it again?

PR #406, opened 2017-02-10
Fixes for testing against non-local rabbit; and always using clean vhost

mattbennett:
It turns out that we had quite a few tests that assume RabbitMQ is running on localhost and port 5672. This PR fixes those tests (although it doesn't update the Travis config to prove it).

It also turns out that many test assume they're running against a clean vhost and fail otherwise. Accordingly I have removed the `:random:` vhost option and always create a fresh one. Since the vhost is part of a normal URI I've replaced this pytest option into four new options that allow the other parts of the URI to be overridden: host, port, username and password.

### DIFF
--- a/docs/examples/test/test_examples.py
+++ b/docs/examples/test/test_examples.py
@@ -286,22 +286,22 @@ class TestWebsocketRpc(object):
 
 class TestConfig:
 
-    def test_config_value_not_set(self, container_factory, empty_config):
+    def test_config_value_not_set(self, container_factory, rabbit_config):
         from examples.config_dependency_provider import (
             Service, FeatureNotEnabled
         )
 
-        container = container_factory(Service, empty_config)
+        container = container_factory(Service, rabbit_config)
         container.start()
 
         with pytest.raises(FeatureNotEnabled):
             with entrypoint_hook(container, "foo") as foo:
                 foo()
 
-    def test_can_get_config_value(self, container_factory, empty_config):
+    def test_can_get_config_value(self, container_factory, rabbit_config):
         from examples.config_dependency_provider import Service
 
-        config = empty_config.copy()
+        config = rabbit_config
         config["FOO_FEATURE_ENABLED"] = True
 
         container = container_factory(Service, config)
@@ -309,4 +309,3 @@ class TestConfig:
 
         with entrypoint_hook(container, "foo") as foo:
             assert foo() == "foo"
-
--- a/docs/examples/test/test_examples.py
+++ b/docs/examples/test/test_examples.py
@@ -2,7 +2,6 @@
 """ Tests for the files and snippets in nameko/docs/examples
 """
 import os
-import pytest
 
 from mock import call, patch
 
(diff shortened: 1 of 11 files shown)


nameko/testing/pytest.py
```diff
@@ -23,14 +23,29 @@ def pytest_addoption(parser):
         help=("The logging-level for the test run."))
 
     parser.addoption(
-        "--amqp-uri", action="store", dest='AMQP_URI',
-        default='amqp://guest:guest@localhost:5672/:random:',
-        help=("The AMQP-URI to connect to rabbit with."))
+        "--rabbit-host", action="store", dest='RABBIT_HOST',
```
davidszotten: how come you're exploding this into separate settings?

nameko/testing/pytest.py
```diff
@@ -23,14 +23,29 @@ def pytest_addoption(parser):
...
+        default='guest',
+        help=("RabbitMQ username."))
+
+    parser.addoption(
+        "--rabbit-pass", action="store", dest='RABBIT_PASS',
+        default='guest',
+        help=("RabbitMQ password."))
+
+    parser.addoption(
+        "--rabbit-mgmt-uri", action="store", dest='RABBIT_MGMT_URI',
```
davidszotten: don't love abbreviations, maybe `rabbit-api-uri`?

mattbennett: 👍  much better

nameko/testing/pytest.py
mattbennett: Because it's not possible to specify an `AMQP_URI` _without_ a vhost. Even `amqp://localhost` implies the default vhost.

davidszotten: uri's are still a convenient way to specify driver (protocol), host, port and credentials in a single string. if we no longer allow the vhost to be specified i guess we can ignore it in the uri?

(even if we change the default, do we not even want to let users override the vhost?)

mattbennett: OK cool. Using a URI but ignoring the vhost sounds good to me.

Letting users specify their own really isn't helpful, because so many tests assume they're being run in a clean vhost.

docs/examples/test/test_examples.py
```diff
@@ -286,27 +286,26 @@ def test_websocket_rpc(self, container_factory, web_config, websocket):
 
 class TestConfig:
 
-    def test_config_value_not_set(self, container_factory, empty_config):
+    def test_config_value_not_set(self, container_factory, rabbit_config):
```
davidszotten: how come these are now using `rabbit_config`?

test/cli/test_run.py
```diff
@@ -46,12 +46,20 @@ def test_run(rabbit_config):
...
+        AMQP_URI: '{}'
+        serializer: 'json'
+    """.format(rabbit_config[AMQP_URI_CONFIG_KEY]))
+
     parser = setup_parser()
     args = parser.parse_args([
         'run',
         '--config',
-        TEST_CONFIG_FILE,
+        config.strpath,
```
davidszotten: i _think_ the public api is `str(config)`

test/cli/test_shell.py
```diff
@@ -45,9 +42,11 @@ def fake_alternative_interpreters():
         yield
 
 
-def test_basic(pystartup):
+def test_basic(pystartup, rabbit_config):
```
davidszotten: how come these all need to test the broker?

docs/examples/test/test_examples.py
mattbennett: The service that it starts has an `@rpc` entrypoint.

The old `empty_config` wasn't really empty -- it had an `AMQP_URI` with an empty value, which was translated into `amqp://localhost`. The test failed if you ran it without having a broker running locally.

test/cli/test_run.py
mattbennett: Hmm. The docs are ambiguous. There is an example of using it that way, but it's not explicit.

I think `.strpath` shows the intent more clearly.

test/cli/test_shell.py
mattbennett: The `n` global creates a `ReplyListener` that connects to rabbit. These tests also used to assume there was a broker running at `amqp://localhost`

davidszotten: ah yes. :(

docs/examples/test/test_examples.py
davidszotten: hm. ok. (given how relative expensive it is to spin use the rabbit dep, maybe we can make those examples not need it)

nameko/amqp.py
```diff
@@ -43,7 +43,7 @@ class TestTransport(Transport):
 
 def verify_amqp_uri(amqp_uri):
     connection = Connection(amqp_uri)
-    if connection.transport_cls != 'amqp':
+    if connection.transport_cls not in ('amqp', 'pyamqp'):
```
davidszotten: what's left?

mattbennett: What do you mean?

I guess you could call it with `librabbitmq://`, but that would break in all kinds of other ways because it doesn't work with eventlet. What was the original purpose of this check?

davidszotten: sorry, read this the wrong way around. you are _adding_ another transport which _can_ use the `TestTransport`

PR #409, opened 2017-02-12
Faster teardown

mattbennett:
Significantly speeds up test teardown by:

* Removing the leftover connection check (9371f2b) which is not useful when every test runs in its own vhost (it was added to prevent state lying around which may cause flakiness in later tests)
* Forcing the RabbitMQ fixture to stop _before_ stopping containers (3610e39), which breaks the consume loop early and lets us clean up sooner (see fixture docstring).

This change also allows to remove the kombu method override (2fea493) which was made to speed up teardown of the QueueConsumer. We now force this loop to exit early (during tests) so it doesn't matter whether `on_iteration` happens before or after the socket read.

As quick benchmark I ran the following test:

``` python
@pytest.mark.parametrize("iter", range(50))
def test_speedup(iter, rabbit_config, container_factory):

    class Service(object):
        name = "serv"

        @rpc
        def service(self):
            pass

    # do nothing except start and kill a container that connects to rabbit
    container = container_factory(Service, rabbit_config)
    container.start()
```

Results:
```
# master
50 passed in 87.97 seconds

# faster-teardown
50 passed in 36.62 seconds
```

### DIFF
--- a/nameko/testing/pytest.py
+++ b/nameko/testing/pytest.py
@@ -23,29 +23,22 @@ def pytest_addoption(parser):
         help=("The logging-level for the test run."))
 
     parser.addoption(
-        "--rabbit-host", action="store", dest='RABBIT_HOST',
-        default='localhost',
-        help=("Hostname of the RabbitMQ broker."))
+        "--amqp-uri", "--rabbit-amqp-uri",
+        action="store",
+        dest='RABBIT_AMQP_URI',
+        default='pyamqp://guest:guest@localhost:5672/',
+        help=(
+            "URI for the RabbitMQ broker. Any specified virtual host will be "
+            "ignored because tests run in their own isolated vhost."
+        ))
 
     parser.addoption(
-        "--rabbit-port", action="store", dest='RABBIT_PORT',
-        default='5672',
-        help=("AMQP port number on RabbitMQ broker."))
-
-    parser.addoption(
-        "--rabbit-user", action="store", dest='RABBIT_USER',
-        default='guest',
-        help=("RabbitMQ username."))
-
-    parser.addoption(
-        "--rabbit-pass", action="store", dest='RABBIT_PASS',
-        default='guest',
-        help=("RabbitMQ password."))
-
-    parser.addoption(
-        "--rabbit-api-uri", action="store", dest='RABBIT_API_URI',
+        "--rabbit-api-uri", "--rabbit-ctl-uri",
+        action="store",
+        dest='RABBIT_API_URI',
         default='http://guest:guest@localhost:15672',
-        help=("URI for RabbitMQ management interface."))
+        help=("URI for RabbitMQ management interface.")
+    )
 
 
 def pytest_load_initial_conftests():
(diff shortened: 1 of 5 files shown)


nameko/testing/pytest.py
```diff
@@ -125,59 +121,90 @@ def rabbit_config(request, rabbit_manager):
...
+    defined by the test, for example::
+
+    def test_foo(container_factory, rabbit_config):
+        pass  # rabbit_config tears down first
+
+    def test_bar(rabbit_config, container_factory):
+        pass  # container_factory tears down first
+
+    This fixture ensures the teardown order is:
+
```
davidszotten: what ensures (1)?

test/test_service_runner.py
```diff
@@ -35,16 +30,21 @@ def warnings():
         yield patched
 
 
-class Service(object):
-    name = "service"
+@pytest.fixture
```
davidszotten: how come this is moving into a fixture?

test/testing/test_pytest_plugin.py
```diff
@@ -40,60 +40,145 @@ def test_rabbit_manager(rabbit_manager):
...
-            name = "service"
-
-            @rpc
-            def method(self):
+class TestFastTeardown(object):
+
+    def test_order(self, testdir):
+
+        testdir.makeconftest(
+            """
```
davidszotten: we might want to think about "hosting" these tests in actual files (maybe outside the main test tree) so they don't have to live inside docstrings

nameko/testing/pytest.py
```diff
@@ -125,59 +121,90 @@ def rabbit_config(request, rabbit_manager):
...
+    That is, `rabbit_config` teardown, which removes the vhost created for
+    the test, happens *before* the consumers are stopped.
+
+    Deleting the vhost causes the broker to sends a "basic-cancel" message
+    to any connected consumers, which will include the consumers in all
+    containers created by the `container_factory` and `runner_factory`
+    fixtures.
+
+    This speeds up test teardown because the "basic-cancel" breaks
```
davidszotten: could we use the same trick to speed up teardown of actual (non-test) services by detecting when all workers are done and deleting the reply queue?

test/test_service_runner.py
mattbennett: So that it can have the `tracker` in scope, which is a fixture.

I originally changed it because I thought the `reset_mock` was responsible for the dodgy test, and only later discovered the duplicate event dispatch.

The "tracker" mock is the pattern we've used elsewhere to track calls like this though, so I left it.

test/testing/test_pytest_plugin.py
mattbennett: I think I prefer them like this. Everything is in the same place, so you can't accidentally let things get out of sync or used for more than one purpose.

davidszotten: ok. just shame about editor support (syntax etc)

nameko/testing/pytest.py
mattbennett: I think it's because of the autouse and the fact that the "reorder fixtures" are all in the same plugin module. I don't think it's an official behaviour that's guaranteed to work in future versions but it does for now.

It _might_ be possible to guarantee it with some deeper pytest magic (I'm following https://github.com/pytest-dev/pytest/issues/1216), but I think this is OK for now given that we pin pytest.

nameko/testing/pytest.py
mattbennett: I guess so, although I'm not sure that speeding up teardown of real services adds much value. You'd save at-most one second per container, which is only really significant during a test run that starts and stops hundreds of containers in sequence.

davidszotten: i think a second is a long time to wait for nothing at all but happy to leave for now

nameko/testing/pytest.py
davidszotten: i think there is (or used to be) a `tryfirst` marker but probably be ok to leave as is for now

mattbennett: The docs for `tryfirst` refer to marking a "hook implementation". I tried applying it to fixtures and couldn't make it have any affect.

nameko/testing/pytest.py
mattbennett: My preferred solution for that is to consume messages with something other than kombu, which is getting closer to being possible.

test/testing/test_pytest_plugin.py
mattbennett: Yeah that is slightly annoying, but I think I still prefer it this way.

nameko/testing/pytest.py
davidszotten: agreed

PR #410, opened 2017-02-13
Pipeline vhosts

mattbennett:


davidszotten:
could you rebase this on top of master please?

nameko/testing/utils.py
```diff
@@ -140,3 +140,65 @@ def find_free_port(host='127.0.0.1'):
     port = sock.getsockname()[1]
     sock.close()
     return port
+
+
+class ResourcePipeline(object):
```
davidszotten: do we see this being re-used? (if not, maybe move to conftest?)

nameko/testing/utils.py
```diff
@@ -140,3 +140,65 @@ def find_free_port(host='127.0.0.1'):
...
+        while True:
+            obj = self.trash.get()
+            if obj is ResourcePipeline.STOP:
+                break
+            self.destroy(obj)
+
+    def get(self):
+        return self.ready.get()
+
+    def discard(self, vhost):
```
davidszotten: if we're keeping this generic, maybe vhost -> resource or similar

nameko/testing/utils.py
```diff
@@ -140,3 +140,65 @@ def find_free_port(host='127.0.0.1'):
...
+        self.trash.put(vhost)
+
+    def shutdown(self):
+        self.running = False
+        while self.ready.qsize():
+            unused = self.ready.get()
+            self.trash.put(unused)
+            eventlet.sleep()  # allow create thread to run
```
davidszotten: how come we need this? (is it not enough to wait in `wait` below?)

nameko/testing/pytest.py
```diff
@@ -94,21 +94,45 @@ def rabbit_manager(request):
...
     rabbit_amqp_uri = request.config.getoption('RABBIT_AMQP_URI')
     uri_parts = urlparse(rabbit_amqp_uri)
     username = uri_parts.username
 
-    vhost = "nameko_test_{}".format(
-        "".join(random.choice(string.ascii_lowercase) for _ in range(10))
-    )
-    rabbit_manager.create_vhost(vhost)
-    rabbit_manager.set_vhost_permissions(vhost, username, '.*', '.*', '.*')
+    vhost = vhost_pipeline.get()
```
davidszotten: thought on using a contextmanager interface for `get` and `discard` as well?

nameko/testing/utils.py
```diff
@@ -140,3 +140,65 @@ def find_free_port(host='127.0.0.1'):
...
+    """ Creates and destroys resources in background threads.
+
+    Creates up to `size` resources ahead of time so the caller avoids waiting
+    for lazy creation.
+    """
+
+    STOP = object()
+
+    def __init__(self, create, destroy, size=3):
+        if size == 0:
```
davidszotten: how did you pick the size?

nameko/testing/pytest.py
mattbennett: Ah, nice idea

nameko/testing/utils.py
mattbennett: I would like to use it in nameko-sqlalchemy to speed up creation of database schemas.

nameko/testing/utils.py
mattbennett: `size=3`? From trial and error in the vhost usage -- 3 resulted in the best speedup. It's a balance between having something available when you need it and the "waste" of destroying what's left unused at the end of the run.

nameko/testing/utils.py
mattbennett: Yep, thanks

nameko/testing/utils.py
mattbennett: It's for the case where the create loop is blocked waiting to put an item into the `ready` queue (i.e. waiting for it to be less than maxsize).

`unused = self.ready.get()` unblocks the create thread but doesn't yield, so we exit the while loop here before giving it a chance to run. Once unblocked, the create thread deposits its item into the ready queue, and if we've already exited the loop, that final item isn't destroyed.

I've expanded the inline comment to hopefully explain

nameko/testing/utils.py
davidszotten: so do you think 3 is still a reasonable default for other use-cases? (or do we e.g. want to force the user to choose?)

nameko/testing/utils.py
```diff
@@ -140,3 +140,70 @@ def find_free_port(host='127.0.0.1'):
...
+            self.trash.put(item)
+
+    def shutdown(self):
+        self.running = False
+        while self.ready.qsize():
+            unused = self.ready.get()
+            # allow create thread to run if it is blocked waiting for capacity
+            # in the ready queue. otherwise we'll exit this loop before the
+            # final item is added to the ready queue and it won't be destroyed
+            eventlet.sleep()
```
davidszotten: i'm still thinking about this but we might have a race here (e.g. if `create` is slow)

nameko/testing/utils.py
mattbennett: I prefer having a default rather than forcing the user to select something. I think 3 is fine.

nameko/testing/utils.py
mattbennett: I think you're right. I'll rework it.

Issue #417, opened 2017-03-14
Performance: one service worker can only process one task per second? 


Ryannnnnnn:
I have a ServiceContainer with only one worker, as below:

```Python
config = {
        AMQP_URI_CONFIG_KEY:AMQP_URL,
        MAX_WORKERS_CONFIG_KEY: 1,
}
ServiceContainer(service_class, config)
```

After sending async calls, I started the service as the consumer. But it turns out that the only one worker can do exactly one task per second, is there something wrong with my configuration?

The service simply returns a number:

```Python
class Consumer(object):
    name = 'test'
    @rpc
    def work(self):
        logging.info('hello')
        return 42
```

logs:

```
[I 170314 20:39:25 main:26] hello
[I 170314 20:39:26 main:26] hello
[I 170314 20:39:27 main:26] hello
[I 170314 20:39:28 main:26] hello
[I 170314 20:39:29 main:26] hello
[I 170314 20:39:30 main:26] hello
[I 170314 20:39:31 main:26] hello
[I 170314 20:39:32 main:26] hello
[I 170314 20:39:33 main:26] hello
[I 170314 20:39:34 main:26] hello
[I 170314 20:39:35 main:26] hello
```

mattbennett:
You're doing something wrong, but I can't tell what from what you've posted.

How are you running the service? And how are you calling it?

Ryannnnnnn:
I think how I call the service is irrelevant, since I push hundreds of async RPC calls into queue all at once and just to see how service works.

The way I run the service is almost the same as `nameko run` (a.k.a. `nameko.cli.run.run()`) except that I call it manually (to rewrite the default `max_workers`). I posted the full demo [here](https://gist.github.com/Ryannnnnnn/ef898d155b824d29e7d710dc01da2bff).

davidszotten:
i think there's a performance regression between `v2.4.3` and `v2.4.4`. even with 10 workers, there's a gap between each chunk of 10 being handled. still looking

mikegreen7892003:
I work with @Ryannnnnnn .

There is a dead lock between `msg.ack` and `drain_events` while `container.max_workers` equals or large than `prefetch_count`.

e.g.

We set `max_workers = 1`, so `QueueConsumer.prefetch_count = 1`  and `consumer.qos(prefetch_count=self.prefetch_count)`. 

RabbitMQ with the `prefetch_count=1` will not delivery new message while the old one is not acked.

[...]

PR #419, opened 2017-03-16
Bump prefetch

davidszotten:


test/test_rpc.py
```diff
@@ -1019,3 +1020,26 @@ def enable_after_retry(args, kwargs, res, exc_info):
         # call 2 succeeds (after reconnecting via retry policy)
         with patch_wait(Connection, 'connect', callback=enable_after_retry):
             assert service_rpc.echo(2) == 2
+
+
+def test_prefetch_throughput(container_factory, rabbit_config):
+    """Make sure even max_workers=1 can consumer faster than 1 msg/second"""
```
mattbennett: Maybe add a note that this is a regression test for #417?

PR #421, opened 2017-03-19
Env var substitution with yaml types

Trex-Boolat:
Using http://nameko.readthedocs.io/en/stable/cli.html#environment-variable-substitution - always receive string in nameco config. Via re-throw PyYaml  resolver - receive int/bool/float in config.

Based on: https://groups.google.com/forum/#!topic/nameko-dev/oC2JtEflLtM

### DIFF
--- a/nameko/cli/main.py
+++ b/nameko/cli/main.py
@@ -45,9 +45,10 @@ def _replace_env_var(match):
 
 
 def _env_var_constructor(loader, node):
-    value = loader.construct_scalar(node)
-    return ENV_VAR_MATCHER.sub(_replace_env_var, value)
-
+    raw_value = loader.construct_scalar(node)
+    value = ENV_VAR_MATCHER.sub(_replace_env_var, raw_value)
+    new_tag = loader.resolve(yaml.ScalarNode, value, (True, False))
+    return loader.yaml_constructors[new_tag](loader, yaml.ScalarNode(new_tag, value))
 
 def setup_yaml_parser():
     yaml.add_constructor('!env_var', _env_var_constructor)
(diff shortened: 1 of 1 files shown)


nameko/cli/main.py
```diff
@@ -45,8 +45,11 @@ def _replace_env_var(match):
 
 
 def _env_var_constructor(loader, node):
-    value = loader.construct_scalar(node)
-    return ENV_VAR_MATCHER.sub(_replace_env_var, value)
+    raw_value = loader.construct_scalar(node)
+    value = ENV_VAR_MATCHER.sub(_replace_env_var, raw_value)
+    new_tag = loader.resolve(yaml.ScalarNode, value, (True, False))
```
davidszotten: what's the `(True, False)` tuple? could we give it a name to make that clearer?

Trex-Boolat: Changed...

mattbennett: The name doesn't make anything clearer to me. But then again, neither does digging through the pyyaml source (http://pyyaml.org/browser/pyyaml/trunk/lib3/yaml/resolver.py#L140)

@Trex-Boolat can you explain what that third argument does?

davidszotten:
please see @mattbennett's comment above (you touched the code so it's considered "outdated" by github) but it's still relevant

nameko/cli/main.py
Trex-Boolat: This (True, False) magic - only way to say pyyaml: try resolve node by content and do not return **string** in all cases. Via hinter from **yaml.add_implicit_resolver('!env_var', IMPLICIT_ENV_VAR_MATCHER)** any yaml node with ${...} routed to _env_var_constructor() and avoid pyyaml resolver. So when u use yaml
`
a: True
b: ${env_b}
`
without this fix - always boolean **a** and string **b**, even if env_b == "True"

PR #422, opened 2017-03-19
Add explicite logger for @http decorator

Trex-Boolat:
eventlet.wsgi.Server in "nameko\web\server.py" leave log_output=True and log=None, so all access log like
127.0.0.1 - * [18/Mar/2017 00:57:12] "GET /health HTTP/1.1" 200 140 0.001000
127.0.0.1 - * [18/Mar/2017 00:57:22] "GET /health HTTP/1.1" 200 140 0.002001
127.0.0.1 - * [18/Mar/2017 00:57:23] "GET /health HTTP/1.1" 200 140 0.001008
fall to stderr...

Inside docker and Kubernetes constant flow of non-formatted/unstoppable/non-configurable lines from health checks - big problem.

Explicit configure logger allow receive logs like
`{"asctime": "2017-03-18 01:05:13,662", "msecs": 662.9447937011719, "levelname": "INFO", "process": 9304, "module": "wsgi", "lineno": 566, "args": [], "msg": "127.0.0.1 - * [18/Mar/2017 01:05:13] \"GET /health HTTP/1.1\" 200 140 0.001000", "request_id": null}`

And full control via nameko config
```
...
LOGGING:
  version: 1
...
  loggers:
    wsgi:
      level: INFO
      handlers: [console]
      propagate: False
```

Based on: https://groups.google.com/forum/#!topic/nameko-dev/HN0MP1PjyT4

### DIFF
--- a/nameko/web/server.py
+++ b/nameko/web/server.py
@@ -117,7 +117,7 @@ class WebServer(ProviderCollector, SharedExtension):
             wsgi_app,
             protocol=protocol,
             debug=debug,
-            log=getLogger('wsgi')
+            log=getLogger(__name__)
         )
 
     def stop(self):
--- a/nameko/web/server.py
+++ b/nameko/web/server.py
@@ -7,6 +7,7 @@ import eventlet
 from eventlet import wsgi
 from eventlet.support import get_errno
 from eventlet.wsgi import BROKEN_SOCK, BaseHTTPServer, HttpProtocol
+from logging import getLogger
 from werkzeug.exceptions import HTTPException
 from werkzeug.routing import Map
 from werkzeug.wrappers import Request
@@ -115,7 +116,8 @@ class WebServer(ProviderCollector, SharedExtension):
             sock.getsockname(),
             wsgi_app,
             protocol=protocol,
-            debug=debug
+            debug=debug,
+            log=getLogger('wsgi')
         )
 
     def stop(self):
--- a/tox.ini
+++ b/tox.ini
@@ -5,7 +5,7 @@ skipsdist = True
 [testenv]
 deps =
     # oldest supported libraries for each python
-    py27-oldest: eventlet==0.17.2
+    py27-oldest: eventlet==0.17.4
     py{33,34}-oldest: eventlet==0.17.4
     py27-oldest: kombu==3.0.30
     py{33,34}-oldest: kombu==3.0.15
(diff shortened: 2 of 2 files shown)


nameko/web/server.py
```diff
@@ -115,7 +116,8 @@ def get_wsgi_server(
             sock.getsockname(),
             wsgi_app,
             protocol=protocol,
-            debug=debug
+            debug=debug,
+            log=getLogger('wsgi')
```
davidszotten: this is a pretty generic name. the most common convention i know is to namespace by module (`getLogger(__name__)`). if you want further customisability you would typically add further namespacing below with another dot

Trex-Boolat: Ok. Changed.

davidszotten:
re: the failing tests: i guess the clue was that it was failing the build that uses our oldest supported dependency versions. in this case, the log behaviour of eventlet's wsgi server that you are making use of was introduced in [0.17.2](https://github.com/eventlet/eventlet/blob/master/NEWS#L130) ([pr](https://github.com/eventlet/eventlet/commit/f2aaa0dda908c1b1a8c75d25d0a781fbfeae7f9f)

the currently oldest supported version of eventlet is 0.16.1

not sure how that was chosen, but maybe we could consider bumping our minimum requirement

Trex-Boolat:
for eventlet's 0.17.2 - i must update setup.py inside my pull request? or you make all work in master?

mattbennett:
Sorry to leave this without a reply for so long.

0.17.2 is two years old, and was actually only released 3 months after 0.16.1. So I'm happy to drop support for 0.16. @davidszotten do you agree? If so, yes please change the requirement in this PR @Trex-Boolat  (it's actually in [tox.ini](https://github.com/nameko/nameko/blob/master/tox.ini#L8) and only needs changing for python 2.7)

davidszotten:
yes that sounds fine

davidszotten:
it seems eventlet removes point releases from pypi but `0.17.4` is only another few months behind `0.17.2` so also fine

Trex-Boolat:
checks done :) https://travis-ci.org/nameko/nameko/builds/218489677

Trex-Boolat:
When i see release with PR's 421+422?

mattbennett:
I'll cut a new release as soon as possible.

PR #426, opened 2017-04-10
Env var substitution with yaml types

kooba:
I've added few tests on top of @Trex-Boolat PR (https://github.com/nameko/nameko/pull/421)
They validate/demonstrate resolver is working as expected. 

Additionally looks like `loader.resolve` PyYAML's API is bit weird and requires unused value to be passed, so added a comment why `None` is passed instead.

This change is not strictly speaking backwards compatible so bumping version to `2.6.0`

### DIFF
--- a/nameko/cli/main.py
+++ b/nameko/cli/main.py
@@ -47,8 +47,11 @@ def _replace_env_var(match):
 def _env_var_constructor(loader, node):
     raw_value = loader.construct_scalar(node)
     value = ENV_VAR_MATCHER.sub(_replace_env_var, raw_value)
-    yaml_implicit_resolver = (True, False)
-    new_tag = loader.resolve(yaml.ScalarNode, value, yaml_implicit_resolver)
+    use_implicit_scalar_resolver = True
+    # PyYAML requires tuple/list value for `implicit` arg in `resolve` method
+    # containing two items. Second one is not used so passing `None` here.
+    new_tag = loader.resolve(
+        yaml.ScalarNode, value, (use_implicit_scalar_resolver, None))
     new_node = yaml.ScalarNode(new_tag, value)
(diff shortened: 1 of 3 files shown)


nameko/cli/main.py
```diff
@@ -45,8 +45,15 @@ def _replace_env_var(match):
...
 def _env_var_constructor(loader, node):
-    value = loader.construct_scalar(node)
-    return ENV_VAR_MATCHER.sub(_replace_env_var, value)
+    raw_value = loader.construct_scalar(node)
+    value = ENV_VAR_MATCHER.sub(_replace_env_var, raw_value)
+    use_implicit_scalar_resolver = True
+    # PyYAML requires tuple/list value for `implicit` arg in `resolve` method
+    # containing two items. Second one is not used so passing `None` here.
+    new_tag = loader.resolve(
+        yaml.ScalarNode, value, (use_implicit_scalar_resolver, None))
```
mattbennett: This makes sense to me now, thanks 👍

test/cli/test_main.py
```diff
@@ -86,7 +86,43 @@ class TestConfigEnvironmentVariables(object):
...
+                - ${INT_2}
+                - ${INT_3}
+            BAR:
+                - 1
+                - 2
+                - 3
+            """,
+            {'INT_1': '1', 'INT_2': '2', 'INT_3': '3'},
+            {'FOO': [1, 2, 3], 'BAR': [1, 2, 3]}
+        )
```
mattbennett: Great to have these extra tests

mattbennett:
FYI I plan to release this a 2.6.0 once the hotfixes in 2.5.4 are out, so the 2.5 series ends in a stable place.

PR #427, opened 2017-04-11
Enable method dispatch.

zhouxiaoxiang:
# When call a remote service, method dispatch is more readable and convient.

Now, write code as follows:
 
```
from nameko.standalone.rpc import ClusterRpcProxy
with ClusterRpcProxy(config) as services:
        services["some_service"]["some_method"]("some_args")
```

### DIFF
--- a/nameko/rpc.py
+++ b/nameko/rpc.py
@@ -319,6 +319,9 @@ class ServiceProxy(object):
         return MethodProxy(
             self.worker_ctx, self.service_name, name, self.reply_listener)
 
+    def __getitem__(self, name):
+        return getattr(self, name)
+
 
 class RpcReply(object):
     resp_body = None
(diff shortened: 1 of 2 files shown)


mattbennett:
Thanks @zhouxiaoxiang for the contribution!

We added dict-access for services because they can contain characters that can't be used in python identifiers (e.g. `"my-service"`). That isn't case for method names because they also need to exist as methods on the service class.

Why do you find this approach more convenient?

zhouxiaoxiang:
I have written a project framework for my project, which can discover and register all services for **my app**. 
　　Now code looks like:
```
def call_service(service, method, **kwargs):
    method = getattr(service, method)
    ......
    return method(kwargs)
```
　　But, I prefer so:
```
def call_service(service, method, **kwargs):
    return service[method](kwargs)
```
　　~~There's more than one way to do it.~~

Issue #428, opened 2017-04-11
Deadlock and AMQP reconnections when prefetched messages exceed max workers


mattbennett:
The interaction of the AMQP `prefetch_count` and Nameko's worker pool can result in strange behaviour in some circumstances.

When the number of unack'd messages exceeds the worker pool size, the QueueConsumer thread blocks waiting for a worker. Consequently, any outstanding message acks are not processed. This can result in deadlock if the running workers use RPC proxies.

The QueueConsumer thread is also responsible for sending heartbeats. If it thread ends up waiting for the worker pool longer two heartbearts, the broker will the connection. Any in-flight messages will be re-queued and the workers duplicated. Any currently running workers that are waiting for an RPC reply will raise "disconnected waiting for reply" errors upon reconnection.

This bug has been present for a long time but is easier to trigger as a result of the change in https://github.com/nameko/nameko/pull/419.

There is a simple fix, which I will raise in a PR now.

This raises the question of whether the worker pool is a useful construct or not. I am still forming an opinion but my current thoughts are:

* 10 workers is way too low for the default
* perhaps we should remove the restriction altogether, and just let throughput be constrained by available resources
* it may be useful for some _entrypoints_ to be constrained (i.e. only run x of these types of workers concurrently) but having a global constraint shared amongst all types of workers causes confusion without really solving a problem

sideboltconor:
Hey Matt,

You must be reading my mind as I have been struggling with these deadlocks today. Last night one of our servers died which created more load on our other nodes, which then starting raising deadlock errors like crazy. 

Our solution to the deadlocks has always been to just deploy more nodes. I've never fully understood the right max_worker count for our deployment. Our gateway (web socket service) is set to 1,000 max_workers and all other services are set to 100. These were just guesses, and the reason we set the web socket service to 1000 is that it seemed the most susceptible to deadlocks. 

Anyway, a fix to this issues would be greatly appreciated by our team!

Thanks,
Conor

mattbennett:
Hi Conor,

Sorry to hear about this! We've also struggled with hung services occasionally at Student.com, but the culprit always seemed to be RPC proxies waiting for replies (usually due to a downstream service that was overloaded). That blinded us to this underlying problem, which has existed in some form for a long time but is made much worse by the 2.5.3 release. I've tried to capture the various failure cases in the tests in the PR.

Which version of Nameko are your services running?

[...]

PR #418, opened 2017-03-16
quick fix low throughout capacity

mikegreen7892003:
Fix issue discussed in #417

PR #429, opened 2017-04-11
Prefetch count worker pool fix

mattbennett:
Fix for https://github.com/nameko/nameko/issues/428


nameko/messaging.py
```diff
@@ -396,8 +396,13 @@ def get_consumers(self, consumer_cls, channel):
         """
         _log.debug('setting up consumers %s', self)
 
+        def handle_message(provider, body, message):
+            eventlet.spawn_n(provider.handle_message, body, message)
```
davidszotten: should this use `container.spawn_managed_thread`?

test/test_regression.py
```diff
@@ -0,0 +1,312 @@
+""" Regression tests for https://github.com/nameko/nameko/issues/428
```
davidszotten: maybe a more specific filename if this is just for this particular problem

nameko/messaging.py
mattbennett: Yes, it should. Thanks

test/test_regression.py
mattbennett: I wondered whether to keep these tests in the repo. They were helpful for verifying the fix, and are a good reference, but they're quite slow even when they pass (~45 seconds).

I could pull them out into a gist and reference them here and in the issue?

davidszotten: not sure what you mean by "as a reference" but i guess i'd lean towards keeping. how come they are so slow?

mattbennett: By "as a reference" I mean it would be nice to see all the different failures that this fix addresses, alongside the fix.

They're slow because several of them only fail after heartbeats are missed, and it can't realistically beat faster than once every few seconds.

nameko/messaging.py
```diff
@@ -389,6 +389,11 @@ def connection(self):
         )
         return Connection(self.amqp_uri, heartbeat=heartbeat)
 
+    def handle_message(self, provider, body, message):
+        self.container.spawn_managed_thread(
+            lambda: provider.handle_message(body, message)
```
davidszotten: `identifier`?

test/test_regression.py
davidszotten: if the problem was that blocking in `handle_message` blocks the whole queue consumer, maybe it would be enough to check that that is no longer happening, rather than explicitly checking that heartbeats still work?

mattbennett: Good idea.

nameko/messaging.py
mattbennett: Good spot, thanks

mattbennett:
Unfortunately this change undoes the performance workaround added in #418.

The test is flakey. It passes reliably for me locally, but fails reliably here on Travis. It seems to be a problem of timing based on network traffic to and from rabbit. You can sometimes end up waiting for a timeout on the socket before being able to ack, and it seems once it happens everything falls into lock-step. 

Now the `QueueConsumer` delegates handling the message to a new thread, we end up in this situation much more frequently in this particular test (on linux at least).

I guess the fix in #418 is also unreliable, but we didn't happen to hit a failure condition in the (very limited) test. In reality, services that are susceptible probably fall back into `prefetch_count / safety_interval` requests/second eventually, even with this fix. Most real services probably aren't susceptible because they're handling traffic for more than one entrypoint, which will break the "timeout" lock-step again.

There's not much good news here. We know there are a couple of deep bugs with the current implementation, which have been around for a long time but are made worse by recent releases:

* Services handling lots of tiny requests can end up with throughput limited to `prefetch_count / safety_interval` requests per second. Nameko 2.4.4 increased the safety interval to a second, so makes this much more obvious. Nameko 2.5.3 contains a quick workaround, which I suspect doesn't work in all scenarios.

* Services ending up with more unack'd messages than their worker pool will block their `QueueConsumer`, which can result in deadlock and/or disconnections. Prior to Nameko 2.5.3 this was only possible with the unack'd messages spread across multiple AMQP queues. Since 2.5.3 it will happen whenever the number messages on a queue exceed the worker pool size, because the `prefetch_count` is always more than `max_workers`.

I think the changes in this PR need applying as soon as possible. We should perhaps also revert the prefetch count change from 2.5.3, since I'm unconvinced that it works as intended.

The performance problem still exists. To mitigate, we could change the `safety_internal` back to `0.1` again, but this doesn't actually _solve_ the problem. It also negates the reason I changed it in the first place, which was to (possibly) fix the connection flakiness we were seeing.

I think my preference in the short term is to reduce the safety internal again in this PR, and revert 2.5.3's change. And then we need to look more urgently at replacing the `QueueConsumer`.

davidszotten:
what makes you "unconvinced that it [prefetch count bump] works as intended"? 

when trying to reproduce the timeout currently on travis locally, are you only trying os x? or also linux (vm and/or docker)?

mattbennett:
When 2.5.3 was released I thought the performance restriction came from the fact that the queue consumer has having to block for the worker pool, and that the extra prefetch would allow it to ack before the worker was complete.

It turns out that this is not what's happening at all, which is obvious now the queueconsumer thread spawns a thread to wait on the pool. The performance becomes slow when the network traffic to rabbit is such that the queueconsumer has to wait for a socket timeout before being able to ack previous messages. The relative size of the worker pool is unrelated, so my conclusion is that bumping the prefetch worked by accident. I will try to prove this.

I can reproduce the `test_prefetch_throughput` failure locally on linux, and the difference between it and OS X is just that in the failing case we end up timing out on the socket on every iteration. I said "once it happens everything falls into lock-step" because I've also observed a test running without timeouts but then falling into them, at which point the performance drops to `prefetch_count / safety_interval` reqs/s.

I've also seen this test pass on Travis, but only once or twice, which further suggests that it's an issue of timing.

mattbennett:
The `prefetch_count = max_workers + 1` fix works because it guarantees that when a worker finishes, there's another thread waiting to run (blocked on the worker pool). So you never end up with all the activity happening at once, then having to wait for a timeout on the socket.

That means the fix *does* always work, but only if you're prepared to block the queue consumer thread on the worker pool.

mattbennett:
Tests seem to be failing on a hanging toxiproxy server (https://github.com/Shopify/toxiproxy/issues/159).

I think lowering the `safety_interval` has changed the timing sufficiently to now reliably hit the hanging scenario. I can reproduce it locally (on linux only) and increasing the `safety_interval` makes it more stable.

mattbennett:
Having said in #428 that this was a "simple" fix, it's turning out to be quite difficult to actually land.

I'm not happy with these changes. I'm convinced that main change here is correct (not blocking the queueconsumer on the worker pool) but the fallout is bad because we re-expose the performance problem stemming from acking messages inside the queue consumer loop.

Trying to fix the performance problem I stumbled across [this stack overflow question](http://stackoverflow.com/questions/7974618/can-eventlet-manage-an-amqp-connection-with-messages-passing-asynchronously-both) (which, ironically enough, I'd previously commented on and then forgotten about).

It turns out that you can safely ack messages from any thread, and doing so completely removes the performance problem. I have a branch in progress that does this: https://github.com/mattbennett/nameko/compare/master...mattbennett:simplify-queue-consumer?expand=1. Unfortunately there is test-fallout to fix here too, because it has a substantial impact on timing and thread execution order, and our tests are too delicate.

mattbennett:
The "branch in progress" mentioned above is now ready for review in #430.

I've implemented the salient changes from this PR on top of it in #431, so closing this in preference.

PR #431, opened 2017-04-18
Dont block queueconsumer on worker pool

mattbennett:
Implements the fix from #429 on top of #430.


nameko/messaging.py
```diff
@@ -389,6 +355,14 @@ def connection(self):
         )
         return Connection(self.amqp_uri, heartbeat=heartbeat)
 
+    def handle_message(self, provider, body, message):
+        ident = "{}.handle_message[{}]".format(
+            type(provider).__name__, message.delivery_info['routing_key']
+        )
+        self.container.spawn_managed_thread(
+            lambda: provider.handle_message(body, message), identifier=ident
```
davidszotten: i think elsewhere we do this with a `partial` instead of a `lambda` though not sure about pros and cons. maybe one less function call / shorter call stack?

nameko/retry.py
```diff
@@ -0,0 +1,53 @@
+import functools
+import itertools
+from time import sleep
+
+import wrapt
+
+
+class RetryDelay(object):
+    def __init__(self, delay, backoff, max_delay):
+        self.delay = delay
```
davidszotten: is this copied from somewhere? we don't seem to actually use e.g. `max_delay`

nameko/retry.py
```diff
@@ -0,0 +1,53 @@
+import functools
```
davidszotten: `nameko/utils/retry.py` maybe?

test/test_retry.py
```diff
@@ -0,0 +1,141 @@
...
+        fn()
+
+    assert tracker.call_count > 1
+
+
+@pytest.mark.parametrize("exceptions", [ValueError, (ValueError,)])
+def test_retry_for_exceptions(exceptions, tracker):
+    threshold = 10
+
+    @retry(for_exceptions=exceptions, max_attempts=float('inf'))
```
davidszotten: isn't `inf` the default `max_attempts`?

nameko/messaging.py
```diff
@@ -389,6 +355,14 @@ def connection(self):
         )
         return Connection(self.amqp_uri, heartbeat=heartbeat)
 
+    def handle_message(self, provider, body, message):
+        ident = "{}.handle_message[{}]".format(
```
davidszotten: are routing keys always ascii?

davidszotten: actually, i guess provider could also be non-ascii

nameko/retry.py
```diff
@@ -0,0 +1,53 @@
...
+    wrapped=None, for_exceptions=None, max_attempts=3,
+    delay=1, backoff=1, max_delay=None
+):
+    if wrapped is None:
+        return functools.partial(
+            retry, for_exceptions=for_exceptions, max_attempts=max_attempts,
+            delay=delay, backoff=backoff, max_delay=max_delay
+        )
+
+    for_exceptions = for_exceptions or Exception
```
davidszotten: how come `Exception` can't be the default arg?

test/conftest.py
```diff
@@ -52,7 +52,18 @@ def toxiproxy_server():
         ['toxiproxy-server', '-port', str(port), '-host', host],
         stdout=subprocess.PIPE
     )
-    time.sleep(0.2)  # allow server to start
+
+    class NotReady(Exception):
+        pass
+
+    @retry(delay=0.1, max_attempts=10)
```
davidszotten: should this be for `NotReady` only?

nameko/messaging.py
mattbennett: Does it matter? The `identifier` is only used to help identify thread when they're killed: https://github.com/nameko/nameko/blob/master/nameko/containers.py#L470

nameko/retry.py
mattbennett: Yes that's better, will move it.

nameko/retry.py
mattbennett: Yeah, I lifted it from our application codebase. Actually it's previously been part of the Nameko codebase too, but we reverted it in favour of a less generic solution (which has since been rolled back).

I think if it's going to live as a helper function it's useful to keep the `max_delay` argument.

nameko/retry.py
mattbennett: No reason! I'll change it :)

test/conftest.py
mattbennett: Yes it should. Thanks

test/test_retry.py
mattbennett: No, default is 3.

I think it would be dangerous to have infinite max attempts _and_ `for_exceptions=Exception` as the defaults --
 too easy to end up retrying forever.

nameko/messaging.py
davidszotten: well your format string needs to be unicode

test/test_retry.py
davidszotten: ah yes, misread it above. and reasoning makes sense to me

test/conftest.py
mattbennett: Actually no, we want to catch exceptions from requests here too, e.g. if the server isn't accepting connections yet.

PR #432, opened 2017-04-18
Update rpc.py

2XL:
may be ==> maybe

### DIFF
--- a/nameko/standalone/rpc.py
+++ b/nameko/standalone/rpc.py
@@ -280,7 +280,7 @@ class ClusterProxy(object):
     cluster. It is commonly used as a context manager but may also be manually
     started and stopped.
 
-    This is similar to the service proxy, but may be uses a single reply queue
+    This is similar to the service proxy, but maybe uses a single reply queue
     for calls to all services, where a collection of service proxies would have
     one reply queue per proxy.
 


davidszotten:
Hi,

thanks for noticing the sentence being off. however, i don't think the maybe(/may be) should be there at all, but rather read "... but uses a single reply queue ..."

2XL:
ok,
another Question:
following the docs
#helloworld.py
class TargetServiceName:
    name = "target_service_name"

    @rpc
    def method(self, name='method'):
        msg = "Hello, {}!".format(name)
        print(msg)
        return msg


#nameko_examples/client_rpc_helper.py
cluster_proxy = ClusterRpcProxy(config)
cluster_proxy = cluster_proxy.start()
cluster_proxy.target_service_name.method()
cluster_proxy.stop() 

`
  File "nameko_examples/client_rpc_helper.py", line 33, in <module>
    cluster_proxy.stop()
TypeError: 'ServiceProxy' object is not callable
`

what does this mean? 

Why am I getting this error?

davidszotten:
looks like another bug in the docs. the problem is variable re-use `cluster_proxy = cluster_proxy.start()`

the `stop` should be called on the first object, but it's replaced by the result of `start()`.

the best option is probably to use the context manager version (though we should fix the docs)

2XL:
ok solved
```python
#################################################################
# CLUSTER_PROXY
#
# with ClusterRpcProxy(config) as cluster_proxy_raw:
#     cluster_proxy_raw.greeting_service.hello()

cluster_proxy = ClusterRpcProxy(config)
cluster_proxy_connector = cluster_proxy.start()
cluster_proxy_connector.target_service_name.method()
cluster_proxy.stop() 
```

Thanks David!, appreciate your support.

PR #435, opened 2017-05-07
bump oldest reqs to support python 3.5 and 3.6. drop support for 3.3.

davidszotten:


davidszotten:
argh, forgot to update travis. are there projects that manage travis/tox integration automatically?

mattbennett:
The best I've found is tox-travis.

I actually have a branch where I was playing with more python versions -- https://github.com/nameko/nameko/compare/master...mattbennett:more-pythons

It's more complex to dynamically build the matrix, but it does avoid having to list out every possible combination as a separate `TOX_ENV` entry.

mattbennett:
Actually now that I think about it I'm remembering things...

The `TOX_ENV` approach doesn't work on Travis if you want to test on anything other than python 2.7, 3.3 or 3.4. Those are the only Pythons available without specifying them in the `python` section of the Travis config file.

Once you have multiple entries in the `python` section, Travis will run every provided environment on each one, so you get enormous duplication if you try to use the environment to specify the Python version.

PR #436, opened 2017-05-10
add test to make sure tox.ini and .travis.yml match

davidszotten:


mattbennett:
Is this preferable to using [tox-travis](https://pypi.python.org/pypi/tox-travis)?

davidszotten:
iirc tox-travis isn't flexible enough

still on the fence about the approach in this pr

PR #437, opened 2017-05-15
Add availability to specify section in Config dependency provider

scarchik:
Hi,

I prepare minor change to config dependency provider which will simplify access to the config. 
Nameko config can contain many sections like LOGGING, ENGINE, TSDB_CONFIG, SERVICE_CONFIG, etc. 
In service most of all you need access only to one of this sections.

How it should work:
```
class Service:
    config = Config("SERVICE_CONFIG")

    def test(self):
        my_value = self.config['request_timeout']
```

### DIFF
--- a/nameko/dependency_providers.py
+++ b/nameko/dependency_providers.py
@@ -8,5 +8,11 @@ class Config(DependencyProvider):
     """ Dependency provider for accessing configuration values.
     """
 
+    def __init__(self, section=None):
+        self.section = section
+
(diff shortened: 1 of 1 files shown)


davidszotten:
Hi Sergey, and thanks for the contribution!

However, as `Config` also serves as simple example of a dependency provider, i think i would prefer to keep it as simple as possible.

You are of course able to create a subclass that better suits your needs as part of your own projects, or a standalone package

Best,
David

PR #438, opened 2017-05-17
Context local worker_ctx

defat:
global.worker_ctx context local variable added (request in flask style).
It is pushed in entrypoints and in standalone proxy creation.
It can be used in the scope of request handling e.g. in log formatters for logging call_ids, etc.
We write logs in json, tag them with call_id_stack[0] and store them in elasticsearch. With such approach we are able to retrieve logs from all microservices for specified request.

### DIFF
--- a/nameko/containers.py
+++ b/nameko/containers.py
@@ -370,7 +370,7 @@ class ServiceContainer(object):
                    worker_ctx, '->'.join(worker_ctx.call_id_stack))
 
         with _log_time('ran worker %s', worker_ctx), \
-             push_worker_ctx(worker_ctx):
+                push_worker_ctx(worker_ctx):
             self._inject_dependencies(worker_ctx)
             self._worker_setup(worker_ctx)
 
--- a/nameko/containers.py
+++ b/nameko/containers.py
@@ -369,7 +369,8 @@ class ServiceContainer(object):
         _log.debug('call stack for %s: %s',
                    worker_ctx, '->'.join(worker_ctx.call_id_stack))
 
-        with _log_time('ran worker %s', worker_ctx), push_worker_ctx(worker_ctx):
+        with _log_time('ran worker %s', worker_ctx), \
+             push_worker_ctx(worker_ctx):
             self._inject_dependencies(worker_ctx)
             self._worker_setup(worker_ctx)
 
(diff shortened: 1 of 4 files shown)


mattbennett:
Nameko makes a conscious choice to use dependency injection rather than global or thread-local variables. Dependency injection is chosen because it makes testing easier, and many of the bundled test helpers rely on this pattern.

The "nameko way" of achieving your use-case would be to define a `DependencyProvider` that performed the logging. For an example, see [this mailing list thread](https://groups.google.com/forum/#!searchin/nameko-dev/Logger%7Csort:relevance/nameko-dev/dPpeETDgduo/IKOeg-HzAQAJ).

It's also viable to expose different parts of the `worker_ctx` directly to workers if you need to (as in [contextdata.py](https://github.com/nameko/nameko/blob/master/nameko/contextdata.py)), but it should always be done via a `DependencyProvider`.

defat:
Implementing this in a DependencyProvider instead of global variable seems extremely reasonable for me.
However there is only one way to append correlation_id or call_id to **every** log record including logs from other libraries - to implement a [filter object](https://docs.python.org/2/library/logging.html#filter-objects) (not formatter as I mentioned earlier, sorry for this). Implementation should be smth like:

```python
class CorrelationIdFilter(logging.Filter):
    def filter(self, record):
        if global_worker_ctx:
            record.correlation_id = global_worker_ctx.call_id_ctack[0]
        return True
```
With logging cfg smth like:

```yaml
LOGGING:
  ...
  formatters:
    ....
      format: '(asctime)(msecs)(levelname)(process)(module)(lineno)(args)(msg)(correlation_id)'
      class: pythonjsonlogger.jsonlogger.JsonFormatter
  filters:
    request_id_filter:
      (): my_module.CorrelationIdFilter
...
```
Such approach gives ability to tag and retrieve **all** log records made for particular request even made by other libraries.
Looks like implementing DependencyProvider storing needed information from worker_ctx in a global variable that will be used in log filter is the right way.

mattbennett:
I understand the desire now. In this case, I would implement a `DependencyProvider` that pulled the current `call_id` out and stashed it in a thread-local variable during `worker_setup`. Your logging filter could then just read the call id from there, and there'd be no need for internal changes to Nameko.

Also, note we use a "correlation_id" as an internal part of the RPC protocol. You might want to change your naming convention to reflect that what you're capturing is really the Nameko _call_id_.

fraglab-admin:
Yeah, I've used correlation_id just for example.
Thanks for so fast feedback and recommendations!
Pull request should be closed.

defat:
Yeah, I've used correlation_id just for example.
Thanks for so fast feedback and recommendations!
Pull request should be closed.

PR #440, opened 2017-06-02
promote sensitive_variables and expected_exceptions to work with all entrypoints

mattbennett:
Allows all entrypoints to accept `sensitive_variables` and `expected_exceptions`, as long as they accept ``**kwargs`` and pass them to `super()`.

Also consolidates the existing tests for these features into a single module.


nameko/testing/services.py
```diff
@@ -476,8 +476,15 @@ class Once(Entrypoint):
     the service container started.
     """
     def __init__(self, *args, **kwargs):
+        expected_exceptions = kwargs.pop('expected_exceptions', ())
```
davidszotten: why is this required?

mattbennett: The `once` entrypoint is a bit special in that is passes `**kwargs` directly to the entrypoint. If you don't pop them off you get unexpected argument errors.

mattbennett:
I'm pretty sure this is backwards compatible. Custom entrypoints _may_ need updating to support the API if they don't accept `**kwargs` and/or call `super().__init__()`.

Good call on the docs. Will add a page 👍

mattbennett:
@davidszotten docs added.

A little more than I anticipated, but I had to add a long overdue thing about _writing_ entrypoints before I could talk about the arguments they accept.

docs/examples/auth.py
```diff
@@ -0,0 +1,70 @@
...
+class Auth(DependencyProvider):
+
+    class Api:
+        def __init__(self, users, worker_ctx):
+            self.users = users
+            self.worker_ctx = worker_ctx
+
+        def authenticate(self, username, password):
+            try:
+                assert self.users[username]['password'] == password
```
davidszotten: i know this is just a toy but i'm still not sure i'm comfortable with example code suggesting stored plain text passwords

mattbennett: Fair. I will update it.

mattbennett:
@davidszotten this is also ready for another review when you have time.

PR #442, opened 2017-06-05
modern requests has strict urllib3 reqs

mattbennett:


setup.py
```diff
@@ -35,14 +35,11 @@
         'dev': [
             "coverage==4.2",
             "flake8==3.2.1",
-            "mccabe==0.6.1",
-            "pycodestyle==2.3.1",
-            "pyflakes==1.5.0",
```
davidszotten: how do you mean we're not using these? does flake8 pin exact versions of these?

setup.py
```diff
@@ -35,14 +35,11 @@
...
             "pylint==1.7.1",
             "pytest==2.7.3",
             "pytest-cov==2.1.0",
             "pytest-timeout==0.4",
-            "urllib3==1.10.2",
+            "urllib3==1.21.1",
```
davidszotten: do we also depend on requests?

setup.py
mattbennett: I forgot that we pinned exact versions of dependencies used by tests. Will revert

setup.py
mattbennett: `requests` is an install requirement. It is used in tests too - should we pin a version if you install `[dev]` requirements?

davidszotten: ah, of course. i think i still like the idea of pinning everything in `dev`. since it's dev, there's no problem bumping these and i think we tend to try to keep them recent when we remember.

(i thought requests bundled its own urllib3 but as above happy to bump)

mattbennett: Until very recently (2.16.0) requests did vendor urllib3, although [here](https://github.com/nameko/nameko/blob/0c44614bd1eb93c00977bb48169d0245a25eb453/test/test_amqp.py#L7) we are using it directly anyway.

I will pin requests here in `[dev]` so everything is consistent, and bump all the versions to something recent.

ornoone:
this PR can solve the build error on #443. can you merge it ?

Issue #443, opened 2017-06-07
nameko shell display wrong banier info


ornoone:
I you use `nameko shell --config myconfig.yaml`, the banniere display some bad info (it seem it display the --broker value instead of the one in the config file.

davidszotten:
Hi,

good catch. would you like to try to write a fix?

ornoone:
no problem. I will work on it this morning

davidszotten:
fixed by #451

PR #444, opened 2017-06-08
#443: fix banner on nameko shell.

ornoone:
display now the broker from --config if given

I did not manage to run tests for the entire project. so no unit test is present.

(some tests fails unexpectedly on master without any modifications from me)

### DIFF
--- a/nameko/cli/shell.py
+++ b/nameko/cli/shell.py
@@ -70,17 +70,21 @@ Usage:
 
 
 def main(args):
-    banner = 'Nameko Python %s shell on %s\nBroker: %s' % (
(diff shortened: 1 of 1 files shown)


mattbennett:
There are a couple of flakey tests in master. I restarted this job and it passed this time.

It would be nice if there was a unit test verifying the banner. Do you want to try adding one? If not, I'm happy to add one on top of this work

ornoone:
I tried to add some tests, but I was unable to run some other random tests with success in my computer. 
If you can write it, I will be eager to see what is does look like and I will try to make it run localy.

PR #446, opened 2017-06-13
Avoid DeprecationWarning for getfuncargvalue in recent pytest

### LINKED ISSUE #445: request.getfuncargvalue is deprecated since PyTest 3.0
https://docs.pytest.org/en/latest/builtin.html#_pytest.fixtures.FixtureRequest.getfuncargvalue
https://docs.pytest.org/en/latest/changelog.html#id79

Yes, I know, nameko pins PyTest at version 2.7.3 in setup.py, but this `DeprecationWarning` shows up when testing code that uses nameko.

frexvahi:
Fixes #445 

I tried to keep compatibility with old PyTest versions while avoiding DeprecationWarnings in new versions. I would appreciate suggestions on how to ensure that the continuous integration gets run both with the currently-used pytest-2.7.3 and a more recent PyTest version.

### DIFF
--- a/nameko/testing/pytest.py
+++ b/nameko/testing/pytest.py
@@ -201,7 +201,11 @@ def fast_teardown(request):
     reorder_fixtures = ('container_factory', 'runner_factory', 'rabbit_config')
     for fixture in reorder_fixtures:
         if fixture in request.funcargnames:
-            request.getfuncargvalue(fixture)
+            # getfuncargvalue is renamed to getfixturevalue in pytest 3.0
+            if hasattr(request, 'getfixturevalue'):
+                request.getfixturevalue(fixture)  # pragma: no cover
+            else:
+                request.getfuncargvalue(fixture)  # pragma: no cover
 
     consumers = []
 


mattbennett:
This is tricky because pytest is both a _test_ requirement (therefore pinned) and a _library_ requirement which we allow to float. For these dependencies we don't actually test  oldest, pinned and latest versions like we do for all the other library requirements. Perhaps we should. 

Unfortunately Nameko's test suite doesn't support pytest>2.7.3 because of some coverage collection issues. Even if it did though, we'd either need to merge the coverage runs or exclude the lines that don't apply to specific versions.

I think the pragmatic fix here is to try to use `getfixturevalue`, fall back to `getfuncargvalue`, and add a `pragma: no cover` to the fallback path.

frexvahi:
The CI failure is due to a `TimeoutError` in an unrelated component, can it be merged anyway?

davidszotten:
manually declaring errors to be unrelated (even if correct in this instance) is too error prone for my liking. have re-triggered the (sub) job that failed. unfortunately i think travis only lets maintainers do this

Issue #448, opened 2017-06-24
Call function by name?


geoffjukes:
Hi,

Is it possible to call a method using a string?

I just discovered nameko - awesome. Fits perfectly into our stack, without the overhead of celery for purely rpc methods. We're using it like an internal-lambda :)

I have written a REST API using flask/connexion/swagger for specific rpc tasks, but I'd like to add a generic 'call' endpoint (such as `/rpc/call/{service}/{method}` and was wondering how to feed that into the cluster rpcproxy

Many thanks,

Geoff

geoffjukes:
Solved! I read thru the source and noted the dictionary-like access for the service. So I combined that with a getattr, and it works. 

```
def call(service, method, **kwargs):
     with ClusterRpcProxy(config) as cluster_rpc:
         return getattr(cluster_rpc[service], method)(**kwargs)
```
Adding the dict-like access for the method names would simplify my code, and I'll see about submitting a pull request. 

Awesome tool. Thanks for releasing it!

Geoff

PR #449, opened 2017-06-24
Enable dict-like access on the ServiceProxy

geoffjukes:
Enables a call like:

```
def call(service, method, **kwargs):
    with ClusterRpcProxy(config) as cluster_rpc:
        return cluster_rpc[service][method](**kwargs)
```

### DIFF
--- a/nameko/rpc.py
+++ b/nameko/rpc.py
@@ -360,6 +360,9 @@ class ServiceProxy(object):
         return MethodProxy(
             self.worker_ctx, self.service_name, name, self.reply_listener)
 
+    def __getitem__(self, name):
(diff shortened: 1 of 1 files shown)


mattbennett:
Thanks @geoffjukes for the contribution! Glad you're enjoying Nameko and managed to solve your problem in #448.

In general I prefer not to have two ways of doing things. We added dict-access for services because they can contain characters that can't be used in python identifiers (e.g. "my-service"). That isn't case for method names because they also need to exist as methods on the service class.

Since Python already provides the `getattr` helper I'd rather not add this extra method to the Nameko codebase.

geoffjukes:
Fair enough :)

PR #451, opened 2017-06-28
Fix banner nameko on shell

mattbennett:
Adds tests on top of @yupeek's #444


ornoone:
thanks for the tests.

mattbennett:
Thank you @ornoone!

PR #456, opened 2017-07-20
Refactor AMQP publishing, expose delivery options 

mattbennett:
This consolidates AMQP message publishing code into a shared module, and exposes delivery options where appropriate for the built-in DependencyProviders that publish messages (`RpcProxy`, `EventDispatcher`, `Publisher`).

The DependencyProviders now allow certain message and delivery options to be specified at instantiation or publish time, for example:

``` python
class Serv:
     
    # publish transient messages, don't confirm receipt, never attempt to retry
    publish = Publisher(delivery_mode=1, use_confirms=False, retry=False)

    @entrypoint
    def meth(self):

        # publish a message with the default options
        self.publish("msg", routing_key="foo")

        # do retry, but only once
        self.publish("msg", routing_key="foo", retry=True, retry_policy={'max_attempts': 1})
```

It's possible to provide headers and declarations in the same way:

``` python
class Serv:

    publish = Publisher(
        # always add this header to published messages
        headers={'foo': 'foo'},
        # ensure these queues and exchanges exist before publishing
        declare=[some_queue, some_exchange]
    )

    @entrypoint
    def meth(self):
        self.publish(
            "msg",
            headers:{'foo': 'FOO', 'bar': 'BAR'},  # extend or override default headers
            declare: [another_queue]  # more entities to declare before publishing
        )
```

The `TestConfigurability` tests for the DependencyProviders show which options can be specified for each (e.g. you can't specify a different routing key or exchange for an `EventDispatcher`).

This feature needs documenting but there's not currently anywhere sensible to describe it. I would like (soon) to add a dedicated page to the docs for each of the AMQP extensions, but this PR is big enough as it is.


nameko/messaging.py
```diff
@@ -101,27 +110,46 @@ def spam(self, data):
...
+
+        if self.exchange:
+            self.declare.append(self.exchange)
+
+        if queue is not None:
+            warnings.warn(
+                "The signature of `Publisher` has changed. The `queue` kwarg "
+                "is now deprecated. You can use the `declare` kwarg "
+                "to provide a list of Kombu queues to be declared. "
+                "See CHANGES, version 2.5.2 for more details. This warning "
```
davidszotten: wrong version?

nameko/messaging.py
```diff
@@ -130,71 +158,31 @@ def serializer(self):
...
-
-            with get_producer(self.amqp_uri, self.use_confirms) as producer:
-                headers = self.get_message_headers(worker_ctx)
-                producer.publish(
-                    msg, exchange=exchange, headers=headers,
-                    serializer=serializer, retry=retry,
-                    retry_policy=retry_policy, mandatory=mandatory,
-                    **kwargs
-                )
+        propagate_headers = self.get_message_headers(worker_ctx)
```
davidszotten: how come this var is called `propagate_` and not `extra_`? (propagate sounds like a bool to me)

nameko/rpc.py
```diff
@@ -184,59 +191,15 @@ def handle_result(self, message, worker_ctx, result, exc_info):
...
-        It is strongly recommended to use publish confirms in RPC Responders.
-        Without them, replies in an unstable network environment may be lost,
-        leaving the caller waiting indefinitely for a response.
-        """
-        return True
-
-    @property
-    def serializer(self):
-        """ Name of the serializer to use when publishing response payloads.
+    Publisher = Publisher
```
davidszotten: i think we tend to name vars like this `publisher_cls`

nameko/standalone/events.py
```diff
@@ -24,50 +21,28 @@ def event_dispatcher(nameko_config, **kwargs):
...
-    use_confirms = kwargs.pop('use_confirms', True)
-    mandatory = kwargs.pop('mandatory', False)
+    serializer = kwargs.pop(
+        'serializer',
+        nameko_config.get(
+            SERIALIZER_CONFIG_KEY, DEFAULT_SERIALIZER
+        )
+    )
+
+    # TODO: standalone event dispatcher should accept context event_data
```
davidszotten: todo?

nameko/testing/utils.py
```diff
@@ -89,6 +90,13 @@ def assert_stops_raising(fn, exception_type=Exception, timeout=10,
             eventlet.sleep(interval)
 
 
+MockCallArgs = namedtuple("MockCallArgs", ["positional", "keyword"])
+
+
+def unpack_mock_call(call):
```
davidszotten: do you prefer this to `call_args, call_kwargs = call`?

davidszotten:
also, didn't realise you were still using cagoule! still working ok? given the size of that file, it might be worth re-writing history to remove all traces of it before landing this pr

nameko/amqp/__init__.py
```diff
@@ -0,0 +1,4 @@
+# backwards compat imports
+from .utils import verify_amqp_uri  # noqa: F401
```
mattbennett: They're "unused imports" otherwise. I'm only pulling them into this package for a backwards compatibility import of `from nameko.amqp import verify_amqp_uri`

nameko/amqp/publish.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def __init__(
+        self, amqp_uri, use_confirms=None, serializer=None, compression=None,
+        delivery_mode=None, mandatory=None, priority=None, expiration=None,
+        declare=None, retry=None, retry_policy=None, **publish_kwargs
+    ):
+        self.amqp_uri = amqp_uri
+
+        # publish confirms
+        if use_confirms is not None:
```
mattbennett: Several of them have truthy defaults, so it's hard to compare whether a new value has been provided. This is a little more verbose but consistent for every param. I also like that the default values sit next to the docs for each option.

A slightly less verbose option could be the following:

``` python
def __init__(self, amqp_uri, **kwargs):
    self.amqp_uri = amqp_uri

    self.use_confirms = kwargs.pop('use_confirms', self.use_confirms)
    ...
```

Slight disadvantage here is that the sphinx docs would only show `**kwargs`.

Re: adding too many ways to change settings. Maybe. This does allow subclassing to override defaults as well as at instantiation (and use-time). Subclassing used to be the only way to change these things previously though, and we encourage subclassing for other reasons.

nameko/messaging.py
```diff
@@ -3,22 +3,22 @@
 '''
 from __future__ import absolute_import
 
-import warnings
 from functools import partial
 from logging import getLogger
+import warnings
```
mattbennett: Perhaps, although it's going to be another annoying step that frequently gets forgotten and fails the build. Out of order imports are annoying though, and so are casual rearrangements inside other commits.

Could we automatically re-write them on a commit hook? Then the order will always be consistent and any re-arrangement would be made automatically and in its own commit.

nameko/messaging.py
mattbennett: Yep, thanks. Shows how long this PR has been in the works!

nameko/messaging.py
mattbennett: I think I was trying to communicate that these are headers that propagate from one call to the next. I agree it sounds like a bool.

I guess it's good enough to call them `extra_headers`.

nameko/standalone/events.py
mattbennett: Wasn't sure whether to add this as part of this PR. It's much easier to do now, and it would bring it in line with the standalone RPC proxy.

nameko/testing/utils.py
mattbennett: I think it's slightly easier to interpret using the named attribute rather than remembering that it unpacks to `args, kwargs`. If you're only interested in one or the other you don't have to do `args, _ = call` or whatever.

I'm not wedded to it if you'd rather not introduce a new helper though.

Really I just want Mock's `call` to be a namedtuple out of the box.

nameko/messaging.py
davidszotten: i would be -1 on auto-rewriting. status quo is probably fine, though a checking is a fairly fast lint run with e.g isort. also has the bonus of _optionally_ autosorting. something for another day

nameko/testing/utils.py
mattbennett: @tonybajan is that an endorsement for `call` being a namedtuple or for the helper?

tonybajan: for call being a namedtuple – makes so much more sense

davidszotten: i had quick look in the source, and `Call` already has a bunch of interesting behaviours, e.g. it compares successfully to both a tuple (compared against `args`) and a dict (matched against `kwargz`) but nothing obvious to check membership in kwargs like you do

happy to keep this helper if you like it

mattbennett:
@davidszotten I think I've addressed all the comments here. History also rewritten to remove the cagoule file.

ianjuma:
@mattbennett besides this PR - there's no mention of `delivery_mode` on the docs? or this feature

Issue #420, opened 2017-03-17
Auto reload app with nameko run?


richardasaurus:
Is anyone planning to add auto reloading through `nameko run`? If not I might add it as its very annoying having to restart it manually all the time.

mattbennett:
This has been discussed before but nothing has landed in the main repo. Check this thread on the mailing list -- https://groups.google.com/forum/#!searchin/nameko-dev/reload%7Csort:relevance/nameko-dev/aNAs2cHBJME/62QS1G3cAQAJ

It would be a welcome addition!

richardasaurus:
will implement this in the next week roughly

PR #457, opened 2017-07-20
adds auto-reloading + --logging-config-file option

richardasaurus:
https://github.com/nameko/nameko/issues/420

### How to enable

add `AUTORELOAD: true` to the nameko config file

---

Also adds `nameko run --logging-config-file ./logging.conf` (in this file format https://docs.python.org/2/library/logging.config.html#logging.config.fileConfig)

You may ask: 
**Why is a seperate logging config file needed?** Where I work there are many different python microservices (using different frameworks), all read their logging config from the same standard python format logging config file, so we cannot put it on the special nameko one, otherwise we'd have duplication.


richardasaurus:
@mattbennett @davidszotten

mattbennett:
This is a very welcome feature! Thank you for raising a PR.

I have a few thoughts --

There's a lot of code here for watching and tracking file changes. Is is possible to outsource some of that to a library? [Watchdog](http://pythonhosted.org/watchdog/) comes to mind. @davidszotten has used it as a _wrapper_ before, but I want this feature to become part of Nameko out of the box.

I'm not convinced by the value of a second config file for logging. I think it can just as easily go in the normal config file and then won't need any special treatment.

Finally, I think we use an `--auto-reload` flag instead of / as well as the config option to specify auto-reload would be helpful. That seems to be the "normal" way of enabling auto-reload in similar libraries, and makes it very easy to do in simple cases that might not have a config file yet.

davidszotten:
since we already depend on werkzeug we could also investigate using the [werkzeug reloader](https://github.com/pallets/werkzeug/blob/master/werkzeug/_reloader.py#L252)

richardasaurus:
@mattbennett @davidszotten 

I did try watchdog in the past but it failed to work, I can try again with watchdog/werkzeug reloader.

I'm happy to try that and add the "autoreload" cli arg, if you'll agree to the log file path addition? (if one thing can be controlled through the cli, why not the other).

Otherwise it makes more sense for me to stick this all in another package which just replaces the run command of nameko.

mattbennett:
My resistance to adding the logging config file option is only that there's already a way to configure logging via the normal config file.

What's the motivation for using a logging config file rather than the same information in a yaml file that can also contain the rest of the service configuration?

richardasaurus:
> What's the motivation for using a logging config file rather than the same information in a yaml file that can also contain the rest of the service configuration?

because when you have many services in different frameworks, they don't all have the same config file format (here we have .ini for flask, .yaml for nameko).

so we need a common format to configure logs for all of those. also if python supports log config from a file, why not support it here, as someone else will run into this at some point i'd think.

if you don't like the cli arg `--logging-config-file` then I could instead add a nameko config key which points to the logging config file path if that suits you better.

richardasaurus:
werkzeug contains this warning  @davidszotten hmmm

```python
def run_with_reloader(*args, **kwargs):
    # People keep using undocumented APIs.  Do not use this function
    # please, we do not guarantee that it continues working.
    from werkzeug._reloader import run_with_reloader
    return run_with_reloader(*args, **kwargs)
```

richardasaurus:
(also i tried that reloader and it doesn't work, the app won't even run)

mattbennett:
OK, let's abandon the werkzeug reloader. 

One more possibility is https://github.com/Pylons/hupper. It looks like it's process based so eventlet compatibility _should_ be fine (which is probably where watchdog fell down).

Every framework out there seems to have implemented their own reloader in the past. Hupper is itself a standardisation of other approaches. If we can leverage someone else's solution rather than rolling our own I think it'd be worth it.

For the logging config file -- the motivation is clear now, thanks. It's probably worth splitting that change out into a separate PR though, so each change can be discussed and landed independently.

richardasaurus:
ran out of time to make those changes so stuck this in another package https://github.com/Babylonpartners/nameko-extras feel free to borrow the code from there

PR #462, opened 2017-08-04
Set stopped flag to allow PollingQueueConsumer object reuse

bobh66:
ClusterRpcProxy objects can only be used once because the PollingQueueConsumer object sets it's stopped flag to False in the __init__() method and to True in the unregister_provider() method.  If the proxy is used a second time, the stopped flag is not reset to False when register_provider() is called, so the ConsumerEvent wait() method thinks the service has stopped, even though the message was sent and a response would have been received.

This PR adds a test for this case and sets stopped to True in the PollingQueueConsumer __init__() method, since register_provider() has not been called yet.  The stopped flag is set to False in register_provider() after the consumer has been set up.

This allows a single proxy instance to be used for multiple service invocations.


mattbennett:
@bobh66 thanks for the PR!

This looks good to me, although I can't remember why we protect against a stopped `PollingQueueConsumer`. @davidszotten do you recall?

davidszotten:
no, i was hoping you would :p. is the issue that stopping and starting again ends up re-doing all the work anyway and so we wanted to make this explicit?

mattbennett:
@davidszotten perhaps that was the reasoning. Perhaps the consumer setup logic was different back then too, and restarting was not an option?

I don't think it matters really, unless you can think of a scenario where this won't work? The test @bobh66 has added suggests it does!

davidszotten:
i guess my only worry would be if we make it seem like we're re-using machinery but are in fact re-setting everything up (with performance implications)

mattbennett:
You have to leave and reenter the context manager though, which usually implies some kind of work.

davidszotten:
fair

davidszotten:
@bobh66 as discussed i'm happy to accept this. thanks again for the report and the pr!

could you please add a small note to the changelog, and (if you want) add yourself to the contributors list?

bobh66:
@davidszotten will do - thanks!

bobh66:
@davidszotten - should I add an entry to the CHANGESfile with no associated version, and assume the version will be added when it is released?  Thanks

davidszotten:
yep, that's our normal process. just mark it 'unreleased'. thanks

PR #463, opened 2017-08-09
Rpc remote error

iky:
This PR ads a decorator for registering remote exceptions to existing exception registry, thus adding the ability to handle RPC exceptions inside RPC entrypoint methods with using standard error handling and without the need of "unwrapping" the original exceptions from `RemoteError` and loosing the original args.

I think that the existing "path" mapping used for "transporting" internal errors works well for custom errors too and that the little decorator brings a straightforward, clear and readable solution from Nameko internals to the API of the framework.

### DIFF
--- a/nameko/exceptions.py
+++ b/nameko/exceptions.py
@@ -36,6 +36,23 @@ class ContainerBeingKilled(Exception):
 registry = {}
 
 
+def remote_error(exc_path):
+    """
+    Register exception as remote error
+
+    Decorator that registers remote exception with matching ``exc_path``
+    to be deserialized to decorated exception instance, rather than
+    wrapped in ``RemoteError``.
+
+    """
+
+    def wrapper(exc_type):
+        registry[exc_path] = exc_type
+        return exc_type
+
+    return wrapper
+
+
 def get_module_path(exc_type):
     """ Return the dotted module path of `exc_type`, including the class name.
 
--- a/nameko/exceptions.py
+++ b/nameko/exceptions.py
@@ -44,6 +44,29 @@ def remote_error(exc_path):
     to be deserialized to decorated exception instance, rather than
     wrapped in ``RemoteError``.
 
+    If a remote service raises an error during an RPC call this decorator
+    can be used to map the remote exception to a local one and handle it
+    as normal::
+
+        @remote_error('service_two.NotFound')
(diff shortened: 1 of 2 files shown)


davidszotten:
hey, thanks for the pr. i like the idea, though we may want to think about how we identify remote exceptions. while "module path" works pretty well for builtin exceptions, i'm not sure if we want to make the internal module structure of a service part of its api (see e.g. `@remote_error('test.test_remote_error.RemoteError')` from your test)

_maybe_ `<service name>.<exception type>`? (though that might have other issues)

thoughts?

iky:
Hi David, thanks for the quick reply.

You are right, addressing errors by service name instead of module path is better. RP calls are addressed by service and method names, the `remote_error` should address exceptions the same way. Maybe the decorator should live in (be imported from) rpc module as well?

The path quite works in our current setup where exceptions always live in the same place of the service package (`myservice.exceptions.NotFound`), but that does not need to be always the case and mainly it brings the contract to the internal module structure of the service as you mentioned.

```python
from nameko.rpc import remote_error

@remote_error('service-a', 'NotFound')
class ItemNotFound(LookupError):
    pass
```

mattbennett:
`@remote_error('<service_name>', '<exc_type>')` seems straightforward and nicely namespaced to me.

This decorator could also be used for non-RPC request-response type entrypoints so I'm not sure the `rpc` package is the right place for it. Perhaps a top-level `serialization` package? The `safe_for_serialization` helper that currently lives in `nameko.exceptions` also needs a new home.

davidszotten:
just thinking about this a little more, we pass exception args along as well, don't we? maybe instead of decorating local classes, we could (/should?) just supply a list of exception types we want re-hydrated, and have the framework generate exception classes. they could then also have a common superclass (maybe some subclass of remoteerror) 

still a half-formed thought but wanted to mention it

what non-rpc req/response entrypoints did you have in mind @mattbennett ?

mattbennett:
@davidszotten there isn't anything specific that already exists, but it's easy to imagine a request-response type entrypoint that might want to map remote errors to specific exceptions in the client.

We do pass exception args, so I think this approach assumes the signatures are compatible (which should probably be tested/documented here, if we don't change the implementation). Having a common superclass and (presumably) guaranteeing that the remote error re-hydrates successfully sounds nice.

PR #466, opened 2017-08-22
Import lint

mattbennett:
Adds linting for imports as discussed here: https://github.com/nameko/nameko/pull/456#discussion_r130462705

### DIFF
--- a/nameko/cli/run.py
+++ b/nameko/cli/run.py
@@ -1,5 +1,8 @@
 from __future__ import print_function
 
+import eventlet
+eventlet.monkey_patch()  # noqa (code before rest of imports)
+
 import errno
 import inspect
 import logging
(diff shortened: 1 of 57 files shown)


nameko/cli/run.py
```diff
@@ -12,9 +12,11 @@
 import signal
 import sys
 
+import eventlet
```
davidszotten: again?

tox.ini
```diff
@@ -37,6 +37,7 @@ commands =
     lib: pip install --editable .[dev]
     lib: make flake8
     lib: make pylint
+    lib: make imports
```
davidszotten: maybe check timings but this is probably fastest (or second after flake8) so may be worth running before the slower pylint

nameko/cli/run.py
mattbennett: Whoops!

tox.ini
mattbennett: It's about 3x faster than flake8 actually. Have rearranged it

mattbennett:
This will create conflicts in the open PRs, although running
```
AUTO_FIX_IMPORTS=1 make imports 
```
should get something back to a state that git can auto-merge.

If you're happy with https://github.com/nameko/nameko/pull/440 and https://github.com/nameko/nameko/pull/465 it might be best to land those first though?

davidszotten:
not had a chance to finish those yet, sorry

Issue #359, opened 2016-09-14
ResourceLocked: Queue.declare: (405) RESOURCE_LOCKED - cannot obtain exclusive access to locked queue 'rpc.reply-standalone_rpc_proxy' ...


jessepollak:
We're having on-and-off connectivity issues with RabbitMQ & nameko and occasionally we're seeing a `ResourceLocked` error bubbling up from inside kombu.

```
>>> n.rpc.browser_session_service.create()
^CTraceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/venv/local/lib/python2.7/site-packages/nameko/rpc.py", line 352, in __call__
    return reply.result()
  File "/venv/local/lib/python2.7/site-packages/nameko/rpc.py", line 333, in result
    self.resp_body = self.reply_event.wait()
  File "/venv/local/lib/python2.7/site-packages/nameko/standalone/rpc.py", line 57, in wait
    self.queue_consumer.get_message(self.correlation_id)
  File "/venv/local/lib/python2.7/site-packages/nameko/standalone/rpc.py", line 163, in get_message
    self._setup_consumer()
  File "/venv/local/lib/python2.7/site-packages/nameko/standalone/rpc.py", line 93, in _setup_consumer
    maybe_declare(self.queue, channel)
  File "/venv/local/lib/python2.7/site-packages/kombu/common.py", line 113, in maybe_declare
    return _maybe_declare(entity, declared, ident, channel)
  File "/venv/local/lib/python2.7/site-packages/kombu/common.py", line 120, in _maybe_declare
    entity.declare()
  File "/venv/local/lib/python2.7/site-packages/kombu/entity.py", line 522, in declare
    self.queue_declare(nowait, passive=False)
  File "/venv/local/lib/python2.7/site-packages/kombu/entity.py", line 548, in queue_declare
    nowait=nowait)
  File "/venv/local/lib/python2.7/site-packages/amqp/channel.py", line 1259, in queue_declare
    (50, 11),  # Channel.queue_declare_ok
  File "/venv/local/lib/python2.7/site-packages/amqp/abstract_channel.py", line 69, in wait
    return self.dispatch_method(method_sig, args, content)
  File "/venv/local/lib/python2.7/site-packages/amqp/abstract_channel.py", line 87, in dispatch_method
    return amqp_method(self, args)
  File "/venv/local/lib/python2.7/site-packages/amqp/channel.py", line 243, in _close
[...]

PR #468, opened 2017-09-07
Fix flakey tests

mattbennett:
Fixes a handful of tests which fail occasionally on Travis.


mattbennett:
Unfortunately these are now failing because Travis has updated their default linux image, which changes the RabbitMQ version.

The changes in https://github.com/rabbitmq/rabbitmq-management/issues/236 (released in 3.6.7) seem to have added a lag to the API so a bunch more tests fail. We _could_ add retries and timeouts, but it will slow down our test suite. I think there are better ways to write the tests that are failing.

In the mean time I will try to get Travis to run with the old images, and we can cherry-pick that change into the current PRs.

test/standalone/test_rpc_proxy.py
```diff
@@ -222,85 +223,119 @@ def test_multiple_calls_to_result(container_factory, rabbit_config):
...
-            # make another call that will trigger the disconnection;
-            # expect the blocking proxy to raise when the service reconnects
-            with pytest.raises(RpcConnectionError):
-                proxy.method("hello")
+        # make an async call that will block,
+        # wait for the worker to have spawned
+        with wait_for_call(container, 'spawn_worker'):
+            res = rpc_proxy.service.method.call_async('msg1')
+
+        # fake a socket error that causes the proxy to reconnect
```
davidszotten: the original test actually disconnected. not sure about the fidelity of just mocking the expected side-effect

mattbennett: The original test isn't particularly representative of real behaviour -- closing the connection from the broker API actually sends a consumer-cancel rather than yanking the connection.

We _could_ put toxiproxy in here, but it seems like overkill if we trust that kombu will throw ConnectionErrors when something goes awry with the connection.

davidszotten: How hard/annoying is it to add a toxiproxi test? Not so keen on relying on mocks here, esp as we’re looking at upgrading or even removing kombu

mattbennett: That's a good point. It's probably not too hard. Will try to add it

mattbennett: Good call on adding this -- the behaviour on a real disconnection is different, because kombu keeps the state of the connection and we use it to determine whether the proxy has been previously stopped.

I've updated the test, although the whole thing will be made moot by the fix for https://github.com/nameko/nameko/issues/359

mattbennett:
@davidszotten this is ready for another round when you have a chance.

davidszotten:
this was failing this morning when i checked. i guess you re-ran the job (so we still have some flaky tests :(

test/standalone/test_rpc_proxy.py
```diff
@@ -249,23 +252,25 @@ def method(self, arg):
...
-            with pytest.raises(RpcConnectionError):
-                res.result()
+        finally:
+            # reconnect
+            toxiproxy.enable()
 
-        # proxy should work again afterwards
-        block.send()
-        assert rpc_proxy.service.method("msg3") == "msg3"
+        # proxy will not work afterwards
```
davidszotten: we changing the behaviour?

mattbennett:
Caught red-handed :-s

There is a second simultaneous read error being thrown in the websocket test_client_closing_connection on py35-oldest-lib, which must have been there for a while. I haven't had chance to investigate further, and would like to get this tranche of flakey fixes out first (and the fix for #359, which is sitting on top of it)

test/standalone/test_rpc_proxy.py
mattbennett: No, it's the test.

When we used the management API to close the connection, it closed the existing connection but didn't actually prevent the consumer reconnecting here

https://github.com/nameko/nameko/blob/fb7560e08eaf14fbe1684abf6939efbd7aa9fbca/nameko/standalone/rpc.py#L168-L170

Now, toxiproxy remains disabled until after the consumer fails to reconnect and bubbles out the `socket.error`, so then the following triggers on next use:

https://github.com/nameko/nameko/blob/fb7560e08eaf14fbe1684abf6939efbd7aa9fbca/nameko/standalone/rpc.py#L57-L65

There is a (slightly ugly) hook we can use the re-enable the connection between the first and second attempts, so I will add a test that does that.

PR #469, opened 2017-09-07
Travis build stages

mattbennett:
Use build stages to avoid performing static analysis in every environment


davidszotten:
are pyflakes and pylint py2/3 independent?

mattbennett:
I assume so, although not really sure. We could easily run both 2x and 3x in parallel.

mattbennett:
Looks like it does make a difference (https://stackoverflow.com/questions/23052637/specify-which-python-version-pylint-should-evaluate-for) so I've added another step to the static build stage

.travis.yml
```diff
@@ -28,46 +28,72 @@ cache:
 
 matrix:
   include:
-    - python: 2.7
+    - stage: static
+      python: 2.7
+      env: TOX_ENV=py27-static
+    -
```
davidszotten: not sure i understand the convention of blank lines in here

mattbennett: It is a bit weird. Was following https://blog.travis-ci.com/2017-05-11-introducing-build-stages .

Re-reading that suggests that they're _optional_ though, so I will put them back in again.

PR #476, opened 2017-09-22
Refactor constants into constants.py

bobh66:
This allows for standalone/events.py to be imported into a
Jython environment that does not support eventlet.

Also updated .gitignore to remove dev-specific directories and a duplicate .DS_Store entry.

I needed to generate standalone nameko events from a Jython application, and when I tried to import nameko.standalone.events it also pulled in nameko.messaging, which pulled in eventlet, which doesn't work in Jython.

By refactoring all of the constants into constants.py I was able to import the module into Jython without pulling in eventlet.

The NON_PERSISENT constant turned out to be unused in publish.py so I corrected the spelling and placed it in constants.py.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,6 @@
 .DS_Store
-.codeintel
 .coverage
-.DS_Store
-.idea
 *.pyc
-*.sublime-*
 _py_cache
 docs/build
 
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .codeintel
 .coverage
 .DS_Store
+.idea
 *.pyc
 *.sublime-*
 _py_cache
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,18 @@ sudo: false
 language: python
 
 services:
-    - rabbitmq
+  - docker
 
 before_install:
+    # start rabbitmq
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -2,6 +2,7 @@
 .codeintel
 .coverage
 .DS_Store
+.idea
```
mattbennett: I prefer to keep these developer's choice ignores out of here, and in a global gitignore instead. Otherwise we'll end up with an ever growing list.

The .codeintel and sublime ones snuck in at some point!

bobh66:
Thanks Matt, will do.  I'll remove the codeintel and sublime too?

Any idea why the build fails?

mattbennett:
There is a problem with our build environment on Travis at the moment, which is why the tests are failing so spectacularly. I will hopefully be able to get that fixed in the next few days

mattbennett:
The build fails because the machine images were upgraded. It's now using a different version of rabbitmq, whose management api is less responsive, which causes some bogus test failures.

bobh66:
Updated with .gitignore changes.  Any other comments?

bobh66:
Thanks Matt!

mattbennett:
@bobh66 the test fixes for the new Travis infrastructure are now in master. Do you want to bring this branch up to date so CI can pass?

bobh66:
Thanks Matt!  I appreciate the quick turnaround.  I realized that this also lets me send events from my gevent-based code, which will be handy.

PR #481, opened 2017-10-09
Close connections when deleting vhost during tests

timbu:
We noticed when running our service tests, that even though each vhost is being deleted, there are connections left behind for those vhosts.
It appears these are connections for each of the producers started during the test run.

This fix ensures all vhost connections are closed when the vhost is deleted (in a brute force way).  Possibly the full fix is to give producers a full extension lifecycle like other entrypoints, so they get torn down properly?

It also adds a retry to ensure the vhost is deleted.

### DIFF
--- a/nameko/testing/pytest.py
+++ b/nameko/testing/pytest.py
@@ -96,10 +96,15 @@ def rabbit_manager(request):
 
 @pytest.yield_fixture(scope='session')
 def vhost_pipeline(request, rabbit_manager):
+    from collections import Iterable
     from six.moves.urllib.parse import urlparse  # pylint: disable=E0401
     import random
+    import socket
     import string
+    from kombu.pools import connections
     from nameko.testing.utils import ResourcePipeline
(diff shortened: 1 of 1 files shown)


nameko/testing/pytest.py
```diff
@@ -115,9 +120,18 @@ def create():
         )
         return vhost
 
+    @retry(for_exceptions=(HTTPError, socket.timeout), delay=1, max_attempts=9)
```
mattbennett: Is there a significance to `max_attempts=9`? (seems rather a lot)

timbu: not really -other than to give it around 10 secs :)

timbu:
The old connections (and vhosts) were all removed at the end of the test run - so normally you wouldn't notice. However, on services with many integration tests, rabbitmq would run out of file-descriptors. This could be worked around by increasing the number allowed, but would require additional config when building.

PR #491, opened 2017-11-14
Fix logging fail when exception has unicode characters

eugeneRover:
Hello,

When my service raises en error (with some non-ascii characters in message), exception and its trace is not visible, because logging cannot process such message:
```
<WorkerContext [PS.storeFollowingEvent] at 0x7f0e141101d0>
  File "/usr/lib/python2.7/logging/__init__.py", line 859, in emit
    msg = self.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 732, in format
    return fmt.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 471, in format
    record.message = record.getMessage()
  File "/usr/lib/python2.7/logging/__init__.py", line 335, in getMessage
    msg = msg % self.args
UnicodeDecodeError: 'ascii' codec can't decode byte 0xd0 in position 0: ordinal not in range(128)
Logged from file containers.py, line 393
127.0.0.1 - - [14/Nov/2017 19:29:07] "POST /storeFollowingEvent/252525252525/99999999999/1?moment=234567891 HTTP/1.1" 500 204 0.015099
```
Actionally, message is cyrillic:
```
ОШИБКА:  нулевое значение в столбце "id" нарушает ограничение NOT NULL
DETAIL:  Ошибочная строка содержит (1977-06-08 07:44:51, 99999999999, t, null).
```

When I add ```decode('utf-8')```, problem is away, exception is printed OK.

Regards, 
Evgeniy

### DIFF
--- a/nameko/containers.py
+++ b/nameko/containers.py
@@ -387,7 +387,7 @@ class ServiceContainer(object):
                 with _log_time('ran handler for %s', worker_ctx):
                     result = method(*worker_ctx.args, **worker_ctx.kwargs)
             except Exception as exc:
-                _log.info('error handling worker %s: %s', worker_ctx, exc,
+                _log.info('error handling worker %s: %s', worker_ctx, exc.message.decode('utf-8'),
                           exc_info=True)
                 exc_info = sys.exc_info()
 


davidszotten:
Hi,

Thanks for the contribution! However, i'm a bit confused, since that file includes `from __future__ import unicode_literals`, so that formatting string should already be unicode, not ascii.

What version of nameko and python is this on? Could you post a minimal example to reproduce the error please?

eugeneRover:
I use Python 2.7.9 and nameko 2.8.0.
Here is reproducing code (reproduce.py) :
```
#!/usr/bin/python

# If I remove below line, UnicodeDecodeError goes away
from __future__ import absolute_import, unicode_literals

import logging 

logging.basicConfig()

try:
	# I've got following utf-8 encoded string in exception from Postgres driver.
	# Actionally, it is cyrillic characters 
	ba = bytearray([208,158,208,168,208,152,208,145,208,154,208,144])
	raise Exception(ba)
except Exception as ex:
	logging.getLogger().error ('error %s', ex, exc_info=True)
```

davidszotten:
Right, now i'm with you. Had misread decode vs encode.

I think the best way to stay sane with regards to unicode is what e.g. ned batchelder calls the ["unicode sandwitch"](https://nedbatchelder.com/text/unipain.html). Basically, you want to decode your bytes as early as possible (and if required, encode back as late as possible)

In your example, since you know you are getting back bytes representing a utf8 encoded string,  i think you want to `raise Exception(ba.dcode('utf-8'))`

mattbennett:
Closing this as unrelated to Nameko. Hope you fixed your issue @eugeneRover

PR #493, opened 2017-11-22
add a command to output parsed config after env var substitution

mattbennett:
Useful for debugging services that have complicated config files with many environment variable substitutions.


nameko/cli/commands.py
```diff
@@ -46,6 +46,31 @@ def main(args):
...
 
+class Config(Command):
+    """ Output as YAML string the configuration that would be passed to a
+    service.
+
+    Useful for viewing config files that load values from environement
+    variables.
+    """
+
+    name = 'config'
```
davidszotten: how about calling this `show-config` or similar to make it more clearly distinct from `--config`

mattbennett: That is a much better name!

davidszotten:
nice idea! like it

nameko/cli/config.py
```diff
@@ -1,11 +0,0 @@
-from __future__ import print_function
```
davidszotten: old file removed but new file not added?

mattbennett: Whoops!

test/cli/test_config.py
```diff
@@ -0,0 +1,41 @@
+from textwrap import dedent
```
davidszotten: `test_show_config`?

test/cli/test_config.py
```diff
@@ -0,0 +1,41 @@
...
+    config = tmpdir.join('config.yaml')
+    config.write("""
+        FOO: ${FOO:foobar}
+        BAR: ${BAR}
+    """)
+
+    parser = setup_parser()
+    setup_yaml_parser()
+    args = parser.parse_args([
+        'config',
```
davidszotten: `show-config`...

mattbennett:
Sorry for the back and forth. All comments addressed now, and I've checked the test actually pass ;)

mattbennett:
Thanks. Will add changelog entry when I cut a release

PR #496, opened 2017-12-09
try dropping log-level pytest plugin arg

davidszotten:
conflicts with new pytest. for old pytest can be restored by adding
pytest-catchlog (now part of pytest; hence the conflict)


peterclearmetal:
For future reference and googlability the error I got running pytest with Nameko <= 2.8.1 installed is:

```
argparse.ArgumentError: argument --log-level: conflicting option string(s): --log-level
```

And the related pytest issue is [eisensheng/pytest-catchlog#36](https://github.com/eisensheng/pytest-catchlog/issues/36).

PR #499, opened 2018-01-04
Enable multi-method URLs

geoffjukes:
Sometimes a URL method shares functional code, and the http method is used further downstream. This simple change enables a DRY approach to these types of URLs.

### DIFF
--- a/nameko/web/handlers.py
+++ b/nameko/web/handlers.py
@@ -23,7 +23,7 @@ class HttpRequestHandler(Entrypoint):
         super(HttpRequestHandler, self).__init__(**kwargs)
 
     def get_url_rule(self):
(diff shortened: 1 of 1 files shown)


mattbennett:
Thanks @geoffjukes for this PR!

Can you please also add a test for this? e.g. add a new method here

https://github.com/nameko/nameko/blob/a0c61ad82650a04b2aa24264974a189fff6bf911/test/web/test_http_handler.py#L11-L23

and verify that it can be called successfully with multiple methods.

This change probably deserves a line in the docs somewhere too.

geoffjukes:
I have no idea how to update Pull requests, so I'll just resubmit this. My apologies.

PR #501, opened 2018-01-12
Restrict eventlet

mattbennett:


davidszotten:
does https://github.com/nameko/nameko/commit/6aa00da824c6e138a249c27cf50abdc47ff72dd9 only affect travis?

mattbennett:
6aa00da will apply if you run your tests locally using tox as well.

davidszotten:
so that fix is now out, but we can't use it because the http is broken instead?

davidszotten:
not sure i understand the reason for the tox change

mattbennett:
Yes :(

I've had some discussion about the breaking change in https://github.com/eventlet/eventlet/issues/420 but honestly got a bit lost with it.

mattbennett:
The tox change is required because we were no longer actually testing against eventlet master. See https://travis-ci.org/nameko/nameko/builds/328005814?utm_source=github_status&utm_medium=notification

Before the change tox was executing the following commands:

```
pip install --upgrade https://github.com/eventlet/eventlet/archive/master.zip (from mastereventlet match)
pip install --editable .[dev] (from lib match)
make test_lib (from lib match)
```

`pip install --editable .[dev]` _after_ `pip install ...master.zip` was pulling eventlet back down to the pinned version

PR #504, opened 2018-01-15
Expiring rpc reply queues

mattbennett:
On top of #468. Fixes the problems reported in #359 and makes the service-to-service RPC proxy more tolerant of disconnections.

Unfortunately the standalone RPC proxy still retains some unhelpful behaviour and can't be used after more than one reconnection attempt has failed. Fixing that requires a rewrite and I would rather do that as a separate refactor.

Also note that this change will result in hanging RPC requests if the proxy is disconnected for longer than the reply queue expiry time (5 minutes, perhaps should be longer). To mitigate hanging requests we could implement RPC timeouts, which I would like to do as a separate feature.


davidszotten:
5 minutes sounds like quite a long time to be disconnected. how did you choose it?

regarding hanging rpc requests if the reply queue has been deleted, would this be possible to detect (e.g. on reconnection, check if reply queue still exists)?

mattbennett:
5 minutes is fairly arbitrary. Long enough to cover extended downtime, but not so long that the broker is full of dead queues.

Re: hanging requests; yes, you can tell whether a queue already existed on declaration, so we could totally do this. I will add.

mattbennett:
@davidszotten update on this branch...

I've added a couple of tests for the desired behaviour of raising if we detect we've been disconnected long enough to lose a reply. Unfortunately the implementation is not possible for a couple of reasons:

1. In the standalone proxy we can't even _test_ the behaviour because the current implementation reconnects immediately on disconnection, without retry and only twice, which doesn't take long enough for the broker to reap the expired queue.

2. In the service proxy, the queue declaration (with its return value that states whether the already queue existed or not) is inside the consumer creation, which delegated to the queue consumer. So the reply listener doesn't have an appropriate hook to know whether its outstanding replies will be answered or not.

Fortunately, I have a couple of branches already in the works that fix these problems:

1. https://github.com/nameko/nameko/compare/master...mattbennett:rpc-refactor?expand=1

In this branch, the standalone proxy has been refactored to use the normal `ConsumerMixin` pattern. This means it's possible to attempt to reconnect for long enough that the reply queue expires.

2. https://github.com/nameko/nameko/compare/master...mattbennett:remove-queue-consumer?expand=1

In this branch, the QueueConsumer is completely removed in favour of making each AMQP consumer (RpcConsumer, ReplyListener, EventHandler, Consumer) implement `ConsumerMixin` themselves. The refactor isn't finished yet, but it would give the ReplyListener the hook it needs to determine whether the queue existed on reconnection.

There would be a lot to review here and the and the `remove-queue-consumer` refactor isn't ready yet.

We can reduce the risk of hanging RPC requests by increasing the expiry timeout to something quite large (e.g. 30 minutes) so that we can release these fixes without waiting for the subsequent refactors to land.

mattbennett:
@davidszotten The PRs listed above address the first problem in my previous comment. They are ready to be reviewed. In those branches the _standalone_ RPC proxy will raise if its reply queue has been expired with pending replies while the proxy was disconnected. The service proxy is still unable to detect it until the queue consumer is refactored out (still a work in progress)

In the interest of reducing the amount of unmerged changes I'm going to merge this PR now. If you're OK with it I think we should also release this even before the other branches are ready -- rationale being that this fixes a known problem that can take down services, which is more bad than the risk of some stuck RPC requests.

PR #505, opened 2018-01-17
Enable multi-method URL declarations

geoffjukes:
Rather than specifying multiple (duplicate) URLs, this change allows specifying a CSV for the methods allowed.

For example:

```
    @http('GET', '/<string(length=4):book_id>/publish/<feed_name>')
    @http('PUT', '/<string(length=4):book_id>/publish/<feed_name>')
```
becomes
```
    @http('GET,PUT', '/<string(length=4):book_id>/publish/<feed_name>')
```
This is very handy when the methods share the same functional code, and reply on the HTTP method for augmentation (for example, 'Do it' vs 'Show me it')

This is a change is backwards compatible, as a 'split' always greats a list - even if there are no matches.

`"GET".split(',')` => `["GET"]`
`"GET,PUT".split(',')` => `["GET", "PUT"]`

### DIFF
--- a/docs/built_in_extensions.rst
+++ b/docs/built_in_extensions.rst
@@ -48,6 +48,8 @@ HTTP
 
 The HTTP entrypoint is built on top of `werkzeug <http://werkzeug.pocoo.org/>`_, and supports all the standard HTTP methods (GET/POST/DELETE/PUT etc)
 
+The HTTP entrypoint can specify multiple HTTP methods for a single URL as a comma-separated list. See example below.
+
 Service methods must return one of:
 
 - a string, which becomes the response body
--- a/docs/examples/http.py
+++ b/docs/examples/http.py
@@ -13,3 +13,7 @@ class HttpService:
     @http('POST', '/post')
     def do_post(self, request):
         return u"received: {}".format(request.get_data(as_text=True))
+
+    @http('GET,PUT,POST,DELETE', '/multi')
+    def get_method(self, request):
+        return request.method
--- a/docs/examples/http.py
+++ b/docs/examples/http.py
@@ -13,3 +13,7 @@ class HttpService:
     @http('POST', '/post')
     def do_post(self, request):
         return u"received: {}".format(request.get_data(as_text=True))
+
+    @http('GET,PUT,POST,DELETE', '/multi/<value>')
+    def get_method(self, request, value):
+        return u"{}: {}".format(request.method, value)
--- a/docs/examples/http.py
+++ b/docs/examples/http.py
@@ -15,5 +15,5 @@ class HttpService:
         return u"received: {}".format(request.get_data(as_text=True))
 
     @http('GET,PUT,POST,DELETE', '/multi')
-    def get_method(self, request):
+    def do_multi(self, request):
         return request.method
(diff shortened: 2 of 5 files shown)


geoffjukes:
@mattbennett OK, so making additional changes to an open pull request is just magical - I've never done it before, so I was expecting it to be a bit more involved/complicated.

geoffjukes:
@mattbennett The build fails on the testing_examples. Pretty sure this is because the updated example needs the updated code, and the testing_examples installs nameko 2.8.3 - so no update.

I can remove the example to make the tests pass, if that is preferred.

mattbennett:
@geoffjukes actually that test is failing because the new method added in the example shadows the existing one (they're both called `get_method`)

FYI test_examples does install the current codebase version (`pip install -e .[dev]`) but it is reported as `2.8.3` because that's the version in setup.py at the moment.

It occurred to me just now that you can actually achieve your desired behaviour just by stacking the decorators:

``` python
@http('GET', '/resource')
@http('POST', '/resource')
def resource(self, request):
    pass
```

I think that's actually more readable, and it doesn't need any new code. What do you think?

geoffjukes:
Oh my word, I cannot believe I missed that. Thanks for not mocking me!

I currently stack the decorators, but it gets pretty ugly. With more complex URLs, it makes changes more involved. It's your choice of course, but for me, I think this would be a very useful change.

geoffjukes:
Example:

```
@http('PUT', '/<string(length=4):book_id>/deliver')
@http('GET', '/<string(length=4):book_id>/deliver')
@http('PATCH', '/<string(length=4):book_id>/deliver')
@http('DELETE', '/<string(length=4):book_id>/deliver')
    def deliver(self, request, book_id=False):
        return do_stuff(book_id, request.method)
```
vs
```
@http('PUT,GET,PATCH,DELETE', '/<string(length=4):book_id>/deliver')
    def deliver(self, request, book_id=False):
        return do_stuff(book_id, request.method)
```

mattbennett:
Ha, we've all done it! Sometimes you can't see the wood for the trees.

I see what you mean about being ugly. Your `do_stuff` method is interesting though...

You pass the `request.method` in, so does that mean you split it back out on the other side? e.g.

``` python
@http('PUT,GET,PATCH,DELETE', '/<string(length=4):book_id>/deliver')
def deliver(self, request, book_id=False):
    return do_stuff(book_id, request.method)

def do_stuff(book_id, request.method):
    if request.method == "GET":
        # do GET stuff
    elif request.method == "POST":
        # do POST stuff
```

If so, why not do:

``` python
deliver_resource = '/<string(length=4):book_id>/deliver'  # for easier refactoring

@http('GET', deliver_resource)
def get_delivery(self, request, book_id=False):  # or a more descriptive name
   # do GET stuff

@http('POST', deliver_resource)
def submit_delivery(self, request, book_id=False):
   # do POST stuff
```
?

geoffjukes:
Yes, I do check the method later. The example was over simplified. More often I find my endpoints share 99% of the same code, changing only the return, for example a GET returns the computed data, but a PUT submits it to a secondary process. These methods can be rather long.

mattbennett:
OK cool, I see. Well this is a tiny amount of logic that is obviously helpful in some situations. I will merge and release it in the next cut.

Thank you for the contribution and the discussion!

geoffjukes:
Thanks, and no worries!

PR #507, opened 2018-01-28
Fix call id properties

fobiols:
Currently, `immediate_parent_call_id` **doesn't fetch** the parent call_id from the stack but rather the **original** call_id.

This fix adds a new property `origin_call_id` with the same behaviour as the current `immediate_parent_call_id` and fixes the `immediate_parent_call_id` to do as it should.


mattbennett:
@fobiols this is failing because of https://github.com/timothycrosley/isort/issues/676

Please add a closing newline to test/test_call_id_stack.py

mattbennett:
Thank you @fobiols!

Issue #511, opened 2018-02-13
RecoverableConnectionError raised on message.ack()


mattbennett:
Discussed on the mailing list [[1]](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/nameko-dev/Hmz659s4dBY)[[2]](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/nameko-dev/n3PSBuvqIGM) but raising here for visibility.

When a consumer connection to RabbitMQ is lost while an AMQP entrypoint is mid-execution, the broker reclaims the unack'd message but the worker continues. We attempt to acknowledge the message when the execution finally completes, and this raises a `RecoverableConnectionError` because the message's connection is closed. We naively retry on this error, but even if a new connection is established, the channel for the particular message will never recover, which ultimately results in the container dying with the `RecoverableConnectionError` exception.

This bug was probably introduced in 2.5.4.

mattbennett:
Fixed in #512

PR #512, opened 2018-02-13
Message ack regression

mattbennett:
Fix for #511.

Includes some commits cherry-picked from #504:

f5901f3a1d80fdaea8bf4cb16fd1ff8541fd6a86: standard toxiproxy tests for RpcConsumer
8841dd1464b56906e0d28c967ac58644daaeb381 and 891a91c94fb20e4328cf6a31765f3f8982d1e823: fix for hanging toxiproxy process

### DIFF
--- a/nameko/messaging.py
+++ b/nameko/messaging.py
@@ -8,7 +8,6 @@ from functools import partial
 from logging import getLogger
 
 import six
-from amqp.exceptions import RecoverableConnectionError
 from eventlet.event import Event
 from kombu import Connection
 from kombu.common import maybe_declare
@@ -25,7 +24,6 @@ from nameko.exceptions import ContainerBeingKilled
 from nameko.extensions import (
     DependencyProvider, Entrypoint, ProviderCollector, SharedExtension
 )
-from nameko.utils.retry import retry
 
 
 _log = getLogger(__name__)
--- a/nameko/messaging.py
+++ b/nameko/messaging.py
@@ -310,12 +310,18 @@ class QueueConsumer(SharedExtension, ProviderCollector, ConsumerMixin):
 
         super(QueueConsumer, self).unregister_provider(provider)
 
-    @retry(for_exceptions=RecoverableConnectionError, max_attempts=3)
     def ack_message(self, message):
+        if not message.channel.connection:
+            # don't attempt to ack if the message connection is gone;
+            # the message will already have been reclaimed by the broker
+            return
         message.ack()
 
-    @retry(for_exceptions=RecoverableConnectionError, max_attempts=3)
     def requeue_message(self, message):
+        if not message.channel.connection:
+            # don't attempt to requeue if the message connection is gone;
+            # the message will already have been reclaimed by the broker
+            return
         message.requeue()
 
     def _cancel_consumers_if_requested(self):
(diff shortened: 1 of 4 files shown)


noisyboiler:
Hi @mattbennett 
We'd love to see this land as the issue is a blocker for us getting nameko into production.
What are your expectations for getting this into a new release?
thanks

mattbennett:
@noisyboiler I'm happy to release as soon as it's been through review. Some critical eyes would be helpful, thanks. I would also like @davidszotten's approval before landing it.

nameko/messaging.py
```diff
@@ -310,12 +308,18 @@ def unregister_provider(self, provider):
 
         super(QueueConsumer, self).unregister_provider(provider)
 
-    @retry(for_exceptions=RecoverableConnectionError, max_attempts=3)
     def ack_message(self, message):
+        if not message.channel.connection:
```
davidszotten: don't you have a race here? could we try and catch+ignore the exception?

test/conftest.py
```diff
@@ -161,6 +161,12 @@ def reset(self):
     yield controller
     controller.reset()
 
+    # delete proxy
+    # allow some grace period to ensure we don't remove the proxy before
+    # other fixtures have torn down
+    resource = 'http://{}/proxies/{}'.format(toxiproxy_server, proxy_name)
+    eventlet.spawn_after(10, requests.delete, resource)
```
davidszotten: does this mean that last one of these might keep the test run from finishing for 10 seconds?

nameko/messaging.py
mattbennett: Race between what?

Try/catch is problematic because the exception raised is `RecoverableConnectionError`, which "connection already closed" just as a string. It's worth noting that the connection is not actually recoverable from here (we used to retry on these errors, and even after a new connection is established the ack is still attempted through the dead channel) and even if it _was_, the broker will already have reclaimed the message.

The logic here is "if the channel is closed, the message will have been reclaimed, so do nothing".

test/conftest.py
mattbennett: No, a process won't wait for greenthreads to finish before exiting.

This cleanup really isn't required because we stop the whole toxiproxy server at the end of the test run. I added it while debugging a hanging toxiproxy (which turned out to be http://tanin.nanakorn.com/blogs/373) thinking that the problem _might_ be the growing number of proxies that we add and never remove.

davidszotten:
Race may be wrong word, but connection might close between checking and acking

nameko/messaging.py
mattbennett: > Race may be wrong word, but connection might close between checking and acking

(GitHub email didn't reply to my comment)

Yes I am with you. I would like to keep the conditional to show intent, but will add a try/catch to make sure we don't blow up if we lose the race.

nameko/messaging.py
```diff
@@ -310,13 +309,23 @@ def unregister_provider(self, provider):
 
         super(QueueConsumer, self).unregister_provider(provider)
 
-    @retry(for_exceptions=RecoverableConnectionError, max_attempts=3)
     def ack_message(self, message):
-        message.ack()
+        # only attempt to ack if the message connection is alive;
+        # otherwise the message will already have been reclaimed by the broker
```
davidszotten: does the broker reclaim immediately? (i.e. no grace period?)

mattbennett: As soon as it detects the connection is closed, yes. I don't think a grace period would make any difference.

davidszotten: didn't mean for us to leave a grace period, but if the broker did (if so, we could try and just detect failures) but this is probably fine

test/conftest.py
mattbennett: I guess not. Will remove

noisyboiler:
much appreciated chaps. 👍

PR #521, opened 2018-03-09
workaround for https://github.com/celery/kombu/issues/838

mattbennett:
If we don't explicitly set the timeout we may inherit the temporary default.

This can cause failures when testing services that use AMQP and HTTP entrypoints with the builtin pytest helpers. The line at https://github.com/nameko/nameko/blob/master/nameko/testing/pytest.py#L120 can execute while a new service is starting, triggering the bug, and the listening socket only waits 0.1s for a connection.

### DIFF
--- a/nameko/cli/run.py
+++ b/nameko/cli/run.py
@@ -100,6 +100,8 @@ def setup_backdoor(runner, port):
             'This would kill your service, not close the backdoor. To exit, '
             'use ctrl-c.')
     socket = eventlet.listen(('localhost', port))
+    # work around https://github.com/celery/kombu/issues/838
+    socket.settimeout(None)
     gt = eventlet.spawn(
         backdoor.backdoor_server,
         socket,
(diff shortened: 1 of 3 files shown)


mattbennett:
There is probably a better workaround here.

I think the bug against kombu is valid, but the change added in https://github.com/nameko/nameko/pull/481 (which is the trigger for the hanging tests) has other problems too:

The kombu pool is shared at the process level and the vhost pipeline `destroy` runs in a background thread, not synchronously at the end of tests. So when tearing down from a previous test we may be closing publisher connections created by a currently running test. It is a little bit surprising that this hasn't caused other problems.

I think we need a different solution for the problem that #481 tries to address, as well as this workaround.

mattbennett:
The alternative discussed in #481 was just to increase the number of file descriptors on the rabbit instance, which now seems much more reasonable than trying to accommodate a particular test suite's requirements with a change to the underlying library. I will revert ecf6c58.

mattbennett:
@iky you're right, connection pools are per vhost. Not sure how I missed that. We use a different vhost per test, so the failure here is only a result of the kombu bug. I will revert the revert 😬

PR #530, opened 2018-04-02
Exclude anonymous cls

zhouxiaoxiang:
- Don't import class without `name` property  in order to ignore parent class.
- Skip normal class such as sqlalchemy.* directly in import_service.


zhouxiaoxiang:
@davidszotten 

This PR permits mix-in parent class as a normal class  and will not be added into runner.

PR #535, opened 2018-04-16
Multi-serialiser support

iky:
Current Nameko setup allows for having one serialisation format across the whole cluster. Serialisation format can be set by `serializer` config entry.

This change adds support for multiple serialisation formats:
* adds support for accepting multiple serialisation formats for built-in messaging, pub-sub and RPC consumers
* RPC replies in the same format as was the request
* ads the ability to define custom serializers in config

Two new config entries added. `ACCEPT` defines serialisation formats accepted by Nameko consumers:
``` yaml
ACCEPT:
    - json
    - yaml
````

`SERIALIZERS` sets up serialisers registry. Adding a custom serialiser:
``` yaml
SERILIZERS:
    myserializer:
        encode: mymodule.encode
        decode: mymodule.decode
        content_type: application:x-mymimetype
        content_encoding: utf-8
```

An example of a service using a custom XJSON format for serialising but accepting both, the built-in JSON and the custom XJSON:
``` yaml
serializer: json
ACCEPT:
    - json
    - xjson
SERIALIZERS:
    xjson:
        encode: xjson.encode
        decode: xjsom.decode
        content_type: application:x-json
```

With the setup above, a service would:
* publish events using JSON
* handle events serialized as both JSON and XJSON
* send RPC request using JSON
* accept RPC requests as both JSON and XJSON, replying to JSON RPC calls with JSON and XJSON RPC calls with XJSON


nameko/containers.py
```diff
@@ -168,6 +166,21 @@ def __init__(self, service_cls, config):
...
+            encoder = import_from_path(kwargs.pop('encoder'))
+            decoder = import_from_path(kwargs.pop('decoder'))
+            serialization.register(
+                name, encoder=encoder, decoder=decoder, **kwargs)
+
+        self.serializer = self.config.get(
+            SERIALIZER_CONFIG_KEY, DEFAULT_SERIALIZER)
+
+        self.accept = self.config.get(
+            ACCEPT_CONFIG_KEY, [self.serializer])
```
mattbennett: I think we need to check whether the specified serializer and accept values have been registered as serializers.

nameko/rpc.py
```diff
@@ -188,13 +188,17 @@ def send_response(self, result, exc_info):
         if exc_info is not None:
             error = serialize(exc_info[1])
 
+        # send response encoded the same way as was the request message
+        content_type = self.message.properties['content_type']
+        serializer = kombu.serialization.registry.type_to_name[content_type]
```
mattbennett: Reinforcing the comment above: we should check earlier than here that there is a registered serializer for this `content_type`.

test/test_serialization.py
```diff
@@ -208,3 +219,170 @@ def echo(self, arg):
...
+    echoer_config = rabbit_config.copy()
+    echoer_config[SERIALIZER_CONFIG_KEY] = 'json'
+    echoer_config[ACCEPT_CONFIG_KEY] = ['json', 'yaml']
+
+    echoer = container_factory(EchoingService, echoer_config)
+    echoer.start()
+
+    payload = {'spam': 'ham'}
+
+    # Forwarder serialiser is set to JSON and should send and receive JSON
```
mattbennett: Might be worth splitting this into multiple tests. If you put them all into a class you can do the setup in fixtures to avoid duplication

test/test_serialization.py
```diff
@@ -208,3 +219,170 @@ def echo(self, arg):
...
+    publisher_config[SERIALIZER_CONFIG_KEY] = 'yaml'
+    publisher = container_factory(PublishingService, publisher_config)
+    publisher.start()
+    publish(consumer, publisher)
+
+    msg = get_messages().pop()
+    assert '{spam: ham}' in msg['payload']
+    assert msg['properties']['content_type'] == 'application/x-yaml'
+
+    assert called.mock_calls == [call(payload), call(payload)]
```
mattbennett: Same here. I think separate tests for these two cases would be more readable (and more targeted if one fails). You might even be able to parametrize them

test/test_serialization.py
iky: Sure thing, parameterised.

test/test_serialization.py
iky: Sure thing, parameterised.

nameko/containers.py
iky: True, it should be checked and it should raise `ConfigurationError` with a clear message rather than `KeyErrors` later on and on various places.

nameko/rpc.py
iky: True, it should be checked and it should raise ConfigurationError with a clear message rather than KeyErrors later on and on various places.

A check added.

iky:
Thanks for the review @mattbennett!

I added multi-serialization support also to standalone RPC proxy and same serializer setup to standalone event publisher too. For that I extracted the serialization setup to its own module.

PR #536, opened 2018-04-25
RPC refactor

mattbennett:
First part of the RPC refactor mentioned in #504, rebased on top of the final state of that branch.

Main changes:

* Updating standalone RPC proxy to implement ConsumerMixin to collect replies
* Dropping the backwards compat shims class attributes on the `MethodProxy` class

This makes the standalone proxy more similar to the service-based one. It also means the standalone proxy will try to reconnect multiple times and that we can detect on reconnection whether any replies have been lost due to a reply queue expiring.


mattbennett:
Included in #542

PR #537, opened 2018-04-25
RPC refactor, continued

mattbennett:
As per #536 but with a few extra changes on top.

Extra changes:

* Replaces `ServiceProxy` and `MethodProxy` chain with a simpler single object (`Proxy`)
* Removes now unused `wait_for_worker_idle`

The diff is hard to follow on this change, but the end-state is easier to understand.

See https://github.com/mattbennett/nameko/compare/rpc-refactor...rpc-refactor-p2?expand=1 for a comparison of just the extras.


mattbennett:
Included in #542

PR #542, opened 2018-05-04
Remove queue consumer

mattbennett:
On top of #536 and #537, this PR removes the `QueueConsumer` shared extension.

The changes affect the `EventHandler` and `Consumer` entrypoints and the RPC `ReplyListener` and `RpcConsumer` extensions. These now maintain their own AMQP consumers rather than delegating to the `QueueConsumer`. This removes a lot of complexity and allows more granular configuration. The refactor also allows us to detect lost RPC replies due to queue expiry and that functionality has been added.

Additional changes included here:

1. Explicitly expose the `prefetch_count` via config, rather than using the `max_workers` value. 

2. Fixes a race condition introduced in https://github.com/nameko/nameko/commit/6f6182aff1457c22ffeed69e7aac79e4e5bf7a6f where we could raise a "queue expired" error when the proxy was first starting up.

3. Make the tests in tests/test_events.py more reliable and easier to understand.

Side-effects of this change are more connections to the RabbitMQ broker. In an ideal world, each entrypoint would maintain its own consumer but the channels and connections would be shared, but that requires more work and probably migration away from kombu. As implemented, we end up with a consumer, channel and connection for every queue from which we consume.


mattbennett:
@davidszotten this is ready for a review. I'm sorry it's so massive :/

mattbennett:
I got carried away and did some more refactoring here, so the interim #537 is not useful anymore. The RPC proxy interface that was added in that branch has been changed slightly.

I will add a summary all the changes between this branch to master in a comment here soon. They are quite significant so I would like to propose that after some rounds of review we release an alpha build for people to experiment with.

mattbennett:
As promised, a list of all the changes incorporated in this PR:

* Adds a new helper at `nameko.amqp.consume.Consumer`. Similar to the helper at `nameko.amqp.publish.Publisher`, we use it any time we want to consume messages from AMQP. It uses the kombu ConsumerMixin but kombu is mostly behind the abstraction now.

* Removes the `QueueConsumer`. All entrypoints, extensions and utilities that consume from AMQP now use the `nameko.amqp.consume.Consumer` helper to do so.

* Adds checks for replies lost in expired RPC reply queues that could not be implemented in #504.

* Simplifies the RPC implementation, replacing `ServiceProxy` and `MethodProxy` with single object (simply `Proxy`, could do with a better name) that is less coupled to the DependencyProvider implementation of the RPC proxy. This means the standalone RPC proxy implementation can be much simpler.

* Drops the graceful stopping of AMQP consumers on container kill (surplus to requirements since we stopped sharing a vhost between tests). The tests are now much faster and the `fast_teardown` fixture hack makes no difference, so has also been removed.

* Replaces fragile uses of the RabbitMQ Management API (which got very slow in 3.6.6) and inspects queues with an AMQP-based inspection.

* Allows Extensions that consume AMQP messages to be explicitly configured with a prefetch count, distinct from the worker count. There is a new configuration option to set this globally, and a value for `prefetch_count` can be passed directly when instantiating the extensions (with the exception of the @rpc entrypoint)

* Drops various deprecations that were slated for removal in 2.9.0.

I will put comments in the diff to explain the changes in each module.

nameko/amqp/consume.py
```diff
@@ -0,0 +1,110 @@
+from logging import getLogger
```
mattbennett: This is the new Consumer utility (obviously)

nameko/extensions.py
```diff
@@ -2,7 +2,6 @@
 
 import inspect
 import types
-import warnings
```
mattbennett: Removal of deprecated options now we're passed the 2.9.0 milestone

nameko/messaging.py
```diff
@@ -3,40 +3,47 @@
...
+    }
+
+
+def decode_from_headers(headers, prefix=HEADER_PREFIX):
+    return {
+        re.sub("^{}\.".format(prefix), "", key): value
+        for key, value in headers.items()
+    }
+
+
```
mattbennett: Split this into functions so they can be used without inheriting from `HeaderEncoder `

nameko/messaging.py
```diff
@@ -183,229 +147,11 @@ def publish(msg, **kwargs):
         return publish
 
 
-class QueueConsumer(SharedExtension, ProviderCollector, ConsumerMixin):
```
mattbennett: Removal of the QueueConsumer

nameko/messaging.py
```diff
@@ -449,12 +228,26 @@ def handle_message(self, body, message):
...
+
+        service_name = self.container.service_name
+        method_name = self.method_name
+
+        # TODO replace global worker pool limits with per-entrypoint limits,
+        # then remove this waiter thread
+        ident = u"{}.wait_for_worker_pool[{}.{}]".format(
+            type(self).__name__, service_name, method_name
+        )
+        self.container.spawn_managed_thread(spawn_worker, identifier=ident)
```
mattbennett: This workaround was present in the QueueConsumer and must now be implemented in every entrypoint that consumes AMQP messages

nameko/messaging.py
```diff
@@ -434,13 +180,46 @@ def handle_message(self, body):
...
+            'heartbeat', config.get(HEARTBEAT_CONFIG_KEY, DEFAULT_HEARTBEAT)
+        )
+        prefetch_count = self.consumer_options.pop(
+            'prefetch_count', config.get(
+                PREFETCH_COUNT_CONFIG_KEY, DEFAULT_PREFETCH_COUNT
+            )
+        )
+        accept = self.consumer_options.pop(
+            'accept', serialization.setup(config)[1]
+        )
```
mattbennett: We now allow overrides of these "consumer options" by passing them to the constructor of this entrypoint

nameko/rpc.py
```diff
@@ -32,69 +36,82 @@
...
-        self._unregistering_providers.add(provider)
-        remaining_providers = self._providers - self._unregistering_providers
-        if not remaining_providers:
-            _log.debug('unregistering from queueconsumer %s', self)
-            self.queue_consumer.unregister_provider(self)
-            _log.debug('unregistered from queueconsumer %s', self)
-            self._unregistered_from_queue_consumer.send(True)
-
-        _log.debug('waiting for unregister from queue consumer %s', self)
-        self._unregistered_from_queue_consumer.wait()
```
mattbennett: This whole "waiting for providers" dance was required to prevent an RPC entrypoint stopping before the RpcConsumer and QueueConsumer that it depended on; if that happened we'd consume a message and then throw `MethodNotFound` exceptions.

We have a test for this (test_rpc_consumer_sharing) and it still passes with the new implementation.

nameko/rpc.py
```diff
@@ -32,69 +36,82 @@
...
+            'heartbeat', config.get(HEARTBEAT_CONFIG_KEY, DEFAULT_HEARTBEAT)
+        )
+        prefetch_count = self.consumer_options.pop(
+            'prefetch_count', config.get(
+                PREFETCH_COUNT_CONFIG_KEY, DEFAULT_PREFETCH_COUNT
+            )
+        )
+        accept = self.consumer_options.pop(
+            'accept', serialization.setup(config)[1]
+        )
```
mattbennett: Same as https://github.com/nameko/nameko/pull/542/files#r188260399

nameko/rpc.py
```diff
@@ -157,12 +175,26 @@ def handle_message(self, body, message):
...
+
+        service_name = self.container.service_name
+        method_name = self.method_name
+
+        # TODO replace global worker pool limits with per-entrypoint limits,
+        # then remove this waiter thread
+        ident = u"{}.wait_for_worker_pool[{}.{}]".format(
+            type(self).__name__, service_name, method_name
+        )
+        self.container.spawn_managed_thread(spawn_worker, identifier=ident)
```
mattbennett: Same as https://github.com/nameko/nameko/pull/542/files#r188260194

nameko/rpc.py
```diff
@@ -253,149 +318,214 @@ def setup(self):
...
+            'heartbeat', config.get(HEARTBEAT_CONFIG_KEY, DEFAULT_HEARTBEAT)
+        )
+        prefetch_count = self.consumer_options.pop(
+            'prefetch_count', config.get(
+                PREFETCH_COUNT_CONFIG_KEY, DEFAULT_PREFETCH_COUNT
+            )
+        )
+        accept = self.consumer_options.pop(
+            'accept', serialization.setup(config)[1]
+        )
```
mattbennett: Same as https://github.com/nameko/nameko/pull/542/files#r188260399

nameko/rpc.py
```diff
@@ -253,149 +318,214 @@ def setup(self):
...
+        if self.pending:
+            try:
+                with get_connection(self.amqp_uri) as conn:
+                    self.queue.bind(conn).queue_declare(passive=True)
+            except NotFound:
+                raise ReplyQueueExpiredWithPendingReplies(
+                    "Lost replies for correlation ids:\n{}".format(
+                        "\n".join(self.pending.keys())
+                    )
+                )
```
mattbennett: This implements the check for lost replies due to an expired RPC reply queue

nameko/rpc.py
```diff
@@ -253,149 +318,214 @@ def setup(self):
...
+    @property
+    def amqp_uri(self):
+        return self.container.config[AMQP_URI_CONFIG_KEY]
+
+    def setup(self):
 
+        # ReplyListener.setup() runs concurrently in another thread, so
+        # we need to wait here until it's defined its queue
+        while self.reply_listener.queue is None:
+            time.sleep(.1)
```
mattbennett: Workaround until https://github.com/nameko/nameko/wiki/Roadmap#refactor-sub-extension-management is done

nameko/rpc.py
```diff
@@ -253,149 +318,214 @@ def setup(self):
...
+        proxy = Proxy(publish, register_for_reply)
+        return getattr(proxy, self.target_service)
 
-        error = self.resp_body.get('error')
-        if error:
-            raise deserialize(error)
-        return self.resp_body['result']
 
+class Proxy(object):
+    """ Helper object for making RPC calls.
```
mattbennett: This object is the replacement for the `ServiceProxy` and `MethodProxy`

nameko/standalone/rpc.py
```diff
@@ -2,242 +2,154 @@
```
mattbennett: Standalone RPC proxy completely refactored to use the new `nameko.rpc.Proxy` object and a `ReplyListener` that manages its own consumer and looks a lot like the service-based version at `nameko.rpc.ReplyListener`

nameko/testing/pytest.py
```diff
@@ -155,89 +158,6 @@ def amqp_uri(rabbit_config):
     return rabbit_config[AMQP_URI_CONFIG_KEY]
 
 
-@pytest.yield_fixture(autouse=True)
-def fast_teardown(request):
```
mattbennett: Fixture provides no speedup now that we don't wait to cancel consumers on `kill`.

nameko/testing/utils.py
```diff
@@ -2,8 +2,6 @@
 Common testing utilities.
 """
 import socket
-import warnings
```
mattbennett: Removed a couple of unused helpers

test/amqp/test_consume.py
```diff
@@ -0,0 +1,118 @@
+import eventlet
```
mattbennett: Tests for the new `nameko.amqp.consume.Consumer` helper

test/standalone/test_rpc_proxy.py
```diff
@@ -1,31 +1,30 @@
...
 from nameko.testing.waiting import wait_for_call
 
 from test import skip_if_no_toxiproxy
 
 
-# uses autospec on method; needs newer mock for py3
-try:
-    from unittest.mock import patch
-except ImportError:  # pragma: no cover
-    from mock import patch
```
mattbennett: `unittest.mock.patch` and `mock.patch` have different APIs (only one returns the patched object) so have reverted to using just one

test/standalone/test_rpc_proxy.py
```diff
@@ -195,37 +215,120 @@ def list_queues():
...
-        """ Not possible to test this scenario with the current design.
-        We attempt to _setup_consumer immediately on disconnection, without
-        any kind of retry and only for two attempts; the broker will never
-        have expired the queue during that window.
-
-        It will be possible once to write test this scenario once the
-        `rpc-refactor` branch lands. This test will then become very similar
-        to test/test_rpc.py::TestDisconnectedWhileWaitingForReply.
-        """
-        pytest.skip("Not possible to test with current implementation")
```
mattbennett: It is possible now :)

test/standalone/test_rpc_proxy.py
```diff
@@ -300,138 +402,34 @@ def method(self, arg):
...
+            # disable toxiproxy to kill connections
+            toxiproxy.disable()
 
+            # re-enable toxiproxy when the connection error is detected
+            def reconnect(args, kwargs, res, exc_info):
+                block.send(True)
+                toxiproxy.enable()
+                return True
 
-class TestConsumeEvent(object):
```
mattbennett: `ConsumeEvent` doesn't exist anymore

test/standalone/test_rpc_proxy.py
```diff
@@ -726,7 +796,25 @@ def enable_after_retry(args, kwargs, res, exc_info):
 
 
 @skip_if_no_toxiproxy
-class TestStandaloneProxyConsumerDisconnections(object):
+class TestStandaloneProxyReplyListenerDisconnections(object):
```
mattbennett: The diff is messy here. `TestStandaloneProxyReplyListenerDisconnections` is completely new and adds the standard disconnection tests now that the standalone reply listener follows the standard "consumer" pattern.

test/test_amqp.py
```diff
@@ -0,0 +1,263 @@
...
+from nameko.events import EventHandler, event_handler
+from nameko.messaging import Consumer, consume
+from nameko.rpc import ReplyListener, RpcConsumer, RpcProxy, rpc
+from nameko.standalone.events import event_dispatcher
+from nameko.standalone.rpc import ServiceRpcProxy
+from nameko.testing.services import entrypoint_waiter, once
+from nameko.testing.utils import get_extension
+
+
+class TestDeadlockRegression(object):
```
mattbennett: Moved unchanged from test_queue_consumer.py

test/test_amqp.py
```diff
@@ -0,0 +1,263 @@
...
+            if next(counter) == count:
+                return True
+
+        with entrypoint_waiter(
+            container, 'handle_event1', timeout=5, callback=cb
+        ):
+            pass
+
+
+class TestHeartbeats(object):
```
mattbennett: TestHeartbeats and TestHeartbeatFailure are new tests verifying that the heartbeat works correctly on entrypoints/extensions that use the new `nameko.amqp.consume.Consumer` helper

test/test_events.py
```diff
@@ -1,32 +1,27 @@
+from __future__ import absolute_import
```
mattbennett: The changes in this module are either small tweaks to accommodate the new implementation (e.g. expect more connections now) or cleanup where we stop using the `rabbit_manager` to inspect queues (slow and flakey) in favour of the new `queue_info` mechanism that determines queue info using AMQP.

test/test_events.py
```diff
@@ -839,3 +767,76 @@ def test_restricted_parameters(
 
         assert producer.publish.call_args[1]['exchange'] == event_exchange
         assert producer.publish.call_args[1]['routing_key'] == event_type
+
+
+class TestContainerBeingKilled(object):
```
mattbennett: New test for existing functionality

test/test_events.py
```diff
@@ -839,3 +767,76 @@ def test_restricted_parameters(
...
+
+        # check message is requeued if container throws ContainerBeingKilled
+        with patch.object(container, 'spawn_worker') as spawn_worker:
+            spawn_worker.side_effect = ContainerBeingKilled()
+
+            with wait_for_call(Consumer, 'requeue_message'):
+                dispatch("service", "eventtype", "payload")
+
+
+def test_stop_with_active_worker(container_factory, rabbit_config, queue_info):
```
mattbennett: New test for existing functionality, to verify that it hasn't changed

test/test_messaging.py
```diff
@@ -1210,3 +1130,139 @@ def test_use_confirms(self, mock_container, get_producer):
...
+
+        # we don't know which handler picked up the first message,
+        # but all the others should've been handled by Handler2
+        assert messages[-3:] == ['m3', 'm4', 'm5']
+
+        # release the waiting consumer
+        consumer_continue.send(None)
+
+
+class TestContainerBeingKilled(object):
```
mattbennett: New test for unchanged functionality

test/test_messaging.py
```diff
@@ -1210,3 +1130,139 @@ def test_use_confirms(self, mock_container, get_producer):
...
+
+        # check message is requeued if container throws ContainerBeingKilled
+        with patch.object(container, 'spawn_worker') as spawn_worker:
+            spawn_worker.side_effect = ContainerBeingKilled()
+
+            with patch_wait(ConsumerCore, 'requeue_message'):
+                publisher.publish("payload", routing_key=queue.name)
+
+
+def test_stop_with_active_worker(container_factory, rabbit_config, queue_info):
```
mattbennett: New test for unchanged functionality

test/test_queue_consumer.py
```diff
@@ -1,495 +0,0 @@
-import itertools
```
mattbennett: All these tests are now defunct or have been moved elsewhere

test/test_rpc.py
```diff
@@ -499,39 +554,39 @@ def test_rpc_unknown_service_standalone(rabbit_config):
     assert exc_info.value._service_name == 'unknown_service'
 
 
-def test_rpc_container_being_killed_retries(
-        container_factory, rabbit_config):
+class TestContainerBeingKilled(object):
```
mattbennett: Better test for unchanged functionality

test/testing/test_utils.py
```diff
@@ -168,55 +162,6 @@ def test_reset_rabbit_connection_errors():
         reset_rabbit_connections("vhost_name", rabbit_manager)
 
 
-def test_wait_for_worker_idle(container_factory, rabbit_config):
```
mattbennett: Helper no longer exists

test/testing/test_pytest_plugin.py
```diff
@@ -142,147 +142,6 @@ def test_accept(
         assert message.payload == payload
 
 
-class TestFastTeardown(object):
```
mattbennett: Fixture no longer exists

davidszotten:
> As implemented, we end up with a consumer, channel and connection for every queue from which we consume.
any concern about the size of this for service which consumes a lot of different events? does rabbit cope ok?

mattbennett:
The [rabbit docs](https://www.rabbitmq.com/networking.html#tuning-for-large-number-of-connections) talk about optimising for hundreds of thousands of connections, so it's certainly capable. Whether it works with the default settings will depend on the host and the size and number of services connecting to it.

At Student.com we have over 200 event handlers across our service cluster so we'll be able to do some testing with a pre-release build.

davidszotten:
sounds good. pre-release testing will also be good to stress-test the rest of the machinery since this is a pretty big re-write. the old stuff had issues but was at least battle-tested

still reading over the code and comments from that will get collated by github and get posted in a batch

kooba:
Would drop of `QueueConsumer` be considered a breaking change and this land in `v3.x.x`?

nameko/messaging.py
```diff
@@ -88,12 +82,14 @@ def __init__(self, exchange=None, queue=None, declare=None, **options):
...
                 Destination exchange
-            queue : :class:`kombu.Queue`
-                **Deprecated**: Bound queue. The event will be published to
-                this queue's exchange.
             declare : list
                 List of :class:`kombu.Exchange` or :class:`kombu.Queue` objects
                 to declare before publishing.
+            **publisher_options
+                Options to configure the
+                :class:`~nameko.amqqp.publish.Publisher` that sends the
```
kooba: `amqqp/amqp`

nameko/rpc.py
```diff
@@ -253,149 +318,214 @@ def setup(self):
...
 
-class RpcProxy(DependencyProvider):
+class RpcProxy(DependencyProvider, HeaderEncoder):
+    """ DependencyProvider for injecting an RPC proxy into a service.
+
+    :Parameters:
+        target_service : str
+            Target service name
+        **publisher_options
+            Options to configure the :class:`~nameko.amqqp.publish.Publisher`
```
kooba: `amqqp/amqp`

nameko/rpc.py
```diff
@@ -253,149 +318,214 @@ def setup(self):
...
-        if error:
-            raise deserialize(error)
-        return self.resp_body['result']
 
+class Proxy(object):
+    """ Helper object for making RPC calls.
 
-class MethodProxy(HeaderEncoder):
+    The target service and method name may be specified at construction time
+    or by attibute or dict access, for example:
```
kooba: `attibute/attribute`

davidszotten: a bit surprised that the spellchecker (with all its false positives!) isn't catching these...

kooba: Actually not sure if the sentence is correct. Is `dict access` really another way to specify target service and method name?

nameko/standalone/rpc.py
```diff
@@ -2,242 +2,154 @@
...
 
 
-class ServiceRpcProxy(StandaloneProxyBase):
+class RpcProxy(object):
     """
-    A single-threaded RPC proxy to a named service. Method calls on the
-    proxy are converted into RPC calls to the service, with responses
-    returned directly.
+    Single-threaded RPC proxy to a cluster of services. The target service
+    and method are specified with attibutes.
```
kooba: Maybe you taught your spellchecked to accept `attibutes` by now ;)

nameko/standalone/rpc.py
```diff
@@ -2,242 +2,154 @@
...
+
+    Method calls on the locsl object are converted into RPC calls to the
```
kooba: `locsl/local`

test/standalone/test_rpc_proxy.py
```diff
@@ -736,70 +824,170 @@ class Service(object):
...
-                service_rpc.echo(2)
-            assert "ECONNREFUSED" in str(exc_info.value)
+        This failure mode means that the socket between the consumer and the
+        rabbit broker times for out `timeout` milliseconds and then closes.
 
-    def test_reuse_when_recovered(self, service_rpc, toxiproxy):
-        assert service_rpc.echo(1) == 1
+        Attempting to read from the socket after it's closed raises a
+        socket.error and the connection will be re-established. If `timeout`
+        is longer than twice the heartbeat interval, the behaviour is the same
```
kooba: Not a typo but I think Nameko settled on American English spelling? `behaviour/behavior`?

test/standalone/test_rpc_proxy.py
```diff
@@ -736,70 +824,170 @@ class Service(object):
...
 
-        with toxiproxy.disabled():
+    def test_upstream_timeout(self, reply_listener, service_rpc, toxiproxy):
+        """ Verify we detect and recover from sockets timing out.
 
-            with pytest.raises(socket.error) as exc_info:
-                service_rpc.echo(2)
-            assert "ECONNREFUSED" in str(exc_info.value)
+        This failure mode means that the socket between the consumer and the
+        rabbit broker times for out `timeout` milliseconds and then closes.
```
kooba: `times for out / times out for`

nameko/standalone/rpc.py
mattbennett: We use the shell fairly often, but that's all. Ideally we want someone who uses https://github.com/jessepollak/flask-nameko to test the pre-release.

nameko/rpc.py
mattbennett: It is slightly annoying. Open to suggestions!

Makefile
```diff
@@ -3,7 +3,7 @@
 ENABLE_BRANCH_COVERAGE ?= 0
 AUTO_FIX_IMPORTS ?= 0
 
-ifneq ($(AUTO_FIX_IMPORTS), 1)
+ifneq ($(FIX), 1)
```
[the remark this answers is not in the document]
mattbennett: I got tired of typing `AUTO_FIX_IMPORTS` whenever I wanted `make imports` to actually fix the problem rather than just complaining. Didn't intend to commit it, will revert.

nameko/messaging.py
```diff
@@ -46,35 +53,22 @@ def get_message_headers(self, worker_ctx):
...
-                   if value is not None}
-        return headers
+        return encode_to_headers(data, prefix=self.header_prefix)
 
 
 class HeaderDecoder(object):
 
     header_prefix = HEADER_PREFIX
 
-    def _strip_header_name(self, key):
```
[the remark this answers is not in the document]
mattbennett: Good idea. I will drop the class

nameko/messaging.py
```diff
@@ -434,13 +180,46 @@ def handle_message(self, body):
...
+        heartbeat = self.consumer_options.pop(
+            'heartbeat', config.get(HEARTBEAT_CONFIG_KEY, DEFAULT_HEARTBEAT)
+        )
+        prefetch_count = self.consumer_options.pop(
+            'prefetch_count', config.get(
+                PREFETCH_COUNT_CONFIG_KEY, DEFAULT_PREFETCH_COUNT
+            )
+        )
+        accept = self.consumer_options.pop(
+            'accept', serialization.setup(config)[1]
```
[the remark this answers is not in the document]
mattbennett: That would be nice. I tried not to make any changes outside the scope of the AMQP refactor (`serialization.setup` landed in #535) but now we're in principle happy with these changes I'll allow myself some out-of-scope improvements too :)

nameko/messaging.py
```diff
@@ -3,40 +3,47 @@
...
 from kombu.common import maybe_declare
-from kombu.mixins import ConsumerMixin
 
+from nameko import serialization
+from nameko.amqp.consume import Consumer as ConsumerCore
 from nameko.amqp.publish import Publisher as PublisherCore
 from nameko.amqp.publish import get_connection
 from nameko.amqp.utils import verify_amqp_uri
 from nameko.constants import (
-    AMQP_URI_CONFIG_KEY, DEFAULT_HEARTBEAT, HEADER_PREFIX,
```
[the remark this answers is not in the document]
mattbennett: They are cumbersome. What's the alternative?

nameko/rpc.py
```diff
@@ -253,149 +318,214 @@ def setup(self):
...
-            self.amqp_uri, serializer=serializer, **options
+    def __getattr__(self, name):
+        if self.method_name is not None:
+            raise AttributeError(name)
+
+        clone = Proxy(
+            self.publish,
+            self.register_for_reply,
+            self.service_name or name,
+            self.service_name and name
```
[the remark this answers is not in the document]
mattbennett: Fair enough. I'll make it more explicit

nameko/rpc.py
mattbennett: Yep, you can do `n.rpc["service-name"]["method-name"]()`

mattbennett: Spellchecker unfortunately only runs on the `.rst` files, not docstrings. Might be because we don't include the apidoc. I'll see if I can fix it.

test/standalone/test_rpc_proxy.py
mattbennett: My spelling always seems to end up as "transatlantic" 😀 Will fix.

mattbennett:
@kooba Thanks for the review. I will fix all the typos.

Yes, I think this should land in a v3.x.x release. It is compatible in terms of the external API, but I think the internal changes (e.g. many more rabbit connections) warrant a version bump.

I would also like to have upgraded kombu as part of the final 3.x release, but we can possibly make an alpha 3.x release to test beforehand.

test/standalone/test_rpc_proxy.py
mattbennett: Actually `spelling_lang = 'en_GB'` in the sphinx config. Explains why these weren't flagged by the spell checker already.

mattbennett:
@davidszotten @kooba comments addressed.

If you're both happy I'd like to create a 3.x branch that we can do some testing with.

mattbennett:
Merged into v3.0.0-rc branch with #578

PR #547, opened 2018-05-31
Log unhandled worker exceptions at ERROR level

iky:
It's a bit confusing when unhandled worker exceptions are logged at INFO level. Any filtering for more serious log level then sort of hides entrypoint failures. This PR changes the level to ERROR.

### DIFF
--- a/nameko/containers.py
+++ b/nameko/containers.py
@@ -390,7 +390,13 @@ class ServiceContainer(object):
                 with _log_time('ran handler for %s', worker_ctx):
                     result = method(*worker_ctx.args, **worker_ctx.kwargs)
             except Exception as exc:
-                _log.exception('error handling worker %s: %s', worker_ctx, exc)
+                if isinstance(exc, worker_ctx.entrypoint.expected_exceptions):
+                    _log.info(
+                        '(expected) error handling worker %s: %s',
+                        worker_ctx, exc, exc_info=True)
+                else:
+                    _log.exception(
+                        'error handling worker %s: %s', worker_ctx, exc)
                 exc_info = sys.exc_info()
 
             if handle_result is not None:
--- a/nameko/containers.py
+++ b/nameko/containers.py
@@ -390,8 +390,7 @@ class ServiceContainer(object):
                 with _log_time('ran handler for %s', worker_ctx):
                     result = method(*worker_ctx.args, **worker_ctx.kwargs)
             except Exception as exc:
-                _log.info('error handling worker %s: %s', worker_ctx, exc,
-                          exc_info=True)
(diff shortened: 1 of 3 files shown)


iky:
Suggestions are welcome. Maybe entrypoint `expected_exceptions` can be also taken into account here and if the exception raised inside a worker is one of the expected ones, it then can be logged as WARNING or maybe even INFO. But the "unexpected" ones I think should be logged as ERRORs or at least as WARNINGs.

mattbennett:
I think the original rationale for logging these as INFO was that a worker exception (expected or not) is different from an unhandled exception in an extension, and we wanted the latter to be more important.

With the benefit of hindsight though, it would be better to elevate these worker exceptions to ERROR if unexpected and possibly WARNING if expected. We should probably do whatever webservers do when they serve 4xx and 5xx errors.

Additionally, when extensions do throw unexpected exceptions, it kills the service container. We could  elevate the log messages in `ServiceContainer.kill` to FATAL.

davidszotten:
`error` sounds fine for unexpected exceptions, but `warning` might be too high for expected ones, which (at least in my understanding) represent caller errors, eg malformed input.

iky:
Thanks for the review.)

Expected worker exceptions are logged as INFO, unexpected as ERROR. Container exceptions as CRITICAL.

@davidszotten I agree that WARNING for expected exceptions might be quite high. It really depends - bad request kind of exceptions logged as WARNING might suggest to tell clients to fix the way they call the service. On the other hand "not found" sort of exceptions are used regularly by clients to check if an object exists and these are better logged as INFO as there might be many of them.


> We should probably do whatever webservers do when they serve 4xx and 5xx errors.

Tornado, Django and Flask all log 5xx as ERROR and 4xx as WARNING. We're pretty close .)

mattbennett:
The "not found" exception case exists in REST APIs too, and would be logged as a 404 WARNING.

I think we should follow the webserver convention and let users provide a `filter` in their logging configuration if their usecase results in too much verbosity.

davidszotten:
might be easier to just namespace the logger

iky:
I changed the expected exception level from INFO to WARNING to follow the common practice. This results to the following logging levels of `nameko.container` logger:
* `CRITICAL` for container errors
* `ERROR` for unhandled worker errors
* `WARNING` for unhandled but expected worker errors
* `INFO` and `DEBUG` for debugging purposes

The main benefit is to "highlight" entrypoint failures in the logging stream and doing that mainly for Nameko users (meaning people writing and running services) whilst keeping the logger well usable also for Nameko developers (people writing and maintaining extensions).

mattbennett:
I think this is great. @davidszotten?

PR #554, opened 2018-07-12
Adding a new env var constructor for config.yml

santiycr:
## The problem:

If I need to use an env var value like the following:
`142336939733.280047291474`

The current config parser does some parsing and interprets it as a float, which ultimately makes it to my container config like this:
`142336939733.28006`

## The solution

To avoid it, I introduce a new constructor that skips the use of `yaml.safe_load` on the value:
```yaml
    # foobar.yaml
    ENV_THAT_IS_NEEDED_RAW: !raw_env_var "${ENV_THAT_IS_NEEDED_RAW:1234.5660}"
```

This will turn into the string value `1234.5660`, instead of a float number.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -74,6 +74,16 @@ If you need to quote the values in your YAML file, the explicit ``!env_var`` res
     # foobar.yaml
     AMQP_URI: !env_var "pyamqp://${RABBITMQ_USER:guest}:${RABBITMQ_PASSWORD:password}@${RABBITMQ_HOST:localhost}"
 
+If you need to use values as raw strings in your YAML file without them getting converted to native python,
+the explicit ``!raw_env_var`` resolver is required:
+
+.. code-block:: yaml
+
+    # foobar.yaml
+    ENV_THAT_IS_NEEDED_RAW: !raw_env_var "${ENV_THAT_IS_NEEDED_RAW:1234.5660}"
+
+This will turn into the string value ``1234.5660``, instead of a float number.
+
 You can provide many levels of default values
 
 .. code-block:: yaml
--- a/nameko/cli/main.py
+++ b/nameko/cli/main.py
@@ -94,7 +94,8 @@ def env_var_constructor(loader, node, raw=False):
 
 def setup_yaml_parser():
     yaml.add_constructor('!env_var', env_var_constructor)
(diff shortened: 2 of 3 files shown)


davidszotten:
Can you not just quote the value?

santiycr:
`nameko/cli/main.py:97:80: E501 line too long (80 > 79 characters)` 🤦‍♂️

santiycr:
@davidszotten that's what I was expecting, but the way things are implemented, neither single. or double quotes work

mattbennett:
It will work if the value of the environment variable is quoted.

The shell strips quotation marks when declaring env vars (`FOO="foo"` is equivalent to  `FOO=foo`), but if you have control of the variable you can double-quote it to ensure Nameko sees it as a string: `FOO="'foo'"`

This PR may still be useful for cases where you _don't_ have control of the underlying variable though.

mattbennett:
There is perhaps a bug here too. I would expect

```
# FLOAT=0.10
BAR: !env_var "${FLOAT}"  # with surrounding quotes
```
to yield
```
BAR: "0.10"
```
... but you get `0.1` instead

santiycr:
oh, that's a nice hack, to double quote the var at the env export side. I do agree there's a bit of a bug on how the `env_var` parser behaves, but I thought I shouldn't change behavior as heavily used as that as the ramifications for current users could be big. Maybe something to fix for a backwards incompatible release?

santiycr:
@mattbennett anything else needed to approve this? Let me know happy to help out further

mattbennett:
@santiycr This is great as it is, thank you very much.

I considered this possible bug a little further. This works:

```
# FLOAT=0.10
BAR: !env_var "'${FLOAT}'"  # with double surrounding quotes
```
Yields:
```
BAR: "0.10"
```

It works for the same reason as the double-quote env var hack: the outer quotes perform a grouping operation so that you can pass values with spaces in, hence are stripped. I don't this constitutes a bug. It has been an interesting detour but I much prefer the explicit approach you've added here.

While experimenting I added some tests for double-quoted values which I have pushed to your branch. 

This is good to go, thank you very much!

santiycr:
Oh, I like your temp solution. My current one while I waited for a release was a lot uglier:

`BAR: !env_var "IGNORE_${FLOAT}"`

Thanks for sharing a better approach in the meantime!

PR #557, opened 2018-07-14
Bump min eventlet

davidszotten:
0.20.1 was released in January 2017, similar age to previous version
bumps, and this way we can e.g. start using a more modern pytest

also bump our "pinned" set of some dependencies for testing to the
latest versions as of today

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,13 @@ Here you can see the full list of changes between nameko versions. Versions
 are in form of *headline.major.minor* numbers. Backwards-compatible changes
 increment the minor version number only.
 
+Version 2.10.0
+-------------
+
+Released: Pending
+
+* Bump the minimum supported eventlet version to 0.20.1 (#557)
+
 
 Version 2.9.1
 -------------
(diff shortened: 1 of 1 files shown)


mattbennett:
Seems sensible to me.

We're failing here cause this version of pytest doesn't pick up the coverage generated by `testdir.runpytest`. I can't remember why I avoided using `coverage run` but perhaps that's an easy fix.

davidszotten:
so don't know for sure but i wonder if we needed `pytest-cov` because `testdir.runpytest` used to launch a subprocess and wouldn't otherwise be seen by coverage. now it seems to happen in-process so `coverage run` works (and `pytest-cov gets confused because the main test suite already imports the plugin before coverage can start, c.f. http://pytest-cov.readthedocs.io/en/latest/plugins.html)

also for some reason `py36` doesn't seem to like eventlet 0.20 so bumped oldest for that one to 0.21

davidszotten:
guess we might cut a point release wit the stuff currently pending on master and then at least bump the minor version for this change

PR #564, opened 2018-08-06
[WIP] Bump kombu to version 4.2

s-maj:
Work in progress.

### DIFF
--- a/nameko/amqp/utils.py
+++ b/nameko/amqp/utils.py
@@ -7,9 +7,6 @@ from kombu import Connection
 from kombu.transport.pyamqp import Transport
 
 
-BAD_CREDENTIALS = (
-    'Error connecting to broker, probably caused by invalid credentials'
-)
 BAD_VHOST = (
     'Error connecting to broker, probably caused by using an invalid '
     'or unauthorized vhost'
@@ -25,10 +22,8 @@ class ConnectionTester(amqp.Connection):
         try:
             super(ConnectionTester, self).__init__(*args, **kwargs)
         except IOError as exc:
-            if not hasattr(self, '_wait_tune_ok'):
+            if not hasattr(self, 'on_tune_ok'):
                 raise
-            elif self._wait_tune_ok:
-                six.raise_from(IOError(BAD_CREDENTIALS), exc)
             else:  # pragma: no cover (rabbitmq >= 3.6.0)
                 six.raise_from(IOError(BAD_VHOST), exc)
         except NotAllowed as exc:  # pragma: no cover (rabbitmq < 3.6.0)
--- a/nameko/messaging.py
+++ b/nameko/messaging.py
@@ -162,7 +162,7 @@ class Publisher(DependencyProvider, HeaderEncoder):
 
         with get_connection(self.amqp_uri, ssl) as conn:
             for entity in self.declare:
-                maybe_declare(entity, conn)
+                maybe_declare(entity, conn.channel())
 
         serializer = self.options.pop('serializer', self.serializer)
 
--- a/nameko/amqp/__init__.py
+++ b/nameko/amqp/__init__.py
@@ -2,8 +2,3 @@
 from .utils import verify_amqp_uri  # noqa: F401
 from .publish import (  # noqa: F401
     get_connection, get_producer, UndeliverableMessage)
-
-from amqp.transport import _AbstractTransport
-
-# backport http://bit.do/do-not-implement-del
-del _AbstractTransport.__del__
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ setup(
     packages=find_packages(exclude=['test', 'test.*']),
     install_requires=[
         "eventlet>=0.20.1",
-        "kombu>=3.0.1,<4",
+        "kombu>=4,<5",
         "mock>=1.2",
         "path.py>=6.2",
         "pyyaml>=3.10",
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ setup(
     packages=find_packages(exclude=['test', 'test.*']),
     install_requires=[
         "eventlet>=0.20.1",
-        "kombu>=4,<5",
+        "kombu>=4.2.0,<5",
         "mock>=1.2",
         "path.py>=6.2",
         "pyyaml>=3.10",
--- a/tox.ini
+++ b/tox.ini
@@ -7,7 +7,7 @@ deps =
     # oldest supported libraries for each python
     {py27,py34,py35}-oldest: eventlet==0.20.1
     py36-oldest: eventlet==0.21.0
-    oldest: kombu==3.0.30
+    oldest: kombu==4.2.0
     oldest: mock==1.2.0
     oldest: path.py==6.2
     oldest: requests==2.4.3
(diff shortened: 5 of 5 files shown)


s-maj:
I've fixed basic stuff but it looks like writing to the the closed connections is somewhat broken. See tests:
* test/test_messaging.py:852 TestPublisherDisconnections.test_down[True]
* test/test_messaging.py:852 TestPublisherDisconnections.test_down[False]
* test/test_messaging.py:870 TestPublisherDisconnections.test_timeout[True]
* test/test_messaging.py:870 TestPublisherDisconnections.test_timeout[False]

 It seems that exceptions is never raised and hanging thread is just killed after timeout.

nameko/amqp/utils.py
```diff
@@ -7,9 +7,6 @@
 from kombu.transport.pyamqp import Transport
 
 
-BAD_CREDENTIALS = (
```
davidszotten: if i recall correctly, newer kombu has much better error messages. if so we can probably drop this whole module

test/amqp/test_publish.py
```diff
@@ -150,7 +149,7 @@ def test_confirms(self, amqp_uri, publisher):
 
         # confirms are enabled by default;
         # if a mandatory message cannot be delivered, expect an exception
-        with pytest.raises(UndeliverableMessage):
+        with pytest.raises(ChannelError):
```
mattbennett: Review comment got lost somehow:

I think we need to keep `UndeliverableMessage` when `publisher.publish` can't deliver a mandatory message.

s-maj:
I've added `transport.options` since some tests were timeouting without specic error (similar behaviour to https://github.com/celery/celery/issues/4296 and other related tickets), where the initial problem with tests was related to dropping  (by accident) `confirm_publish` from `transport.options` in `get_producer`. 
If I understand this correctly, connections with kombu later then 4.0.2 will hang indefinitely without transport options.

mattbennett:
That's interesting. Given all those reported problems, the safest course of action is probably to set the `transport_options` as you have. Then at least we won't be silencing exceptions that people may be relying on.

The best solution would be to make `transport_options` configurable somehow, in the same way that `publish_confirm` are. This is best saved for a later PR though.

test/test_rpc.py
```diff
@@ -487,16 +488,16 @@ def test_rpc_unknown_service(container_factory, rabbit_config):
         with pytest.raises(RemoteError) as exc_info:
             proxy.call_unknown()
 
-    assert exc_info.value.exc_type == "UnknownService"
+    assert "NO_ROUTE" in str(exc_info.value)
```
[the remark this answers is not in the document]
s-maj: Ok, with newer Kombu this exception looks like this: https://gist.github.com/s-maj/3cf85e49ae1aae69f1054243aa6a2c67. I'm not sure how I can preserve UnknownService here.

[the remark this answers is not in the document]
mattbennett: Can you not catch and inspect the `ChannelError` from https://github.com/nameko/nameko/blob/master/nameko/amqp/publish.py#L179?

Something like

```python
try:
    publisher.publish(
       ...
    )
except ChannelError as exc:
    if str(exc) == "NO_ROUTE":
        raise UndeliverableMessage()
    raise
```

mattbennett:
These lines can be removed now, never reach the original place where `UndeliverableMessage` was raised:

https://github.com/nameko/nameko/blob/c7e8f43a9f2c31d1d16dc76efd464f6e47aa9ee3/nameko/amqp/publish.py#L217-L223

Also it looks like the `mock_connection` fixture no longer has any uses. You can either remove it or `#pragma: no cover` it.

s-maj:
I've changed this slightly to keep constants immutable and provide reasonable defaults for low level methods. Let me if this is ok (tests yet to be fixed).

nameko/amqp/publish.py
```diff
@@ -15,17 +18,22 @@ class UndeliverableMessage(Exception):
...
-def get_producer(amqp_uri, confirms=True, ssl=None):
-    transport_options = {
-        'confirm_publish': confirms
-    }
+def get_producer(amqp_uri, confirms=True, ssl=None, transport_options=None):
+    if transport_options:
+        transport_options['confirm_publish'] = confirms
+    else:
+        transport_options = deepcopy(DEFAULT_TRANSPORT_OPTIONS)
+        transport_options['confirm_publish'] = confirms
```
mattbennett: This would be slightly neater here:

``` python
if transport_options is None:
    transport_options = DEFAULT_TRANSPORT_OPTIONS.copy()
transport_options['confirm_publish'] = confirms
```

No need for deepcopy if `transport_options` doesn't contain any nested mutable objects.

I wonder if we should deprecate passing `confirms` as its own keyword argument and expect it to be passed as part of `transport_options`.

nameko/messaging.py
```diff
@@ -157,12 +158,15 @@ def serializer(self):
     def setup(self):
 
         ssl = self.container.config.get(AMQP_SSL_CONFIG_KEY)
+        transport_options = self.container.config.get(
+            TRANSPORT_OPTIONS_CONFIG_KEY, DEFAULT_TRANSPORT_OPTIONS
+        )
```
mattbennett: This makes `transport_options` configurable at the service level, rather than the extension level. That might be what we want, but some transport options (e.g. confirms) may want to be different across extensions in the same service.

For the DependencyProviders (`Publisher`, `EventDispatcher` and `RpcProxy`) you could allow `transport_options` to be passed to the constructor (in the same way that `use_confirms` can be) so you could override the default for that particular extension. The corresponding entrypoints don't have the same mechanism for accepting parameters in their constructors, so we'll have to use the service-level options there.

s-maj: Hmm, I can't see it any references to `use_confirms ` in https://github.com/nameko/nameko/blob/master/nameko/messaging.py#L73 except https://github.com/nameko/nameko/blob/master/nameko/messaging.py#L132. 

All `use_confirms` usage seems to be in https://github.com/nameko/nameko/blob/master/nameko/amqp/publish.py#L35 and  I already added transport options https://github.com/nameko/nameko/pull/564/commits/413006b183fdef0b0098771ee5a68428ee2285a6#diff-da23ae33a4d5c536cd0bfb9ca895717aR120.

mattbennett: `use_confirms` can be passed into https://github.com/nameko/nameko/blob/master/nameko/messaging.py#L73  as part of ``**options``, and then extracted out of them when that dict is passed to the constructor of https://github.com/nameko/nameko/blob/master/nameko/amqp/publish.py#L35. Actually `transport_options` could already be provided to the DependencyProvider constructor and would also end up in `**options`.

The test at https://github.com/nameko/nameko/blob/0bffbf7b3c2a26ab664075a48d5fc3e136205413/test/test_messaging.py#L1189-L1213 demonstrates this for `use_confirms`. It'd be good to have an equivalent for `transport_options`.

At the moment, the line https://github.com/nameko/nameko/pull/564/files#diff-309aaafd3b61875d021b62c5b3926f91R178 will cause a duplicate keyword argument exception if you did pass `transport_options` to the `Publisher` DependencyProvider constructor. I think if you just removed that line we'd get the desired behaviour.

Does that make sense?

test/test_messaging.py
```diff
@@ -1205,11 +1205,12 @@ def test_use_confirms(self, mock_container, get_producer):
         publish = publisher.get_dependency(worker_ctx)
 
         publish("payload")
-        (_, use_confirms, _), _ = get_producer.call_args
+        print(get_producer.call_args)
```
mattbennett: Rogue print statement here

davidszotten:
sounds reasonable. i haven't done a thorough review of the pr (but happy if you have). we need a changelog entry and if you want you can add yourself  to `contributors.txt`

PR #566, opened 2018-08-09
no need for verify_amqp_url; amqp has nice error messages now

davidszotten:


davidszotten:
i did add a comment to this effect on the kombu upgrade pr but it was ignored

mattbennett:
So you did, sorry. I completely missed that.

davidszotten:
i figured i could just do it afterwards

PR #573, opened 2018-09-03
Include new custom headers in messaging.publish

timbu:
Fixes a problem where additional custom headers are not propagated via a messaging Publish call.

Only headers present at the start of the worker-request were previously propagated.

### DIFF
--- a/test/test_messaging.py
+++ b/test/test_messaging.py
@@ -247,6 +247,41 @@ def test_publish_custom_headers(
     ]
 
 
+@pytest.mark.usefixtures("predictable_call_ids")
+def test_publish_custom_headers_added_in_current_request(
+    mock_container, mock_producer, rabbit_config
+):
+    container = mock_container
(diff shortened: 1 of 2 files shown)


test/test_messaging.py
```diff
@@ -247,6 +247,41 @@ def test_publish_custom_headers(
...
+    worker_ctx = WorkerContext(
+        container, service, DummyProvider('method'), data=ctx_data
+    )
+
+    publisher = Publisher(queue=foobar_queue).bind(container, "publish")
+    publisher.setup()
+
+    # test publish
+    msg = "msg"
+    headers = {'nameko.language': 'en',
```
mattbennett: Can we call this `expected_headers` and define it _after_ the call to `publish`?

mattbennett:
Something has broken here with the examples tests. I don't think it's related to this PR but I will leave this open until the issue is fixed.

PR #576, opened 2018-09-24
get result's exchange from incoming message (or eventualy from config file).

rejoc:
This will allow to respond to messages coming from "parent" exchange of an alternate exchange processing the message.

For example, if you have multiple people working on a project, each one can have it's own exchange where they run the services they are working on, and all these exchanges can have the same alternate exchange where you can run the "common" services. So each programmer does not need to run all the services of a project.

### DIFF
--- a/nameko/rpc.py
+++ b/nameko/rpc.py
@@ -123,7 +123,7 @@ class RpcConsumer(SharedExtension, ProviderCollector):
         serializer = self.container.config.get(
             SERIALIZER_CONFIG_KEY, DEFAULT_SERIALIZER
         )
-        exchange = get_rpc_exchange(self.container.config)
+        exchange = message.delivery_info.get('exchange', get_rpc_exchange(self.container.config))
         ssl = self.container.config.get(AMQP_SSL_CONFIG_KEY)
 
         responder = Responder(amqp_uri, exchange, serializer, message, ssl=ssl)


mattbennett:
This seems quite a niche use-case to me. It's not outrageous to send the RPC reply to the exchange from which the request came, but I'm sure there will be implications of making this change. There may already be some failing tests, and if not perhaps we're missing some coverage.

Is the situation you're describing in the summary a hypothesis, or reality after this change? If it's reality then some tests demonstrating as much would be great ;)

rejoc:
Yes it's a reality ;-)

It's not very easy to buid automatic tests as you need to setup a policy to describe the alternate exchange of one exchange.

Then, in a pure amqp philosophy, I think it's better to answer a RPC request to the connection_id AND the exchange given by the request (or to respond directly to the queue that sent the request without using an exchange). 
The exchange defined in the configuration file is to be used to send requests..

mattbennett:
Alternate exchanges are a nice feature, and this use-case of a shared set of RPC services is pretty neat! 

I also agree with you that replying to the exchange specified in the message is nicer than to some arbitrary application-configured one. Since RPC requests and replies all currently flow through the same exchange, making this change should be backwards compatible. Is the fallback to the configuration-specified exchange actually necessary?

I would love to see an end-to-end test for this use-case still, if nothing else than because tests are (currently) the best documentation we have. [It is possible](https://www.rabbitmq.com/ae.html#define-using-arguments) to declare an AE with the client, which should make doable during a test-run.

rejoc:
I also think that defaulting the exchange to the config definition is not necessary. But I was not sure 100%. I'll take it off.

I have to check the definition of the alternate exchange within the exchange definition. This is not the solution I choosed for my configuration because it is not the "prefered" method (I did the AE definition policy from the RMQ web interface..). 
My other concern about this method is about : what happens if the exchange already exists and you add or remove an alternate exchange. Does it accept the change or return an error ?

To implement the alternate exchange through policies (the preferred solution for RMQ) , I think I need to use the python rabbitmq_admin library [https://pypi.org/project/rabbitmq-admin/](url). What do you think about that ?

Whatever the method to create/bind this AE, I could add a "rpc_alternate_exchange" parameter in the config file and use it to automaticaly create the AE.

mattbennett:
> To implement the alternate exchange through policies (the preferred solution for RMQ) , I think I need to use the python rabbitmq_admin library https://pypi.org/project/rabbitmq-admin/. What do you think about that?

This library wasn't around when Nameko first needed to manipulate the RabbitMQ REST API, so we implemented our own wrapper in `nameko.testing.rabbit`. We could replace this with a dependency on `rabbitmq_admin`. But since (I think) RabbitMQ 3.6.7 the management API is harder to use -- the HTTP request returns before the changes are applied, which makes for flakey tests.

Even though AMQP declaration of AEs is "discouraged", if it works then I think it's a better approach for the specific use-case of Nameko's test suite. And for this specific use-case, the exchange will never previously exist, because each test is run in its own vhost.

PR #582, opened 2018-10-11
Hide password in logged amqp uri

daviskirk:
On connection errors, the amqp uri is logged unfiltered, leaking the users (possibly) sensitive password.


test/utils/test_utils.py
```diff
@@ -240,3 +242,26 @@ def test_import_module(self):
...
+    (
+        'amqp://user:supersecret@127.0.0.1:5672//',
+        'amqp://user:{}@127.0.0.1:5672//'.format(REDACTED)
+    ),
+    (
+        'amqp://user@127.0.0.1:5672//',
+        'amqp://user@127.0.0.1:5672//'
+    ),
+    (
+        'amqp://user@127.0.0.1:5672//',
```
mattbennett: This looks like a duplicate test case

daviskirk: Sure does... sorry about that

davidszotten:
nice catch! +1

PR #585, opened 2018-10-29
Add ClusterRpc DependencyProvider

mattbennett:
As discussed [on the forum](https://discourse.nameko.io/t/non-persistent-clusterrpcproxy/350/7).

A dynamic RPC proxy DependencyProvider has been requested several times before, and the refactor landing in the v3.x release makes it much easier to provide.

This change simply makes the `target_service` optional when instantiating the `RpcProxy`.

### DIFF
--- a/docs/built_in_extensions.rst
+++ b/docs/built_in_extensions.rst
@@ -8,7 +8,7 @@ Nameko includes a number of built-in :ref:`extensions <extensions>`. This sectio
 RPC
 ---
 
-Nameko includes an implementation of RPC over AMQP. It comprises the ``@rpc`` entrypoint, a proxy for services to talk to other services, and a standalone proxy that non-Nameko clients can use to make RPC calls to a cluster:
+Nameko includes an implementation of RPC over AMQP. It comprises the ``@rpc`` entrypoint, a client for services to talk to other services, and a standalone client that non-Nameko clients can use to make RPC calls to a cluster:
 
 .. literalinclude:: examples/rpc.py
 
--- a/docs/community_extensions.rst
+++ b/docs/community_extensions.rst
@@ -53,7 +53,7 @@ Extensions
     * `nameko-sendgrid <https://github.com/invictuscapital/nameko-sendgrid>`_
 
         SendGrid dependency for nameko, for sending transactional and marketing emails.
-        
+
     * `nameko-cachetools <https://github.com/santiycr/nameko-cachetools>`_
 
         Tools to cache RPC interactions between your nameko services.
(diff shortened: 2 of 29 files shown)


mattbennett:
That's a good point. Introducing this as a new feature would make it easier to communicate too.

We use the thin wrapper pattern in the standalone package. In the v3 branch I renamed the standalone `ClusterRpcProxy` to simply `RpcProxy`, which is unfortunate if we want to alias the `ServiceRpcProxy` DP to `RpcProxy` for backwards compatibility here.

 I think the best thing to do is roll back that change so we end up with a `ServiceRpcProxy` and `ClusterRpcProxy` in both the standalone and normal packages. We should add an `RpcProxy` alias to `ServiceRpcProxy` in the normal package to maintain backwards compatibility, and perhaps switch the new standalone `RpcProxy` alias to point to the `ServiceRpcProxy` instead, so it matches.

davidszotten:
sounds reasonable.

though this made me think of something else: if we are considering changing the names, maybe we should also put the word `Proxy` on the table. not sure how we landed on that name in the first place, but maybe e.g `Client` or something else would make a better name

mattbennett:
Ha! Yes, `Proxy` has always been awkward. What do you think about the following?

* `RpcProxy` / `ServiceRpcProxy` => `ServiceRpc` or ``ServiceRpcClient`` or `ServiceClient` 
* `ClusterRpcProxy` => `ClusterRpc` or `ClusterRpcClient` or `ClusterClient`

Or possibly one object called ``RpcClient`` with the API as currently implemented, service name optional. Maybe less confusing if we're replacing the existing API with a brand new one?

I think my preference is `ServiceRpc` and `ClusterRpc`.

davidszotten:
I don't have any strong views on actual names, though i think i prefer two separate ones. `ServiceRpc` and `ClusterRpc` seem fine

mattbennett:
I have ended up using `ServiceRpc` and `ClusterRpc` for the DependencyProviders, and `ServiceRpcClient` and `ClusterRpcClient` for the standalone clients. Mostly because we have several test files that import from both modules.

mattbennett:
@davidszotten are you happy with the changes here now?

davidszotten:
this seems fine to me

not sure i'd say a `ServiceRpc` _is a_ `ClusterRpc` etc (this inherits atm)

mattbennett:
> not sure i'd say a ServiceRpc _is a_ ClusterRpc etc

Have I said that somewhere?

davidszotten:
It’s defined as a subclass

davidszotten:
sorry for being unclear. it's the fact that it's a subclass ([="is a"](https://en.wikipedia.org/wiki/Is-a)) that i'm (mildly) poking at

mattbennett:
Oh I'm with you. I guess it would be nicer to use composition rather than inheritance, but I don't think this is the first use of inheritance in Nameko that isn't strictly a "is-a" type relationship.

I will merge this and cut another pre-release for the 3x branch.

davidszotten:
Indeed. Hence only mildly prodding at this

PR #586, opened 2018-10-31
Fix license

mattbennett:
GitHub could not previously recognise this as the Apache License


NOTICE.txt
```diff
@@ -1,10 +1,10 @@
-Copyright 2015 Lifealike, Ltd. DBA onefinestay
+Copyright 2015 Nameko Project Contributors
```
davidszotten: was the copyright reassigned?

mattbennett: As I understand it, the copyright is shared by all contributors. It's also implied and does not need to be stated anywhere in order to be in effect, but including a statement somewhere is recommended for the avoidance of doubt. 

So this doesn't change anything material but is a better reflection of reality.

See https://softwarefreedom.org/resources/2012/ManagingCopyrightInformation.html

mattbennett:
@davidszotten are you good with this?

PR #587, opened 2018-11-05
Updated Support to go to https://discourse.nameko.io/

shirishc:
Pointing to the new discussion forum rather than google groups. Keeping the google group line still there in case the old Q&A knowledge base hasn't been migrated to https://discourse.nameko.io/. Cheers

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -62,7 +62,7 @@ Getting Started
 Support
 -------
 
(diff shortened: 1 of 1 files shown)


mattbennett:
Thanks @shirishc! The old list has actually been completely imported into Discourse, so I think you can drop the old link.

Issue #276, opened 2015-08-10
Make it more obvious the example needs rabbitmq


mattbennett:
The example service on the index page encourages you to just try to run it, which fails if you haven't installed RabbitMQ. Make it more obvious you need to follow the installation process first.

This should perhaps get lumped into a larger piece of work to separate out the project page from the docs page.

jiamo:
Just about the doc. I only find code in test uinsg py.test framework to test service. Can we have some examples don't using py.test. In real project we may don't use py.test helper as client code to call service.

mattbennett:
Nameko uses py.test for its own tests, and we encourage others to do so as well. It's also basically compatible with the other testing frameworks, so I don't see much value in adding non-pytest examples. You should be able to port the same concepts to another testing framework if you really need to.

omidraha:
It's obviously, but please add something like this to the example of index page:

> To run above code, insure that you have RabbitMQ running, for example:
> `$ docker run  -p5672:5672 rabbitmq:latest`

tyler46:
@mattbennett I can handle this addition on the docs if it's ok.

mattbennett:
@tyler46 yes please!

tyler46:
@mattbennett great, I will prepare a PR for review.

iky:
Fixed by https://github.com/nameko/nameko/pull/591

PR #591, opened 2018-11-21
Fix for issue #276

tyler46:
I think now it explicit that rabbitmq is needed in order to run nameko example service.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -24,6 +24,18 @@ A nameko service is just a class:
         def hello(self, name):
             return "Hello, {}!".format(name)
 
+Specific example requires `RabbitMQ <https://www.rabbitmq.com>`_, because
+it's using the built-in AMQP RPC features. `RabbitMQ installation guidelines <https://www.rabbitmq.com/download.html>`_
+offer several installation options, but you can quickly install and run it
+using `Docker <https://docs.docker.com/install/>`_.
+
+To install and run RabbitMQ using docker:
+
+.. code-block:: shell
+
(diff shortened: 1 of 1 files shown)


davidszotten:
Hi,

Thanks for the contribution!

A small note: You write "to run nameko you need [rabbitmq]" which isn't quite right. You need rabbitmq to use the built-in rpc functionality that's used in the example, but it's not required by the core framework. I'm not sure what the bast way to phrase things but i think it's worth making this distinction clear

(@mattbennett This is making me wonder if we should look for a different hello-world example. http has fewer dependencies but i think may be less good at making it clear how we're different from web frameworks like flask)

tyler46:
Hi @davidszotten,

I thought by writing "nameko has built its RPC and other features over AMQP", I was explaining why we  need [rabbimtq]. Of course I can rephrase it to be more accurate. Do you think something like the following would be better:

> Before running nameko service, please be sure that you have a [rabbitmq] instance running, since nameko built-in rpc functionality relies on AQMP that is being implemented by [rabbitmq].

I agree with you that using a hello-world example using http entrypoint would be less good that current one.

mattbennett:
@tyler46 thanks for the PR.

The difficulty here is that _this specific_ example requires RabbitMQ, because it's using the built-in AMQP RPC features. I think it'd be sufficient to just say that.

On the project page we feature an RPC example and an HTTP example, but it may be distracting to have both here. The RPC one has the advantage of also being able to show the shell usage underneath.

tyler46:
Hi @mattbennett,

OK I will update PR accordingly.

Thank you

PR #592, opened 2018-11-22
Compat fixes

mattbennett:
While testing [nameko-amqp-retry](https://github.com/nameko/nameko-amqp-retry) against the v3 pre-release, I discovered two compatibility issues:

1. A typo in the backwards compat shims in `nameko.standalone.rpc` (45034c1)
2. `expected_exceptions` and `sensitive_arguments` were swallowed by the `Consumer` (720cdc3)

c7a7ac2 adds tests to _all_ built-in entrypoints verifying that `expected_exceptions` and `sensitive_arguments` can be passed.  6768d1a fixes the `Timer` implementation, which did not accept them.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,5 +3,5 @@ universal = 1
 
 [flake8]
 ignore =
-max-line-length = 79
+max-line-length = 88
 exclude = test/nameko_doc/example/*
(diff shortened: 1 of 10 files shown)


setup.cfg
```diff
@@ -3,5 +3,5 @@ universal = 1
 
 [flake8]
 ignore =
-max-line-length = 79
+max-line-length = 88
```
iky: nice .)

mattbennett: This is a pre-cursor to just turning on black ;)

PR #596, opened 2018-11-25
wrap the note about rabbit in a `note`

davidszotten:
helps make it clear where the sidenote ends


davidszotten:
not sure why this is still marked as failing in gh, looks fine to me in travis

PR #597, opened 2018-11-29
Exceptions raised in shell will not be caught when not in TTY

vmikki:
This PR changes the behaivour of Nameko shell so it doesn't catch exceptions when used in non-TTY mode. A use case when this would be useful is when nameko shell used for cron jobs and you need to know wether the executed command was successful or not. For example after executing the following command:
```sh
$ echo "n.rpc.other_service.foo()" | nameko shell
```
the exit code will be set to 1 if the execution of the remote method failed.

### DIFF
--- a/nameko/cli/code.py
+++ b/nameko/cli/code.py
@@ -1,3 +1,5 @@
+from __future__ import absolute_import
+
 from code import InteractiveConsole
 
 
--- a/nameko/cli/code.py
+++ b/nameko/cli/code.py
@@ -1,0 +1,36 @@
+from code import InteractiveConsole
+
+
+class RaisingInteractiveConsole(InteractiveConsole):
+    """ Custom InterativeConsole class that allows raising exception if needed.
+    """
+
+    def __init__(
+        self, locals=None, filename="<console>", raise_expections=False
+    ):
+        super(RaisingInteractiveConsole, self).__init__(
+            locals=locals, filename=filename
+        )
+        self.raise_expections = raise_expections
+
+    def runcode(self, code):
+        try:
+            exec(code, self.locals)
+        except SystemExit:
+            raise
+        except:
+            self.showtraceback()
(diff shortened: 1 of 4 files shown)


mattbennett:
We pipe input to the shell quite frequently, as a way of essentially writing small scripts that use the `n` helper. It is helpful to bubble exceptions up to the process level in this case.

The alternative is to write a standalone script, but you have to reimplement all the config parsing and helper setup.

nameko/cli/code.py
```diff
@@ -0,0 +1,36 @@
+from code import InteractiveConsole
```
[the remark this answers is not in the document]
vmikki: I have added `from __future__ import absolute_import` and now it works in Python 2.7.

vmikki:
I have fixed the broken builds and now my PR is ready to be reviewed.
I had to use subprocesses in order to reliably test the new feature and the code covered by those tests is marked as not covered in coverage. The possible solution for that would have been a bit hacky and unreliable so I just added `no cover` instead, I hope it's fine.

nameko/cli/code.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import absolute_import
+
+from code import InteractiveConsole
+
+
+class RaisingInteractiveConsole(InteractiveConsole):  # pragma: no cover
```
mattbennett: You might be able to `no cover` the whole module in one go.

I think it'd also be worth putting a little note explaining why it's excluded from coverage.

vmikki: Is there a way to add `no cover` to the whole module? I think excluding the whole file can only be done by specifying the `omit` parameter for coverage but in that case it will not be visible in the module that it is omitted from coverage. I agree that it should be explained why `no cover` is there but perhaps it would be better to be explicit and keep the current `no cover`s with some comment next to them.

PR #600, opened 2019-01-11
Set extra headers for outbound messaging as late as possible

iky:
Allows for in-entrypoint context data updates. This is useful when setting a language for example:
```python

class LanguageContext(DependencyProvider):

    def get_dependency(self, worker_ctx):
        @contextmanager
        def set_language(language):
            original_language = worker_ctx.data.get('language')
            worker_ctx.data['language'] = language
            yield
            worker_ctx.data['language'] = original_language
        return set_language

class ExampleService(object):

    name = 'exampleservice'

    language_context = LanguageContext()

    @rpc
    def greet_in_french(self):
        with self.language_context('fr'):
            return self.example_rpc.say_hello()
```
Another use we have is services signing up outbound calls with its own authentication.

I removed the lazy load and caching of `call_id` and `call_id_stack` of worker context. These properties are super-cheap and most of the time loaded anyway (if entrypoint contains an outbound call, or in logging debug call_id_stack is logged by the container, if tracer is on and so on...)

### DIFF
--- a/nameko/rpc.py
+++ b/nameko/rpc.py
@@ -478,12 +478,13 @@ class Client(object):
 
         # target at construction time
         client = Client(
-            publish, register_for_reply, "target_service", "target_method"
+            publish, register_for_reply, context_data,
+            "target_service", "target_method"
         )
         client(*args, **kwargs)
 
         # equivalent with attribute access
-        client = Client(publish, register_for_reply)
+        client = Client(publish, register_for_reply, context_data)
         client = client.target_service.target_method  # now fully-specified
         client(*args, **kwargs)
 
@@ -497,6 +498,8 @@ class Client(object):
         register_for_reply : callable
             Function to register a new call with a reply listener. Returns
             another function that should be used to retrieve the response.
+        context_data: dict
+            Worker context data to be sent as extra headers
         service_name : str
             Optional target service name, if known
         method_name : str
--- a/nameko/rpc.py
+++ b/nameko/rpc.py
@@ -443,13 +443,10 @@ class ClusterRpc(DependencyProvider):
 
     def get_dependency(self, worker_ctx):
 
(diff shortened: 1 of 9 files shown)


timbu:
@mattbennett Is this the same problem from here? https://github.com/nameko/nameko/pull/573/files

mattbennett:
Gah, yes it is! Sorry @timbu, I completely forgot about that PR.

Interesting that this was introduced for the non-RPC extensions before the big refactor that landed in 3.x.

mattbennett:
I guess we should land #573 and then bring master into the v3 branch.

iky:
Ah, did not see that. @timbu @mattbennett It's not just Publisher of messaging module, EventDispatcher of events module needs to be changed too, see https://github.com/nameko/nameko/pull/600/commits/5c416cee1dcfd791839e2a7bedd2b8ee557a1aca

iky:
I've updated the remaining places. The fixes are tiny, maybe we can merge this to 3.0.0 before #573 lands in master and then dealing with merging the two together?

mattbennett:
> @timbu @mattbennett It's not just Publisher of messaging module, EventDispatcher of events module needs to be changed too, see 5c416ce

Ah, quite right. And that means #573 is missing the equivalent fix for the event dispatcher.

I agree, let's make the fixes here, then #573 can land, then we'll merge master into the v3rc branch.

nameko/rpc.py
```diff
@@ -507,10 +504,12 @@ class Client(object):
     """
 
     def __init__(
-        self, publish, register_for_reply, service_name=None, method_name=None
+        self, publish, register_for_reply, context_data,
```
kooba: Might want to add `context_data` to `:Parameters:` doc above.

iky: good catch, thanks! .)

kooba:
👍

PR #602, opened 2019-01-12
[WIP] Better config api

iky:


nameko/__init__.py
```diff
@@ -1,8 +1,71 @@
...
+        """
+        A context manager and decorator for updating config just for the given context
+        or decorated function execution.
+
+        Intended to be used mainly in tests:
+
+            @pytest.yield_fixture
+            def memory_rabbit_config():
+                with nameko.config.patch({"AMQP_URI": "memory://"}):
+                    yield
```
mattbennett: It's a little confusing that this fixture is defined but not used. I think you can drop this and just demonstrate the decorator usage.

nameko/__init__.py
```diff
@@ -1,8 +1,71 @@
...
+            @nameko.config.patch({"AMQP_URI": "memory://"})
+            def test_spam():
+                assert nameko.config["AMQP_URI"] == "memory://"
+
+        If `clear` is set to True then the config is completely replaced with the given
+        `context_config` otherwise it's only updated.
+
+        """
+
+        config = self.data
```
mattbennett: Might be clearer to call this `data`. I was briefly confused and thought the implementation below was writing to the _global_ config var.

iky: :+1:

nameko/__init__.py
iky: :+1:

PR #605, opened 2019-01-18
Minor change to add -c alternative to --config option

iky:
Just run Nameko `shell` intuitively with `-c` and found it does not work .)


timbu:
LGTM 👍

PR #610, opened 2019-03-11
Global config improvement

iky:
Nameko 2.X backward compatible passing of custom config arguments to `container_factory` and `runner_factory` test helpers. Config patch is applied if a config dictionary is passed to these factories.

### DIFF
--- a/test/testing/test_pytest_plugin.py
+++ b/test/testing/test_pytest_plugin.py
@@ -321,39 +321,38 @@ def test_container_factory_with_custom_container_cls(testdir, plugin_options):
     assert result.ret == 0
 
 
-@config.patch({'FOO': 1})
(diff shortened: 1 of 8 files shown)


test/testing/test_pytest_plugin.py
```diff
@@ -321,6 +321,41 @@ def test_container_factory(container_factory):
...
+    )
+
+    class Service(object):
+        name = "x"
+
+    assert config['FOO'] == 1
+
+    container_factory_fixture = container_factory_fixture()
+
+    container_factory = next(container_factory_fixture)
```
mattbennett: I think it would be easier to understand these tests if they used the `testdir.runpytest` approach like the other tests in this module. Is there a reason not to do that here?

PR #613, opened 2019-03-15
Deprecation warnings

mattbennett:
Some changes that reduce the warnings shown after running tests:

1. Stop setting the deprecated `auto_delete` flag on the Event exchange
2. Stop forcing all `DeprecationWarning`s to be logged
3. Ignore all `DeprecationWarning`s coming from libraries we cannot control
4. Detect `DeprecationWarning`s we do care about with `pytest.deprecated_call`
5. Silence any warnings that we intentionally trigger in the test suite. 

Required bumping pytest to a recent version.

Compare the build logs of https://travis-ci.org/nameko/nameko/jobs/515280316 (this branch) to https://travis-ci.org/nameko/nameko/jobs/513850376 (current master)

### DIFF
--- a/nameko/messaging.py
+++ b/nameko/messaging.py
@@ -42,10 +42,10 @@ class HeaderEncoder(object):
         data = worker_ctx.context_data
 
         if None in data.values():
-            _log.warn(
+            warnings.warn(
                 'Attempted to publish unserialisable header value. '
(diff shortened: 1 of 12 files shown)


mattbennett:
Branch is failing because it also needs to handle pyyaml upgrade. This is being addressed in other current PRs so waiting for them to land rather than fixing it here.

PR #614, opened 2019-03-16
Use generator for timer to avoid drift

bobh66:


nameko/timer.py
```diff
@@ -56,27 +56,28 @@ def kill(self):
 
     def _run(self):
         """ Runs the interval loop. """
-        sleep_time = 0 if self.eager else self.interval
 
+        def get_next_interval():
+            start = time.time()
+            count = 0
+            while True:
+                count += 1
```
mattbennett: Would be slightly terser to use an itertools counter here:

``` python
start = time.time()
for count in itertools.count(start=1):
    yield max(start + count * self.interval - time.time(), 0)
```

With this, `eager` also approximates `start=0`, which may or may not be a nicer way to implement it. Can't quite decide.

mattbennett:
Also, CI is failing here because of a recent release of pyyaml. This is being fixed in other branches so merging master after they land should fix the failures here.

PR #615, opened 2019-03-18
Describe and bump up version for v2.12.0 release

iky:


CHANGES
```diff
@@ -5,6 +5,19 @@ Here you can see the full list of changes between nameko versions. Versions
...
 increment the minor version number only.
 
+Version 2.12.0
+-------------
+
+Released: 2019-03-18
+
+* Refactor utils so standalone.events does not import eventlet (#580)
+* Compatibility with latest dependencies (moto #577, pyyaml and kombu #612)
+* Fix timer waiting (#579, #303) and improve it to avoid drift (#614)
```
mattbennett: Can we add a little more here? Something like "timer now waits for the spawned entrypoint to complete before firing again, as documented"

As @bobh66 [suggested on the forum earlier today](https://discourse.nameko.io/t/timer-interval-drift/409/8?u=mattbennett) we can add a `waits` flag in the near future (defaulting to `True`) so that people can opt back in to the old (technically buggy) behaviour if desired.

PR #616, opened 2019-03-19
V3.0.0 rc with latest master

iky:
Making 3.0.0 release candidate up to date with latest master changes.


setup.py
```diff
@@ -25,14 +25,15 @@
...
-        "pyyaml>=3.10",
+        "pyyaml>=5.1",
         "requests>=1.2.0",
         "six>=1.9.0",
         "werkzeug>=0.9",
         "wrapt>=1.0.0",
     ],
     extras_require={
         'dev': [
+            "astroid==1.6.5",
```
mattbennett: Is this here because we pin it? I don't think `astroid` is used directly

mattbennett: Oh yes, I see david's commit (https://github.com/nameko/nameko/pull/596/commits/1a72c70d10ae9c81eaa6590e9c8a1140d797ca85)

PR #628, opened 2019-04-10
Pytest fixtures Nameko 2.x backward compatible

juliotrigo:
Following up on the work done here https://github.com/nameko/nameko/pull/610, this would make the following fixtures backward compatible with Nameko `2.x`:

* `empty_config`
* `rabbit_config`
* `rabbit_ssl_config`
* `web_config`

Migrating Nameko services to `v3.0.0` it's easier now, since there's no need to keep backward compatibilitty in the tests with previous Nameko versions. However, trying to test Nameko extensions with both Nameko 2.x and 3.x is not trivial at the moment.

I'm trying to do that for a few dependency providers and extensions, in order to make sure they're tested with Nameko 2.x and and the same time compatible with the latest code on `3.0.0-rc`.

By making those fixtures also return the part of the config that they create, we can keep them backward compatible with Nameko `2.x` and testing extensions becomes much easier.

Any suggestions or comments are more than welcome!

### DIFF
--- a/nameko/testing/pytest.py
+++ b/nameko/testing/pytest.py
@@ -96,8 +96,9 @@ def always_warn_for_deprecation():
 
 @pytest.yield_fixture
 def empty_config():
-    with config.patch({}, clear=True):
-        yield
+    conf = {}
+    with config.patch(conf, clear=True):
+        yield conf  # nameko 2.X backward compatible
 
 
 @pytest.yield_fixture
@@ -182,8 +183,9 @@ def rabbit_uri(request, vhost_pipeline):
 
 @pytest.yield_fixture
 def rabbit_config(rabbit_uri):
-    with config.patch({'AMQP_URI': rabbit_uri}):
-        yield
+    conf = {'AMQP_URI': rabbit_uri}
+    with config.patch(conf):
+        yield conf  # nameko 2.X backward compatible
 
 
 @pytest.fixture
@@ -217,7 +219,7 @@ def rabbit_ssl_config(request, rabbit_ssl_uri, rabbit_ssl_options):
     }
 
     with config.patch(conf):
-        yield
+        yield conf  # nameko 2.X backward compatible
 
 
 @pytest.fixture
@@ -340,8 +342,9 @@ def web_config():
 
     port = find_free_port()
 
-    with config.patch({WEB_SERVER_CONFIG_KEY: "127.0.0.1:{}".format(port)}):
-        yield
+    conf = {WEB_SERVER_CONFIG_KEY: "127.0.0.1:{}".format(port)}
+    with config.patch(conf):
+        yield conf  # nameko 2.X backward compatible
 
 
 @pytest.fixture()
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .coverage
+.pytest_cache
 *.pyc
 _py_cache
 docs/build
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,7 @@ services:
   - docker
 
 before_install:
-    # start rabbitmq, grab client ssl certs
-    - docker run -d --rm -p 15672:15672 -p 5672:5672 -p 5671:5671 --name nameko-rabbitmq nameko/nameko-rabbitmq:3.6.6
-    - docker cp nameko-rabbitmq:/srv/ssl certs
+    - make rabbitmq-container
     # install toxiproxy
     - mkdir $PWD/bin
     - wget -O $PWD/bin/toxiproxy-server https://github.com/Shopify/toxiproxy/releases/download/v2.0.0/toxiproxy-server-linux-amd64
--- a/Makefile
+++ b/Makefile
@@ -37,3 +37,7 @@ spelling:
 
 linkcheck:
 	sphinx-build -W -b linkcheck -d docs/build/doctrees docs docs/build/linkcheck
+
+rabbitmq-container:
+	docker run -d --rm -p 15672:15672 -p 5672:5672 -p 5671:5671 --name nameko-rabbitmq nameko/nameko-rabbitmq:3.6.6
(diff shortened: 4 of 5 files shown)


juliotrigo:
I have also taken the liberty of moving the RabbitMQ container creation to the Makefile so that it's easier to perform this action when running the tests locally.

mattbennett:
Nice! These changes make sense to me.

@iky, do you agree?

iky:
Hi,

config fixtures used to yield configs for a while in a transition period of a few commits before removing it in https://github.com/nameko/nameko/pull/520/commits/115bb1d8e195e42783397837d6b22c07887231c3 ;) This PR brings it back somehow. I wonder if we should support it. I tried to run tests against Nameko 3.X in number of extensions and none of them need these fixtures. It looks like config fixtures are used mainly when testing services, but services are usually run against one chosen version of Nameko so you don't have to deal with backward compatibility.

@juliotrigo do you have an example of extension tests using Nameko config fixtures?

Some may use them lightly and in that case making it work for both major versions of Nameko is not hard. I hit a place when testing an extension where I have to deal with backward compatibility, but it wasn't config fixtures, it was `ServiceContainer` and making it working with both versions of Nameko  [wasn't difficult](https://github.com/iky/nameko-slack/commit/d34f628639b6e7fdbf53fe0c9f610e7baf2b7c43#diff-b946cfc9c7d6cd8ba6ce887aa8cdcd44).

nameko/testing/pytest.py
```diff
@@ -182,8 +183,9 @@ def rabbit_uri(request, vhost_pipeline):
 
 @pytest.yield_fixture
 def rabbit_config(rabbit_uri):
-    with config.patch({'AMQP_URI': rabbit_uri}):
-        yield
+    conf = {'AMQP_URI': rabbit_uri}
+    with config.patch(conf):
+        yield conf  # nameko 2.X backward compatible
```
iky: yielding just config may not work on places where the dictionary is changed in tests using the fixture I think. If we have to yield, it would be better to yield the whole `nameko.config` object.

iky: Here is an example where a test uses `rabbit_config`, but also uses it to write to it an unrelated `serialiser` setting:
https://github.com/nameko/nameko-amqp-retry/blob/master/test/test_retry.py#L458-L475

I'ts one of the places where returning only the passed part of the config may not work.

mattbennett:
I guess it is possible to work around these things as you are doing in https://github.com/nameko/nameko-amqp-retry/pull/30 @iky . The question is whether we should force library developers to do this or not. And if we don't, will we be stuck with these fixtures always being used in two ways.

What @juliotrigo has added here is not costly in terms of implementation, but there's a separate question about the on-going price of having more than one way of doing things.

juliotrigo:
@mattbennett I agree.

After discussing it with @iky and (re)thinking about it, this is how I see it.

My suggestion here was to make those pytest Nameko fixtures to just return the part of the config they are adding. If `rabbit_config` is used, then I find intuitive that this fixture also returns the RabbitMQ configuration that it's also being patched.

In that case, this is an example that could work for both Nameko `2.x` and `3.x`:

```python
@pytest.fixture
def foo_config():
    return {'a': 1, 'b': 2}


@pytest.fixture
def config(rabbit_config, web_config, foo_config):
    conf = {}
    conf.update(rabbit_config)
    conf.update(web_config)
    conf.update(foo_config)
    yield config


@pytest.fixture
def create_service(container_factory, config):
    # ...
    container = container_factory(MyService, config)
    # ...
```

or maybe also inspect/amend the config that we're about to use:

```python
def test_foo(config):
    config['foo'] = 'bar'

    container = container_factory(MyService, config)

   # ...
```

However, after @iky 's comment, I realised/remembered that most of those Nameko `2.x` fixtures return the whole config, instead of just the part of the config they are patching. This allows users to do something like [this](https://github.com/nameko/nameko-amqp-retry/blob/master/test/test_retry.py#L458-L475), already posted above, which, in theory [would not be backward compatible anyway](https://github.com/nameko/nameko-amqp-retry/pull/30/files).

So if we decide to go ahead with this, then we either return only the partial config that we're patching (which it feels more natural to me but it's also backward incompatible with Nameko `2.x`) or we keep returning the whole config for all the fixtures (which, as you said, gives us two ways of doing things and does not make it compatible for all cases).

Additionally, if we just make the following change to the example above (with the current `3.0.0-rc` code):

```python
@pytest.fixture
def config(rabbit_config, web_config, foo_config):
    conf = {}
    conf.update(rabbit_config or {})
    conf.update(web_config or {})
    conf.update(foo_config)
    yield config
```

then only the parts of the config created in our tests are passed to the `container_factory` and [patched](https://github.com/nameko/nameko/blob/v3.0.0-rc6/nameko/testing/pytest.py#L263-L269) there in Nameko `3.x` (`{'a': 1, 'b': 2, 'foo': 'bar'}` and not the Nameko fixtures since they `yield` `None`).

This makes backwards compatibility easier, since most of the tests (not all of them though) now pass for both Nameko `2.x` and `3.x`. Another questions is whether we think this should be the recomended way of doing it: that "partial" patching there is not that obvious but the change is minimal and it works as expected in most cases.

Anyway, we could probably add some guidelines about what's the recomended way of making extensions tests compatible for both Nameko major versions. I'm happy to help with any work that may be needed.

mattbennett:
Thanks @juliotrigo for this analysis.

Taking everything into account I think it's best to leave these fixtures as backwards-incompatible with Nameko 2.x. It's not difficult for library developers that want to support 2.x and 3.x to do so, and it will encourage upgrades.

I definitely agree we should describe the process as part of docs for  "upgrading to 3.x". There should be a section for upgrading services and a section for upgrading libraries.

juliotrigo:
That sounds good @mattbennett I'll close this PR.

Thank you both for your comments.

PR #632, opened 2019-04-17
Fix: Could not use pycharm debug when using eventlet.monkey_patch in pytest.py

daimon99:
Fix: Could not use pycharm debug when using eventlet.monkey_patch in pytest.py


mattbennett:
This change disables the eventlet monkey patch applied by the `pytest` plugin that ships with Nameko. It might solve your particular issue with PyCharm but it's not a general purpose "fix" -- in fact it will cause problems for all users of the `pytest` plugin.

If you're not using pytest with PyCharm, you could simply uninstall it. This would disable the plugin, which would have the same effect as your change here.

PR #636, opened 2019-05-12
Rename service name in running command

lypro09539:
This PR add an alternative para in running command to rename service class's name.

why it may be needed: I am working with others on same micro service and sharing same amqp, oftenly we will run service locally in which case service name must be changed to different name before running. And change back before push to svn. It maybe bring some convenience.

### DIFF
--- a/nameko/cli/commands.py
+++ b/nameko/cli/commands.py
@@ -102,8 +102,10 @@ class Run(Command):
             ' connected to for an interactive interpreter within the running'
             ' service process using `nameko backdoor`.')
         parser.add_argument(
-            '--cname', default='', type=str,
-            help='Cover service cls name')
+            '--rename', default='', type=str, nargs='+',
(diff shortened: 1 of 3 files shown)


davidszotten:
i would be hesitant to add api for this; it seems like a pretty rare use-case

i would recommend running your own rabbitmq instance for local dev. if that's not possible, a workaround might be to have a local-only custom module/subclass with a new name

e.g. 

```
cat renamed_service.py

from .real_service import Service as RealService

class Service(RealService):
    name = "renamed_service"
```

bobh66:
Wouldn't it be a lot simpler to just use different vhosts/users on the single RabbitMQ?  Assign each developer their own username and vhost in the RabbitMQ instance and override the AMQP_URI while in the dev environment.

mattbennett:
Thank you for your contribution @lypro09539, but I'm going to close this. I agree with @davidszotten that this is a very specific use-case.

PR #643, opened 2019-06-27
Pre-releasing v3.0.0-rc7

iky:
* porting a bug-fix from the stable release

### DIFF
--- a/test/test_container.py
+++ b/test/test_container.py
@@ -374,9 +374,6 @@ def test_container_only_killed_once(container):
     ) as kill_managed_threads:
 
         with patch.object(container, 'kill', wraps=container.kill) as kill:
-            # insert an eventlet yield into the kill process, otherwise
-            # the container dies before the second exception gets raised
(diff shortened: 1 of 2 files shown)


kooba:
So many good things in v3... can we cut the relase yet?

mattbennett:
@kooba I want to release it along with a docs refresh. There's a lot which has changed and the current docs are quite out of date with 3.x. I'm working on an outline -- would be great to have a community effort filling in the details ;-)

iky:
@kooba only a few doc entries missing! we should release soon.

kooba:
Point me to the outline... I'll be happy to pitch in some docs.

mattbennett:
@kooba https://www.notion.so/namekopython/Docs-Outline-64e72d7870d84292894d47cef61858b3

This is a fairly complete overhaul. We perhaps don’t need to do all of it before 3.0 but we need some of it at least (the intro, the restructure, and the "upgrading from 2.0" section)

PR #644, opened 2019-06-29
Add tests for python 3.7

daviskirk:
Since eventlet 0.25 fixed the main issues with running 3.7 (and eventlet also tests against 3.7) it's probably sensible to test nameko against 3.7 as well also.


.travis.yml
```diff
@@ -113,6 +125,9 @@ matrix:
     - stage: mastereventlet
       python: 3.6
       env: TOX_ENV=py36-mastereventlet
+    - stage: mastereventlet
+      python: 3.7
+      env: TOX_ENV=py36-mastereventlet
```
davidszotten: `py37`?

mattbennett: Typo here

.travis.yml
```diff
@@ -90,6 +90,18 @@ matrix:
     - stage: test
       python: 3.6
       env: TOX_ENV=py36-extra-lib
+    - stage: test
+      python: 3.7
+      env: TOX_ENV=py37-oldest-lib
+    - stage: test
+      python: 3.7
+      env: TOX_ENV=py37-pinned-lib
+    - stage: quick_test3
```
mattbennett: I think the idea behind the `quick_test` is that there's one variant on each Python version that runs first. So we should drop either this one or the python 3.6 one. Probably best to keep the latest version

mattbennett:
@davidszotten there is a discussion about python versions [on the forum](https://discourse.nameko.io/t/python-versions-to-be-supported-by-nameko-v3-x/415/3)

daviskirk:
There are a few more problems after all with eventlet.  I'll leave this pull request as it is until those are fixed and then continue

.travis.yml
daviskirk: changed all the "single py2/py3" tests to only use py37

PR #645, opened 2019-06-30
Skip ssl tests if certs are not available or ssl port is not open

daviskirk:
These are a few possibilities for implementing https://www.notion.so/Skip-RabbitMQ-SSL-tests-if-secure-port-not-available-73a50e2111704b50b01fabfb27683879

1. Check for the cert files and skip if they are not found
2. Check the configured host and check if ssl port is available and skip if it is not

I would say the first is probably simpler and might be enough (considering that if you went through the trouble of copying over the ssl certs you probably also have the correct docker container running), but I thought I'd put all the ideas in here (maybe there's also something completely different I didn't think of).

### DIFF
--- a/nameko/testing/pytest.py
+++ b/nameko/testing/pytest.py
@@ -1,9 +1,5 @@
 from __future__ import absolute_import
 
-import os
-from contextlib import closing
-
-import eventlet
 import pytest
 
 from nameko import config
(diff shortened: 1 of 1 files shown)


davidszotten:
like with the toxiproxy tests we probably want to prevent the skip if we detect travis

daviskirk:
As suggested, I removed the port checking and added the `on_travis` check

PR #648, opened 2019-07-10
Fix pytest funcargnames warnings

adamchainz:
I'm testing with Pytest 5.0.x and seeing this warning three times for every test in my suite that uses nameko's fixtures:

```
============================== warnings summary ==============================
tests/integration/test_nameko.py::test_home
tests/integration/test_nameko.py::test_home
tests/integration/test_nameko.py::test_home
...
/private/tmp/tox/py37-django22/lib/python3.7/site-packages/nameko/testing/pytest.py:274: PytestDeprecationWarning: The `funcargnames` attribute was an alias for `fixturenames`, since pytest 2.3 - use the newer attribute instead.
  if fixture in request.funcargnames:
```

This fixes that warning. I can't find documented the minimum version of Pytest that Nameko intends to support, but Pytest 2.3 was released [Oct 19, 2012](https://pypi.org/project/pytest/2.3.0/) so it's getting a bit old.

### DIFF
--- a/nameko/testing/pytest.py
+++ b/nameko/testing/pytest.py
@@ -265,12 +265,8 @@ def fast_teardown(request):
 
     reorder_fixtures = ('container_factory', 'runner_factory', 'rabbit_config')
     for fixture in reorder_fixtures:
-        if fixture in request.funcargnames:
-            # getfuncargvalue is renamed to getfixturevalue in pytest 3.0
-            if hasattr(request, 'getfixturevalue'):
-                request.getfixturevalue(fixture)  # pragma: no cover
-            else:
-                request.getfuncargvalue(fixture)  # pragma: no cover
+        if fixture in request.fixturenames:
+            request.getfixturevalue(fixture)  # pragma: no cover
 
     consumers = []
 


nameko/testing/pytest.py
```diff
@@ -265,7 +265,7 @@ def test_bar(rabbit_config, container_factory):
 
     reorder_fixtures = ('container_factory', 'runner_factory', 'rabbit_config')
     for fixture in reorder_fixtures:
-        if fixture in request.funcargnames:
+        if fixture in request.fixturenames:
             # getfuncargvalue is renamed to getfixturevalue in pytest 3.0
```
mattbennett: Pytest 3.0 was released in 2016, which seems plenty old enough. Let's commit to requiring >3.0 and then we could remove this conditional as well

adamchainz:
🚀

nameko/testing/pytest.py
ravish510: I have pytest-6.0.1, still getting the same warning as stated by adamchainz.
please help me @mattbennett .

dubesoftware: > I have pytest-6.0.1, still getting the same warning as stated by adamchainz.
> please help me @mattbennett .

I'm getting the same running PyTest v 6.0.1. I wonder why your comment was marked as spam.

antl31:
Hello guys,  i have issue https://github.com/nameko/nameko/issues/725

PR #649, opened 2019-07-10
Fix Python 3.7 warning on collections.abc

adamchainz:
Python 3.3 moved the ABC's from `collections` into `collections.abc`, Python 3.7 deprecates the use of them at the original location. Hence in my tests I saw Pytest gathering the warnings emitted from `collections.abc`:

```
tests/integration/test_nameko.py::test_server_error
  /private/tmp/tox/py37-django22/lib/python3.7/site-packages/nameko/exceptions.py:76: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    if isinstance(value, collections.Iterable):
```

This PR fixes it to use a conditional import based on the Python version.

I see Python 3.7 will be tested in #644, but this fix isn't in there.

### DIFF
--- a/nameko/exceptions.py
+++ b/nameko/exceptions.py
@@ -5,6 +5,7 @@ import sys
 
 import six
 
+
 if sys.version_info >= (3, 3):
     from collections.abc import Iterable
 else:
--- a/nameko/exceptions.py
+++ b/nameko/exceptions.py
@@ -1,9 +1,14 @@
 from __future__ import unicode_literals
 
-import collections
 import inspect
+import sys
 
 import six
+
+if sys.version_info >= (3, 3):
+    from collections.abc import Iterable
+else:
+    from collections import Iterable
 
 
 class ExtensionNotFound(AttributeError):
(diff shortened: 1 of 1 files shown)


nameko/exceptions.py
```diff
@@ -1,10 +1,15 @@
 from __future__ import unicode_literals
 
-import collections
 import inspect
+import sys
 
 import six
 
+if sys.version_info >= (3, 3):
```
mattbennett: Nameko only supports 3.4 upwards (and even 3.4 is being removed in the 3.x release) so I think we can skip this conditional and just keep the updated import.

adamchainz: Sorry I'm using 2.12.0 which still supports Python 2, didn't realize master had dropped Python <3.4 support, fixed!

mattbennett: We will still support Python 2 in Nameko's 3.x release actually, but are dropping Python 3 versions that are already end of life.

Thanks for the PR!

adamchainz: I've put it back as 3.3+ because that's when the feature was introduced

mattbennett: Perfect, thank you. Sorry for the unnecessary back and forth

PR #651, opened 2019-07-21
Click config fix

mattbennett:
Converting the CLI to click (#620) had an unintended consequence of moving the service import before the config parsing. As a result services could no longer read configuration at import time.

This PR adds a failing test and then the fix. It also adds an extra test for importing multiple services at once, and updates one broken example in a docstring.

### DIFF
--- a/test/cli/test_run.py
+++ b/test/cli/test_run.py
@@ -7,7 +7,7 @@ from textwrap import dedent
 
 import eventlet
 import pytest
-from mock import call, patch
+from mock import patch
 
 from nameko import config
 from nameko.cli.run import run, setup_backdoor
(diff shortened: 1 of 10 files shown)


test/cli/test_run.py
```diff
@@ -151,6 +151,63 @@ def test_main_with_logging_config(command, tmpdir):
...
+    ))
+
+    service_file = tmpdir.join("service.py")
+    service_file.write(dedent(
+        """
+        from nameko import config
+        from nameko.testing.services import dummy
+
+        class Service:
+            name = config.get("NAME", "notfound")
```
iky: nice, this test is actually a use case somebody asked for in the discussion board, or in one of issues here on github .)

chewfred:
Just found this issue as well. Thanks

PR #652, opened 2019-07-26
WIP: Add client and server timeouts to rpc

daviskirk:
This is an implementation of the suggested feature mentioned here: https://www.notion.so/Full-support-for-RPC-timeouts-592cd7903cc84a049be11abfe778d5ce

There are still some additional tests missing but the basic concept would be to add to arguments to all rpc clients (standalone and dependency provider):

* timeout: will raise a timeout if when the total roundtrip time is exceeded (raised in at the callers machine)
* delivery_timeout: Uses rabbitmqs "expiration" to remove undelivered messages from the server.

I built in logic to try and prevent race conditions (message removed from server and client is still waiting forever for an answer).

As always there are a few open questions I'd like to discuss with the maintainers (@mattbennett ?):

The current implementation no longer uses the socket timeout on the standalone rpc client (as suggested in the roadmap).  However, it now relies on the timeout being raised by the rpc client.  Previously (as far as I can tell) the standalone client did not rely on eventlet at all, but now the rpc client uses this to implement timing out after waiting for a reply for a while.  The question is whether this is ok, or if this was a conscious decision to allow non-monkeypatched clients to still use the standalone rpc client.


nameko/rpc.py
```diff
@@ -607,9 +630,21 @@ def _call(self, *args, **kwargs):
...
 
+    def _consolidate_timeouts(self, timeout, delivery_timeout):
+        if delivery_timeout is not None and timeout is None:
+            raise ValueError(
+                '`delivery_timeout` is given. '
+                'An explicit non-null `timeout` argument is required.'
+            )
+        if timeout is not None and delivery_timeout is None:
+            # If a timeout is given, we might as well remove the message from
+            # the server.
```
mattbennett: Not sure this is valid. It is semantically different for the client to say _I no longer care about the result_ and _I don't care if this work is carried out_.

daviskirk: Yes, I guess this is correct.  I'll take it out, thanks for catching it.  Along those lines:  is the naming ok (I have both of them named "timeout" to make clear their related, but we could also name it "expiration" is that is essentially what "delivery_timeout" does, pass that argument to the "expiration" header of rabbitmq.

mattbennett: gRPC calls them client and server deadlines. I think in gRPC the server will actually abort a request in progress if the deadline is reached though.

I quite like "expiration" honestly. It lines up with the AMQP name for what happens to the message, and the server timeout will only be actioned here if the message is still queued in the broker when it's reached.

daviskirk: Ok :+1:, unfortunately while working on this I stumbled on a few more tricky design decisions resulting in some more questions:

* "expiration" can already be set just by using the normal publisher options.  We could just leave that as it is.  The only disadvantages to this are:
  1. The option is for the entire rpc client (if you call different methods the same "expiration" would be set on all method calls).  
  2. The API might be a bit less intuitive especially once [Separate Publish & Consume options](https://www.notion.so/12116ff4a4b14d1ba0cf52e2faa84ea9?v=22ddef13d2774fedb53e9eea7572187a&p=83a2aefead1d45318b341913897f4cd0) get done.

In order to have both timeout and expiration to be set per call, we would need to have the options within the client.  Currently, my idea would be to have a seperate "publish_options" argument that would be passed to the `Client` object (for expiration), and pass the "timeout" argument to the client as well.  If one wanted to change the timeout per call, one would need to do something like:

```python
class TestService:
    name = 'test'
    other_service = ServiceRpc('other_service, timeout=2.0, publisher_options={'expiration': 5.0})
    
    def run(self):
        other_service.method1()  # called with expiration 5 and timeout 2
        method2 = other_service.method2
        method2.publish_kwargs['expiration'] = 10.0
        method2.timeout = 5
        method2()  # called with expiration 10.0, timeout 2
```

which is fine (even if it looks a bit ugly) ... but we can also cause a deadlock doing this since the "publisher_options" set in the main service are not passed the the client and we can set an expiration without setting a timeout, causing the call to wait forever:

```python
class TestService:
    name = 'test'
    other_service = ServiceRpc('other_service, timeout=2.0, publisher_options={'expiration': 5.0})
    
    def run(self):
        other_service.method1()  # called with expiration 5 and timeout 2
        method2 = other_service.method2
        method2.timeout = None
        method2()  # called with expiration 5.0 as set be the publisher, the client does not know this, so it will allow timeout to be set to `None`
```

The only way I can think of to work around this is to not allow the publisher options to be bound to the publisher at all and ALWAYS set them when "publish" is actually called, (which would allow us to check consistency when the call occurs).

Alternatively we can just not support setting timeouts per call and bind timeout and expiration to the `ClusterProxy` level and check everything there.

mattbennett: Yes, this is a good point.

I think we should try to get the API right and then work from there.

I would want the existing publishers to continue to allow "expiration" to be passed. Both as a default when declaring it and optionally per-message.

The RPC client should have its own API and we should perhaps delineate the ones that apply to the underlying publisher. Something like:

```
client = RpcClient("service", timeout=10, expiration=20, publisher_options={}, consumer_options={})
```

`publisher_options` applies to the client publisher
`consumer_options` applies to the reply listener consumer

(It's arguable whether being able to change the consumer options of the reply listener is a good idea or not, but I'll leave it here for now)

We'd need to decide what to do when `publisher_options` includes `expiration` -- either ignore, warn or override the other value. I would probably warn and ignore.

Ideally every option could be provided at declaration time and also at method-call time. There is a general problem of how to do that with the RPC client though, because args and kwargs are both passed through to the remote function. In hindsight this was a bad design decision. As a workaround I generally favour using a context manager. Something like:

```
client = RpcClient("service", timeout=10, expiration=20, publisher_options={}, consumer_options={})

client.method()  # 10 second timeout
with RpcClient.override(client, timeout=20) as client:
    client.method()  # 20 second timeout
```

To avoid the deadlocked client issue I would just check at method-call time whether you've set an expiry without a client timeout, and raise an exception if so. If we were really clever we'd raise the exception at the end of the expiration period, if the call hadn't returned. That way the only effect of the safety mechanism would be that you'd get a a helpful exception in the client rather than waiting forever.

daviskirk: Regarding 

```python
client = RpcClient("service", timeout=10, expiration=20, publisher_options={}, **publisher_options_kwargs)
```

I tried to implement this as suggested:
* `publisher_options_kwargs` still works but is now deprecated.  If we agree on everything I'd change the tests to use the new api accordingly as well.  
* `timeout` and `expiration` are explicit options and can be changed at any time up until a method call (they are basically passed all the way to the `rpc.Client` object).
* I did not implement the consumer options yet as it seems like a separate problem.


> We'd need to decide what to do when publisher_options includes expiration -- either ignore, warn or override the other value. I would probably warn and ignore.

Did this as suggested (warn and ignore).

> Ideally every option could be provided at declaration time and also at method-call time. [...] As a workaround I generally favor using a context manager.

Playing around with this I feel like a context manager is a bit clumsy since we're just changing attributes of the client object, but then creating a new client object anyway when we select a method.  Instead I thought a simple method might be simpler, so the current version implements the following api:

```python
self.service.method()  # normal method call
self.service.with_options(timeout=1).method()  # using timeout 1
```

the `with_options` name could also perhaps be changed to `using` resulting in something like `self.service.using(timeout=1).method()` ?

> To avoid the deadlocked client issue I would just check at method-call time whether you've set an expiry without a client timeout, and raise an exception if so

Did this

> If we were really clever we'd raise the exception at the end of the expiration period, if the call hadn't returned

Considering your previous comment about not implicitly setting the timeout if it's left out and thinking about this a bit I think it would be better to raise the exception directly at call time since I think it implies that the caller probably misconfigured the client and it's better to know this early (earliest point that this is the case is call time, since up until that time it could be changed).
Open to other interpretations (don't think it would be impossible to raise the exception after the expiration has run out). 

### Final tricky questions 

One thing that is still a bit tricky is this:

Obviously most of the `publisher_options` apply to the underlying publisher but can just as well apply to the `publish` method (where they then override the publisher options), so we are explicitly deciding that `expiration` can be set at call time but `priority`  and `headers` can not. 

If we would want to do something like:

```python
self.service.with_options(timeout=2, headers={'a': 3}, priority=3).method()
```

would change the api to:

```python
client = RpcClient("service", timeout=10, expiration=20, headers=None, priority=None, publisher_options={}, **publisher_options_kwargs)
```
??

daviskirk:
> For now though people rely on it working without a monkey patch. Can you implement a native-thread based timeout mechanism?

I used the builtin "timeout" argument in the consumer and added a test `test_timeout_with_multiple_calls` for the scenario mentioned in https://www.notion.so/Full-support-for-RPC-timeouts-592cd7903cc84a049be11abfe778d5ce (regarding the standalone rpc proxy implementation).  
If this is not enough I can try to implement a threading based approach (I tried that first but had problems as there's no easy way to kill native threads in python which caused problems for me trying to implement a threading based timeout ... I'm admittedly the opposite of proficient when it comes to using native threads in python though).

nameko/rpc.py
```diff
@@ -561,12 +621,31 @@ def call_async(self, *args, **kwargs):
         rpc_call = self._call(*args, **kwargs)
         return rpc_call
 
+    def with_options(self, **options):
+        """Create a new client with modified options."""
```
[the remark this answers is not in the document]
daviskirk: In fact I started building it as one.  I'll try and explain how we came to the conclusion that a direct version might be better:

* A context manager implied "context" that is wrapped in a block.  In this case the client is truly a new object independent from the old object and there is no entering or closing functionality necessary, so there is no real reason for a context manager.
* For many options we more or less want to use it for a single call (especially timeouts, expiry which might be different for every call).  In the case of a normal method, there's nothing keeping use from making a single line method call, which makes for uncluttered code especially in method heavy code where various timeouts will be applied very generously (in our case almost every call will have different options).  Especially when these methods are used in a couple of ways, it's really easy to get really nested code without much benefit:

```python
with self.service_rpc.with_options(timeout=5) as client1:
    client.method('this', 'that')

    with self.service_rpc.with_options(timeout=7, expiration=2) as client2:
        client2.method2('something') 

        with client2.with_options(timeout=4) as client3:
             client3.method2('something else') 

    client1.method('first again')
```

```python
f = self.service_rpc.method.with_options(timeout=5)
client2 = self.service_rpc.with_options(timeout=7, expiration=2)

f('this', 'that')
client2.method2('something')
client2.with_options(timeout=4).method2('something else')
f('first again')
```

Most of our code my code that I tried tried to test this with got to be nested 2,3 or 4 levels deep fairly quickly and using an exit stack (https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack) works but doesn't make the code look nicer (and as there is no real context to close without any real benefit as far as I can tell)

nameko/rpc.py
```diff
@@ -561,12 +621,31 @@ def call_async(self, *args, **kwargs):
...
             raise ValueError(
                 "Cannot call unspecified method {}".format(self.identifier)
             )
 
+        if (
+                self.expiration is not None and
+                self.timeout is None
+        ):
+            # TODO: Figure out how to check this if "expiration" is set on the
+            # publisher itself, which we don't have access to here.
```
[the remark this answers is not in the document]
daviskirk: Yes this was the kind of thing I was concerned about.
However, in the mean time I think this case might be covered:

The publisher option "expiration" is also passed to the client, so in this case while the client could not inspect the publisher, the client itself would actually have the expiration option set to the same value as the publisher and would trigger this `ValueError`.

While this isn't 100% "safe" since someone could later change the code without realizing this can cause problems, I think it's probably good enough if it works, but I will make extra sure these cases are tested (which would also guard against changes) and change the comment appropriately if that sounds ok?

test/test_rpc.py
```diff
@@ -813,6 +836,149 @@ def echo(self, arg):
             missing("foo")
 
 
+@pytest.mark.usefixtures("rabbit_config")
+@pytest.mark.parametrize("override", [None, 'service', 'method'])
+@pytest.mark.parametrize("timeout, async_delay, call_duration, error", [
+    pytest.param(0.1, None, 1, True,
+                 id="long duration with short timeout"),
```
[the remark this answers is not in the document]
daviskirk: I think it's fairly new (just a few years in pytest), I like it too, it makes the test output a lot nicer and easier to understand

test/test_timers.py
```diff
@@ -21,10 +21,10 @@ def tracker():
...
     (5, True, 0, 1),
     # call duration delays second call
-    (0.1, False, 0.3, 1),  # fires at 0.1
-    (0.1, True, 0.3, 1),  # fires at 0, 0.2
-    (0.025, False, 0.07, 2),  # fires at 0.25, 0.095, 0.165
-    (0.025, True, 0.07, 3),  # fires at 0, 0.07, 0.14
+    (0.1, False, 0.3, 1),  # fires at 0.1, 0.4
+    (0.1, True, 0.3, 1),  # fires at 0, 0.3
+    (0.04, False, 0.06, 2),  # fires at 0.04, 0.1, 0.16
+    (0.04, True, 0.06, 3),  # fires at 0, 0.06, 0.12
```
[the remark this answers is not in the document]
daviskirk: These were failing once in a while on the CI server.
Locally they always work but I think and I made them as short as possible and the delay on the server seems to be a tiny bit longer.
Only noticed it because I was pushing to this branch a lot.

chewfred:
any update on this PR. I think timeout on RPC seems to be a good idea

daviskirk:
@chewfred we're waiting for the nameko 3 release (sorry I can't find the thread where we decided this).  After 3.0 get's out of rc we'll complete this.

litnimax:
Hi folks!
Will timeouts be supported in 3.0?
Thx.

PR #662, opened 2019-10-13
Add --version CLI flag

bow:
This PR adds a `-v` and `--version` flag to the base `nameko` CLI.

```
$ nameko -v
2.12.0

$ nameko --version
2.12.0
```

Let me know if the version output should be changed (e.g. `nameko v2.12.0`) and/or if there should also be tests for this :).

### DIFF
--- a/nameko/cli/main.py
+++ b/nameko/cli/main.py
@@ -4,9 +4,9 @@ import argparse
 import os
 import re
 from functools import partial
-from pkg_resources import get_distribution
 
 import yaml
+from pkg_resources import get_distribution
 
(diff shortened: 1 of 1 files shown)


vlcinsky:
@bow I think, your PR is built against master branch.

For next version to be released, it shall be built against branch `v3.0.0-rc`, where the CLI is based on `click` library. I supports this feature as documented here http://click.palletsprojects.com/en/7.x/api/?highlight=version#click.version_option.

mattbennett:
This is great, thanks @bow. Thanks also @vlcinsky for your equivalent PR to the 3x branch.

Issue #664, opened 2019-10-18
(3.0.0-rc) nameko fails, if AMQP url is not configured


vlcinsky:
**Describe the bug**
run `nameko shell` without having broker url configured.

The command fails with `KeyError: AMQP_URI'`

**To Reproduce**
Make sure, broker (AMQP) uri is not defined/configured.

Run:
```
$ nameko shell
Traceback (most recent call last):
  File "/home/javl/sandbox/nameko/.tox/py36-pinned-lib/bin/nameko", line 11, in <module>
    load_entry_point('nameko', 'console_scripts', 'nameko')()
  File "/home/javl/sandbox/nameko/.tox/py36-pinned-lib/lib/python3.6/site-packages/click/core.py", line 764, in __call__                                                                                          
    return self.main(*args, **kwargs)
  File "/home/javl/sandbox/nameko/.tox/py36-pinned-lib/lib/python3.6/site-packages/click/core.py", line 717, in main                                                                                              
    rv = self.invoke(ctx)
  File "/home/javl/sandbox/nameko/.tox/py36-pinned-lib/lib/python3.6/site-packages/click/core.py", line 1137, in invoke                                                                                           
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/javl/sandbox/nameko/.tox/py36-pinned-lib/lib/python3.6/site-packages/click/core.py", line 956, in invoke                                                                                            
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/javl/sandbox/nameko/.tox/py36-pinned-lib/lib/python3.6/site-packages/click/core.py", line 555, in invoke                                                                                            
    return callback(*args, **kwargs)
  File "/home/javl/sandbox/nameko/nameko/cli/__init__.py", line 83, in shell
    main(interface)
  File "/home/javl/sandbox/nameko/nameko/cli/shell.py", line 88, in main
    config[AMQP_URI_CONFIG_KEY]
  File "/home/javl/sandbox/nameko/.tox/py36-pinned-lib/lib/python3.6/collections/__init__.py", line 991, in __getitem__                                                                                           
    raise KeyError(key)
KeyError: 'AMQP_URI'
```

**Expected behavior**
CLI shall either report cause of the problem (broker uri is not defined), or it shall accept such condition and follow up running, possibly printing warning or information about broker being unconfigured.

**Environment (please complete the following information):**
 - Nameko version: branch v3.0.0-rc
 - Python version: 3.6.9
[...]

PR #665, opened 2019-10-18
CLI nameko shell survives unconfigured broker (but prints warning)

vlcinsky:
Resolve issue #664 by surviving attempt to read unconfigured AQMP uri and by printing "Broker: undefined!!!" at shell banner.

### DIFF
--- a/nameko/cli/shell.py
+++ b/nameko/cli/shell.py
@@ -84,7 +84,7 @@ def main(interface):
     banner = "Nameko Python %s shell on %s\nBroker: %s" % (
         sys.version,
         sys.platform,
-        config[AMQP_URI_CONFIG_KEY],
+        config.get(AMQP_URI_CONFIG_KEY, "undefined!!!"),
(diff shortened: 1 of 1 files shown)


mattbennett:
Thank you for this @vlcinsky.

I wonder if it makes more sense to raise an error if the broker is undefined though, or use a default value? `nameko shell` is currently only useful for interacting the AMQP-enabled Nameko services, so perhaps we should just fail loudly with a clear error?

vlcinsky:
@matbennet I was considering "fail loudly" option too.

The reason I prefer to continue (and print "undefined!!!" in banner) is, that it allows to run shell and investigate current configuration. Imagine the case, when someone things, his configuration defines the broker, but has there some typo in name of the variable or within the broker url.

With current "let shell go on" the user has a chance to explore the config.

Let me know your preference, I will do eventual modification to the code, if you prefer to fail the call.

PR #676, opened 2020-03-17
added DLX retry delay

sanek9:
https://medium.com/@dotbox/delayed-requeuing-with-rabbitmq-dcbdf0026bf0


joaomcarlos:
Theres a nameko retry package that has retry functionality. I am not sure if its exactly the same implementation though

PR #680, opened 2020-05-04
Fix builds

mattbennett:


setup.py
```diff
@@ -79,8 +79,8 @@
         "Programming Language :: Python :: 2",
         "Programming Language :: Python :: 2.7",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.3",
-        "Programming Language :: Python :: 3.4",
+        "Programming Language :: Python :: 3.5",
+        "Programming Language :: Python :: 3.6",
```
davidszotten, line 83: time to try some newer pythons? (i guess the limiting factor will be eventlet support)

mattbennett: Yes indeed. It's on the list.

https://github.com/nameko/nameko/pull/644 got stuck on [this eventlet bug](https://github.com/eventlet/eventlet/issues/526) which I have been vaguely following. Seems like there may be a workaround we can try.

Issue #692, opened 2020-07-26
ClusterRpcProxy using default credentials even with given config


asciifaceman:
**Please use the [Discourse forum](https://discourse.nameko.io/) for support requests and questions.**

**Describe the bug**
Seemingly no matter what I do, `ClusterRpcProxy` uses the default guest:guest configuration.

```
    with ClusterRpcProxy(config={"AMQP_URI":"pyamqp://notguest:password@127.0.0.1"}) as rpc:
        result = rpc.stream_service.get(stream_id)
        return result, 200
```
results in

```
AMQPLAIN login refused: user 'guest' - invalid credentials
2020-07-26 03:15:34.787 [info] <0.7322.1> closing AMQP connection <0.7322.1> (172.25.0.1:60004 -> 172.25.0.2:5672)
2020-07-26 03:16:58.766 [info] <0.7505.1> accepting AMQP connection <0.7505.1> (172.25.0.1:60048 -> 172.25.0.2:5672)
2020-07-26 03:16:58.771 [error] <0.7505.1> Error on AMQP connection <0.7505.1> (172.25.0.1:60048 -> 172.25.0.2:5672, state: starting):
AMQPLAIN login refused: user 'guest' - invalid credentials
```

I have tried a series of variations of this (from instantiating the config different ways, with or without explicit definition) and have been unsuccessful. 

Additionally, I was trying originally to use `config.get()` but it appears to not be set outside of a nameko run service context. This may just be an expected behavior I'm unaware of tho.

I have no issues so far using the nameko CLI to run the rpc calls manually, so config is working within the context of a nameko service.

**To Reproduce**
```
from flasgger import Swagger
from flask import Flask, jsonify, request
from nameko.rpc import RpcProxy
from nameko.standalone.rpc import ClusterRpcProxy

app = Flask(__name__)
Swagger(app)

@app.route('/stream/<stream_id>', methods=['GET'])
def get(stream_id):
    """Returns a single stream
    ---
    parameters:
        - name: stream_id
          in: path
          type: string
          required: true
    responses:
        200:
            description: A stream represented by json
    """

    config= {
        'AMQP_URI': 'pyamqp://user:password@127.0.0.1'
    }
    
    with ClusterRpcProxy(config) as rpc:
        result = rpc.stream_service.get(stream_id)
        return result, 200

app.run(debug=True)

```

**Expected behavior**
For the RPC call to be made via ampq

**Environment (please complete the following information):**
 - Nameko version: 3.0.0rc8
[...]

Issue #687, opened 2020-07-17
Bug when using ClusterRpcClient


tyler46:
**Describe the bug**
It seems that If I pass an **amqp uri**  different that ampq://guest:guest@localhost:5672, I get a `ConnectionRefusedError: [Errno 111] Connection refused`. The full traceback can be found bellow:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/amqp/transport.py", line 138, in _connect
    host, port, family, socket.SOCK_STREAM, SOL_TCP)
  File "/usr/local/lib/python3.6/socket.py", line 745, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -9] Address family for hostname not supported

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/app/cron/scheduler.py", line 71, in <module>
    main()
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/app/cron/scheduler.py", line 67, in main
    process()
  File "/app/cron/scheduler.py", line 25, in process
    with ClusterRpcClient(uri=settings.AMQP_URI) as cluster_rpc:
  File "/usr/local/lib/python3.6/site-packages/nameko/standalone/rpc.py", line 262, in __enter__
    return self.start()
  File "/usr/local/lib/python3.6/site-packages/nameko/standalone/rpc.py", line 268, in start
    self.reply_listener.start()
  File "/usr/local/lib/python3.6/site-packages/nameko/standalone/rpc.py", line 94, in start
    ch = conn.channel()
  File "/usr/local/lib/python3.6/site-packages/kombu/connection.py", line 289, in channel
    chan = self.transport.create_channel(self.connection)
[...]

PR #694, opened 2020-07-27
Add conifg.setup() to standalone rpc example

asciifaceman:
Adds `config.setup()` to the given example for a standalone RPC call, which should make the example actually operational.

I didn't find `config.setup()` documented anywhere else in my searches when I was trying to get this to work.

### DIFF
--- a/docs/examples/standalone_rpc.py
+++ b/docs/examples/standalone_rpc.py
@@ -1,8 +1,11 @@
+from nameko import config
 from nameko.standalone.rpc import ClusterRpcProxy
 
-config = {
+
+conf = {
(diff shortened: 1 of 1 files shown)


asciifaceman:
This should prevent more problems like 

https://github.com/nameko/nameko/issues/692

https://github.com/nameko/nameko/issues/687

asciifaceman:
I don't understand what failed in the build but happy to fix it if anyone does

chewfred:
I just found this issue as well :-( Debugged half a day... i think i will switch out not to use rc release

pace-gene:
This is fine and can be merged for `master` but isn't applicable to the upcoming 3.0.0 release.

PR #704, opened 2020-09-14
use a fake rather than the real service

mattbennett:
This test fails occasionally because it was hitting the real Travis web service. Now uses a fake instead.


iky (approved):
:+1:

PR #711, opened 2020-12-26
Improve standalone RPC more reliable and fix nameko shell loading

tranvietanh1991:
Allow me to introduce my self, i am leader developer of a start up company which recently got acquired in Singapore. The core micro services internal RPC system of the product we developed was built using this nameko framework. In the past 3 years i've put a few changes to the framework to fix a couple of reliability issue (most are network issue). The main changes are the **Changes 1** mentioned below, which is to fix many weird network issues we were encountering on Production environment (RabbitMQ 3.6.x). Unfortunately there is no easy way to write a test-case for all these network issue so this PR may have to relied on the people review, comment and trust. Hope the owner of `nameko` framework can take a look in this PR and give some comment for all the changes implemented here.

**p/s**: I'm also the active maintainer/owner of these libraries: 
+[django-nameko](https://github.com/and3rson/django-nameko) : RPC connection pool implementation to use in conjunction with Nameko standalone RPC Proxy in any multi-threaded based non-nameko services such as: Django View, Django-Restframework API, Flask, ..)
+[nameko-django](https://github.com/tranvietanh1991/nameko-django) : A custom kombu serializer which does auto convert Django ORM/Queryset/DateTime/Date/Decimal using msgpack which can be configure to used in nameko services

### Changes 1: Improve standalone RPC to be more reliable
+ Add support for heartbeat in standalone RPC
+ Auto recover connection and retry if encounter network failure (only issue which know to be recoverable)
+ Change the behavior of PollingQueueConsumer to ack message on received instead of when result() method is called (this is to match the behavior of non-standalone rpc QueueConsumer)
### Changes 2: Fix nameko shell to load correctly
+ Fix nameko shell to properly import and init ClusterRpcProxy after PYTHONSTARTUP script loaded and ShellRunner started
+ Add current working directory to System path when start nameko shell (This is to solved auto import feature of other framework such as: Django ORM)
### Changes 3: Fix compatible issue with system was running older version of nameko (prior to 2.12)
+ Add back `auto_delete=True` option when init Exchange for events to fix compatible with system was running older version of nameko. Also this feature is not deprecated from AMQP 0-9-1 anymore (see: https://github.com/celery/py-amqp/issues/286)

### DIFF
--- a/nameko/cli/shell.py
+++ b/nameko/cli/shell.py
@@ -6,30 +6,28 @@ import yaml
 
 import nameko.cli.code as code
 from nameko.constants import AMQP_URI_CONFIG_KEY
-from nameko.standalone.events import event_dispatcher
-from nameko.standalone.rpc import ClusterRpcProxy
 
 from .commands import Shell
 
 
 class ShellRunner(object):
 
-    def __init__(self, banner, local):
+    def __init__(self, banner, get_local):
         self.banner = banner
-        self.local = local
+        self.local = get_local
 
     def bpython(self):
         import bpython  # pylint: disable=E0401
-        bpython.embed(banner=self.banner, locals_=self.local)
+        bpython.embed(banner=self.banner, locals_=self.local())
 
     def ipython(self):
         from IPython import embed  # pylint: disable=E0401
-        embed(banner1=self.banner, user_ns=self.local)
+        embed(banner1=self.banner, user_ns=self.local())
 
     def plain(self):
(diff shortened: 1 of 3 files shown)


tranvietanh1991:
Its seem travis-ci build is failed with Docker pull error: ```anyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit.```

PR #713, opened 2020-12-30
provide a way to config login_method

skycastlelily:
users are not able to config login_method with current code,if the
default one(AMQPLAIN) of amqp/connection doesn't work for them


skycastlelily:
My service need to use EXTERNAL login_method which is not the default one.
I've tested the code on my machine.

skycastlelily:
Hi, 
I have updated the code to make static  check pass.

mattbennett:
Thanks for this PR @skycastlelily. I can see that there's a valid requirement here, but I think there is more work to be done for it to work correctly. 

You've applied the change to the queue consumer connection, but none of the publishers. This means that while you might be able to consume messages from a broker that uses a different `login_method`, the rest of Nameko's AMQP extensions won't work -- for example publishing a message or dispatching an event.

skycastlelily:
Hi,
I have updated the PR accordingly

PR #718, opened 2021-06-05
fix ReplyListener on standalone rpc with custom publisher_options

alfaro28:
The standalone rpc ReplyListener does not inherit any `amqp_uri` or `ssl` options passed to the constructor of its owning client.


mattbennett:
Thank you for this @alfaro28.

mattbennett (approved):

PR #723, opened 2021-09-15
Safe yaml v3

mattbennett:
Equivalent to #722, for v3 branch

--

Updates the logic parsing environment variables in configuration files. Now uses the safe version of pyyaml's parser rather than the unsafe version.

Behaviour is unchanged except for the unsupported case of using nested environment variables without the optional `regex` package. Previously, depending on the structure of the string, a nested environment variable would _sometimes_ be parsed correctly and sometimes incorrectly. Now, if your config file contains a nested environment variable and you do not have the optional `regex` package installed, a `ConfigurationError` will be raised.


timbu (approved):
👍

PR #752, opened 2021-12-14
allow to set priority for message, so consumer can handle message based on priority

genepeng:
Allow to set priority for the message, so the consumer can handle the message based on priority.

**The scenario needs this feature:**
The HTTP request and cronjob share the same RPC, the performance of the HTTP request will be bad even blocked while cronjob is running, as the cronjob usually trigger lots of RPC call, so set higher priority for HTTP RPC message to avoid wanting too long for HTTP request


pace-gene:
Not sure about this, I feel this could be solved by having more than one instance of the service and using a load-balancer for the HTTP requests, or by isolating the HTTP handlers in dedicated server instances by grouping them in a class and specifying that class on the `nameko run` command line.

genepeng:
> Not sure about this, I feel this could be solved by having more than one instance of the service and using a load-balancer for the HTTP requests, or by isolating the HTTP handlers in dedicated server instances by grouping them in a class and specifying that class on the `nameko run` command line.

Maybe I didn't describe that clearly, **Http request** doesn't mean **nameko http request**, in my project, we used GraphQL as Http gateway, and GraphQL call nameko rpc, I want GraphQL call rpc with the highest prority to avoid GraphQL http timeout while cronjob is running

PR #757, opened 2022-02-23
REDACT secrets in error message for HA connection strings

trolldbois:
When using a multi host connection string "amqp://127.0.0.1;amqp://127.0.0.2", kombu.connection.Connection accepts either a ';' delimited list, or a Sequence (list, set...)
The nameko sanitize_url does not properly sanitize the multiple connection strings in these cases as it expects only a single url.

### DIFF
--- a/nameko/utils/__init__.py
+++ b/nameko/utils/__init__.py
@@ -134,3 +134,14 @@ def sanitize_url(url):
     parts = parts._replace(netloc='{}:{}@{}'.format(
         parts.username, REDACTED, host_info))
     return parts.geturl()
+
+
+def sanitize_connection_string(s):
+    """Sanitize the connection string"""
+    if not isinstance(s, str):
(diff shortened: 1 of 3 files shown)


nameko/utils/__init__.py
```diff
@@ -134,3 +134,14 @@ def sanitize_url(url):
     parts = parts._replace(netloc='{}:{}@{}'.format(
         parts.username, REDACTED, host_info))
     return parts.geturl()
+
+
+def sanitize_connection_string(s):
```
pace-gene, line 139: This function should have a name that clearly indicates it can support multiple connection strings, so it isn't mistakenly used instead of `sanitize_url()` for those places that only support a single connection string.

Please rename this function. Also please add a docstring to indicate the supported input types and formats.

pace-gene (changes requested):

PR #758, opened 2022-02-24
feature: support_sustom_backdoor_bind_address, 0.0.0.0 and 127.0.0.1

ponponon:
When deploying a `nameko` project using a `K8s` cluster, a health detection feature is needed. It just so happens that `nameko` makes use of `eventlet`, which provides a `backdoor` mechanism that can be used to implement health detection.

For example, start the `nameko` service using the following.

```shell
nameko run services:AddService --config . /config.yaml --backdoor-port 50000       
```

I can simulate the health detection function with the following command.

```shell
─➤ nc -zv 127.0.0.1 50000
Connection to 127.0.0.1 50000 port [tcp/*] succeeded!
```

But `nameko`'s fixed `localhost`, which makes it impossible to do health checks from remote machines!

```python
def setup_backdoor(runner, port):
    def _bad_call():
        raise RuntimeError(
            'This would kill your service, not close the backdoor. to exit, '
            'use ctrl-c.')
    socket = eventlet.listen(('localhost', port))
    # work around https://github.com/celery/kombu/issues/838
    socket.settimeout(None)
    gt = eventlet.spawn(
        backdoor.backdoor_server,
        socket,
        locals={
            'runner': runner,
            'quit': _bad_call,
            'exit': _bad_call,
        })
    return socket, gt
```

I think the author limited it to `localhost` for security reasons, but I wanted to be able to `0.0.0.0`

So I added a `--backdoor-bind-address` parameter, so we can only choose whether it's `0.0.0.0` or `127.0.0.1`, the default is `127.0.0.1`

### DIFF
--- a/nameko/cli/commands.py
+++ b/nameko/cli/commands.py
@@ -97,6 +97,11 @@ class Run(Command):
             help='RabbitMQ broker url')
 
         parser.add_argument(
+            '--backdoor-bind-address', type=str, default='127.0.0.1',
+            help='Support 0.0.0.0 and 127.0.0.1,'
+            ' the former can be accessed by other machines in the network')
+
+        parser.add_argument(
             '--backdoor-port', type=int,
             help='Specify a port number to host a backdoor, which can be'
             ' connected to for an interactive interpreter within the running'
@@ -171,3 +176,4 @@ class Test(Command):
 
 
 commands = Command.__subclasses__()  # pylint: disable=E1101
+
--- a/nameko/cli/run.py
+++ b/nameko/cli/run.py
@@ -92,12 +92,14 @@ def import_service(module_name):
     return found_services
 
 
-def setup_backdoor(runner, port):
+def setup_backdoor(runner, address, port):
     def _bad_call():
         raise RuntimeError(
             'This would kill your service, not close the backdoor. To exit, '
             'use ctrl-c.')
-    socket = eventlet.listen(('localhost', port))
+    address = address if address in {'0.0.0.0', '127.0.0.1'} else '127.0.0.1'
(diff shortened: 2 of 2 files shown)


nameko/cli/run.py
```diff
@@ -92,12 +92,14 @@ def import_service(module_name):
...
 
 
-def setup_backdoor(runner, port):
+def setup_backdoor(runner, address, port):
     def _bad_call():
         raise RuntimeError(
             'This would kill your service, not close the backdoor. To exit, '
             'use ctrl-c.')
-    socket = eventlet.listen(('localhost', port))
+    address = address if address in {'0.0.0.0', '127.0.0.1'} else '127.0.0.1'
```
pace-gene, line 100: I think it would be valid to specify other bind addresses than just `0.0.0.0` and `127.0.0.1`. Also this doesn't take into account that the bind address could support ipv6.

It may be simpler to just accept the bind address that is supplied or default to `127.0.0.1` and log a warning if the bind address is not `127.0.0.1` or `::1`.

nameko/cli/commands.py
```diff
@@ -96,6 +96,11 @@ def init_parser(parser):
             '--broker', default='pyamqp://guest:guest@localhost',
             help='RabbitMQ broker url')
 
+        parser.add_argument(
+            '--backdoor-bind-address', type=str, default='127.0.0.1',
+            help='Support 0.0.0.0 and 127.0.0.1,'
```
pace-gene, line 101: See my other comment about allowing any address here.

pace-gene (changes requested):

PR #774, opened 2023-06-12
Fix bug in utils.retry where delay is shared between calls

joechild-pace:
If a function decorated with `utils.retry` is called more than once, the backoff shouldn't accumulate between calls

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,7 +20,6 @@ jobs:
       matrix:
         python-version:
           - 3.9
-          - 2.7
 
     steps:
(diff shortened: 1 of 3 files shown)


joechild-pace:
It looks like github has dropped support for python 2.7 in setup-python https://github.com/actions/setup-python/issues/672

The easiest fix is probably to remove it from CI (which I've done here) and look towards dropping support for it too

mattbennett (approved):
Nice catch and test.

Happy to drop 2.7 support here in CI. The next major release will drop it entirely.

pace-gene (approved):

PR #777, opened 2023-10-03
drop py2 support

stephenc-pace:


setup.py
```diff
@@ -42,14 +42,10 @@
     ],
     extras_require={
         'dev': [
-            "coverage==5.5 ; python_version<'3'",
             "coverage==7.1.0 ; python_version>'3'",
```
pace-gene: Do we need to retain the `python_version>'3'` condition if we've dropped support for 2.x?

stephenc-pace: nope, I've removed these all.

pace-gene (approved):

iky (approved):

PR #780, opened 2023-11-03
Add Upgrading to Nameko 3 section to docs

iky:


docs/upgrading_to_nameko_3.rst
```diff
@@ -0,0 +1,347 @@
...
+What's new in Nameko 3
+----------------------
+
+Configuration improvements
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* Configuration loaded from a file or declared in the CLI is now available in a :ref:`global config object <nameko-config>`.
+* Configuration may now be :ref:`declared on the CLI <cli-define-option>`.
+* Configuration access via the service container is now deprecated.
+* Builtin extensions and other facilities that read the configuration now access it via the global.
```
pace-gene, line 15: >now access it via the global.

That isn't very grammatically correct or understandable. I suggest instead:

>Builtin extensions and other facilities that access the configuration now access it via the `nameko.config` import rather than the deprecated service container.

docs/upgrading_to_nameko_3.rst
```diff
@@ -0,0 +1,347 @@
...
+
+CLI improvements
+~~~~~~~~~~~~~~~~
+
+* The CLI has been rebuilt using `click <http://click.pocoo.org/>`_ to be more robust, reliable and extensible..
```
pace-gene, line 20: `rebuilt` => `rewritten`

pace-gene, line 20: Two periods at the end of the sentence can become one.

docs/upgrading_to_nameko_3.rst
```diff
@@ -0,0 +1,347 @@
...
+* The `backdoor` command has been removed; `run`` informs about how to connect to the backdoor port.
```
pace-gene, line 21: Two backquotes at the after `run`.

pace-gene, line 21: I suggest the following:

```
The `backdoor` command has been removed; when executing the `run` command, the backdoor port is instead communicated on the standard output
```

docs/upgrading_to_nameko_3.rst
```diff
@@ -0,0 +1,347 @@
...
+                exchange=config["MY_EXCHANGE"],
+                routing_key=config["MY_ROUTING_KEY"],
+                name=config["MY_QUEUE_NAME"]
+            ),
+            prefetch_count=config["MY_CONSUMER_PREFETCH_COUNT"]
+        )
+        def consume(self, payload):
+            pass
+
+As you can see having config available outside service container also helps with configuring entrypoints and dependency providers by passing options to their constructors:
```
pace-gene, line 62: I suggest:

```
This change makes it much easier to configure entrypoints and dependency providers by setting their constructor options directly from the global config object as illustrated below:
```

docs/upgrading_to_nameko_3.rst
```diff
@@ -0,0 +1,347 @@
...
+
+    class Service:
+        db = Database(engine_options=config["DB_ENGINE_OPTIONS"])
+
+.. _cli-define-option:
+
+Declaration of config in the CLI
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Configuration can now be declared on the CLI, rather than only being loaded from a file. This is useful for setting configuration that is specific to the environment in which the service is running, such as the hostname of a database server.
```
pace-gene, line 77: Please specify if the CLI definitions are config overrides. i.e. if a config file is provided, and still some options are defined on the CLI, which will take precedence?

docs/upgrading_to_nameko_3.rst
```diff
@@ -0,0 +1,347 @@
...
+
+.. _using-config-in-tests:
+
+Using config in tests
+.....................
+
+**Required ⚠** — Read global `config` rather than any of the `x_config`` fixtures.
+There is a breaking change to the way that the `rabbit_config`, `test_config` and `empty_config` fixtures work. They no longer return the config dictionary.
+
+If you are using one of these fixtures, see the recommended pattern in the detailed breakdown of these configuration changes.
```
pace-gene, line 127: Could we link to the "detailed breakdown" here so users can easily see what they're expected to do?

docs/upgrading_to_nameko_3.rst
```diff
@@ -0,0 +1,347 @@
...
+
+See the detailed breakdown of these changes for more information.
+
+.. _using-custom-runners:
+
+Using custom runners
+....................
+
+Required ⚠ — Don’t pass config to ServiceContainer or ServiceRunner
+In Nameko 3, the ServiceContainer and ServiceRunner classes do not accept a config argument anymore. If you are programmatically creating a service runner, rather than using nameko run in the CLI, you must set up config in the global context before using them, for example with nameko.config.patch.
```
pace-gene, line 140: Format nameko.config.patch as code:

```rst
... example with `nameko.config.patch`.
```

pace-gene, line 140: Please link to `Config patching` patching below so users can navigate quickly.

docs/upgrading_to_nameko_3.rst
```diff
@@ -0,0 +1,347 @@
...
+
+See the detailed breakdown of these changes for more information.
```
pace-gene, line 142: This had me looking for another document when I think you're mentioning the `Config patching` section below. Please reformulate to `See below for the detailed breakdown [...]` or `See the *Config patching* section for [...]`.

docs/upgrading_to_nameko_3.rst
```diff
@@ -0,0 +1,347 @@
...
+
+In particular:
+
+The `container_factory` and `runner_factory` fixtures still accept config, even as a deprecated argument. The `ServiceContainer.config` attribute still returns a reference to the config object. This is also deprecated and will raise a warning.
+Before Nameko 2 is sunsetted, extensions should make the necessary adjustments to read from the global config object.
+
+Test Upgrades
+.............
+
+Although extensions will be backwards-compatible when used in services running Nameko 3, some changes may be required to their tests. These are the same changes that apply when :ref:`reading config inside tests for service code <using-config-in-tests>`.
```
pace-gene, line 167: It is Nameko that is backwards compatible with the extensions, not the reverse. Please reformulate as:

> Although Nameko 3 is backwards-compatible with Nameko 2 extensions, some changes [...]

pace-gene (changes requested):
Saving this first half of the review and carrying on.

docs/upgrading_to_nameko_3.rst
```diff
@@ -0,0 +1,347 @@
...
+    runner.add_service(ServiceB)
+
+    runner.start()
+    # AMQP extensions will connect as someuser to RabbitMQ broker at somehost ...
+    runner.stop()
+
+Other configuration related changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* The `--broker`` CLI option is deprecated in favour of `--define` and `--config`. TODO: doesn’t actually raise a warning anymore
```
pace-gene, line 248: Is it worth opening an issue for this rather than have a `TODO`?

docs/upgrading_to_nameko_3.rst
```diff
@@ -0,0 +1,347 @@
...
+
+Testing services
+----------------
+
+In Nameko 2, configuration was always explicitly passed as an object, which made it easy to manage in tests. The global config object needs some different ways of managing configuration in tests.
+
+Config patching
+~~~~~~~~~~~~~~~
+
+The global `config` object has a `patch` helper object which can be used as a context manager or as a decorator. Here is an example of patching the config object in a pytest fixture:
```
pace-gene, line 259: As the config global object is basically just a dictionary, there is nothing to prevent the user from simply changing its values with a `config["blah"] = 123`. Please add an explanation as to why `patch()` is important. Namely because it only applies the patched changes to the current execution context and will revert to the initial values once out of the context.

docs/upgrading_to_nameko_3.rst
```diff
@@ -0,0 +1,347 @@
...
+    @pytest.fixture
+    def memory_rabbit_config():
+        with nameko.config.patch({"AMQP_URI": "memory://"}):
+            yield
+
+While this fixture is in place, any code accessing `config['AMQP_URI']` will receive the value `"memory://"`.
+
+Changes to the `container_factory` and `runner_factory` fixtures.
+
+In Nameko 2, the `container_factory` and `runner_factory` fixtures accepted `config` as a required argument. In Nameko 3, it is optional. If passed, `nameko.config.patch` is used to wrap running container. The patch is applied before container start and returned it back to its original value when exiting the test:
```
pace-gene, line 275: Reformulate:

> [...]. If `config` is passed, the factory will use `config.patch()` to emulate the old behaviour. This will cause the config values to be reverted to their original values once the test has exited.

docs/upgrading_to_nameko_3.rst
```diff
@@ -0,0 +1,347 @@
...
+
+.. code-block:: python
+
+    def test_with_container_factory(container_factory):
+        container = container_factory(Service, config={"AMQP_URI": "..."})
+        container.start()  # <- starts config patching scope, ending it when container
```
pace-gene, line 281: Aesthetics: comments are not lined vertically.

docs/upgrading_to_nameko_3.rst
```diff
@@ -0,0 +1,347 @@
...
+
+    def test_with_runner_factory(runner_factory):
+        runner = runner_factory(config={"AMQP_URI": "..."}, ServiceX, ServiceY)
+        runner.start()  # <- starts config patching scope, ending it when runner
+                        #    stops (exit of the runner_factory fixture)
+        # ...
+
+The optional argument provides backward compatibility so that tests for Nameko extensions can be compatible with both major versions.
+
+When creating or upgrading existing services to run on Nameko 3, it is recommended to use the new config patch pattern instead:
```
pace-gene, line 293: Because using the alternative to `config.patch()` is deprecated, I would reformulate this in a more directive manner:

> When creating or upgrading services to run on Nameko 3, manipulate config options through the new `config.patch` pattern like so:

docs/upgrading_to_nameko_3.rst
```diff
@@ -0,0 +1,347 @@
...
+
+.. warning:: Breaking Change!
+
+    The following section contains a breaking change.
+
+* `empty_config`
+* `rabbit_config`
+* `web_config`
+
+These fixtures no longer return the config dictionary. They simply add their configuration values to the existing `config` object.
```
pace-gene, line 330: Those functions make use of `config.patch()` as shims too, so the config change is transient and context-scoped (it's undone at the end of the current test). Please mention this.

docs/upgrading_to_nameko_3.rst
```diff
@@ -0,0 +1,347 @@
...
+
+A recommended pattern for setting up a config object that draws from these fixtures is as follows:
```
pace-gene, line 332: > Find below an example of how to set up the config values prior to running a test suite. The use of `config.patch` ensures those changes do not leak to other test suites.

pace-gene (changes requested):

gawry:
Hey, do you need any help with this?

